PR #396, opened 2020-10-22
allow compute status without signature

alexcos20:
requires new provider.
New change:  added optional sign parameter to status function:

 /**
   * Returns information about the status of all compute jobs, or a single compute job.
   * @param  {Account} consumerAccount The account of the consumer ordering the service.
   * @param  {string} did Decentralized identifier.
   * @param  {string} jobId The ID of the compute job to be stopped
   * @param  {boolean} sign If the provider request is going to be signed(default) (full status) or not (short status)
   * @return {Promise<ComputeJob[]>} Returns the status
   */
  public async status(
    consumerAccount: Account,
    did?: string,
    jobId?: string,
    sign = true
  ): Promise<ComputeJob[]> {

### DIFF
--- a/src/provider/Provider.ts
+++ b/src/provider/Provider.ts
@@ -186,7 +186,7 @@ export class Provider extends Instantiable {
       const signature = await this.createHashSignature(consumerAccount, signatureMessage)
       url += `&signature=${signature}`
     }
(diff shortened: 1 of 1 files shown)


trentmc:
Making high priority to prevent further slippage of quality.

PR #424, opened 2020-10-28
Feature/ddo-validate

### LINKED ISSUE #400: Validate DDO before publish
Validate DDO before publish/update by:

- create a new function in src/metadatacache/MetadataCache.ts which calls aquarius's endpoint validateDDO
- update src/ocean/Assets.ts and alter create & updateMetadata functions to validate a DDO before publish/update
- have unit tests with bad DDO

0x3bfc:
closes #400


src/metadatacache/MetadataCache.ts
```diff
@@ -312,6 +312,19 @@ export class MetadataCache {
...
   }
 
+  /**
+   * Validate DDO.
+   * @param  {ddp} string DDO.
+   * @return {Promise<any>} Result (200 OK, 500 INVALID).
+   */
+  public async validate(ddo: DDO): Promise<any> {
+    return await this.fetch
+      .post(`${this.url}${apiPath}/validate`, JSON.stringify(ddo))
```
ssallam, line 322: The validate endpoint expects the metadata json not the whole DDO. Get the metadata service `ddo.service[0].attributes`

kremalicious: `ddo.findServiceByType('metadata)`. Never assuming the metadata service is the first service is safer.

kremalicious:
My main question would be: what is it validating? json structure? metadata structure? Plecos? This should be mentioned somewhere what exactly this validates.

src/metadatacache/MetadataCache.ts
0x3bfc: thanks for the feedback .. working on it.

src/metadatacache/MetadataCache.ts
```diff
@@ -218,6 +218,61 @@ export class MetadataCache {
...
 
   /**
+   * Edit Metadata for a DDO.
+   * @param  {did} string DID.
+   * @param  {newMetadata}  EditableMetadata Metadata fields & new values.
+   * @param  {String} updated Updated field of the DDO
+   * @param  {String} signature Signature using updated field to verify that the consumer has rights
+   * @return {Promise<String>} Result.
+   */
+  public async editMetadata(
```
kremalicious, line 228: so there has been a drastic change, this is now basically in `ocean.assets.editMetadata()` but it only does an object restructuring https://github.com/oceanprotocol/ocean.js/blob/main/src/ocean/Assets.ts#L253

and then the real update is done with `ocean.assets.updateMetadata()` https://github.com/oceanprotocol/ocean.js/blob/main/src/ocean/Assets.ts#L275

super confusing naming and all but if we want to validate anything it needs to happen over there

0x3bfc: >> if we want to validate anything it needs to happen over there

You mean in the `ocean.assets.updateMetadata()` ?

PR #441, opened 2020-11-06
logs/history speed optimizations

alexcos20:
pool logs/history should be much faster.

- added startBlock in the config, so we can narrow the blocks range searched
- lots of parallelization

### DIFF
--- a/src/balancer/OceanPool.ts
+++ b/src/balancer/OceanPool.ts
@@ -1,6 +1,6 @@
 import Web3 from 'web3'
 import { AbiItem } from 'web3-utils/types'
-import { TransactionReceipt } from 'web3-core'
+import { TransactionReceipt, Log } from 'web3-core'
 import { Pool } from './Pool'
 import { EventData, Filter } from 'web3-eth-contract'
 import BigNumber from 'bignumber.js'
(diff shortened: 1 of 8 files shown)


kremalicious:
testing this locally against local market and does not really load faster for me. But see the config changed so we might need to have more changes in react or market somewhere

alexcos20:
do you use a custom config or you are just calling getConfig('mainnet') ?

mihaisc:
The config is just passed to ocean lib in react, so no changes needed there

alexcos20:
@kremalicious @mihaisc - check this new improvement

mihaisc:
I get a lot of nulls from `tokens` , i added `if (tokens != null)` before `for`, not sure if this is ok.
 https://github.com/oceanprotocol/ocean.js/blob/1d717cb3b03b150d16bb684bd13a1d7bdafa55a6/src/balancer/OceanPool.ts#L174

mihaisc:
Isn't this redundant and costly, since in market at least we already know the dt address
https://github.com/oceanprotocol/ocean.js/blob/1d717cb3b03b150d16bb684bd13a1d7bdafa55a6/src/balancer/OceanPool.ts#L889

kremalicious:
This makes the loading time for the pool transactions table go from 4 min. down to 5 sec. for me. SHIP IT

sorcery

kremalicious (approved):

alexcos20:
have some worries that infura limits the number of requests to 10/s, so doing all that in parallel might brake something.

but let's try it :)

kremalicious:
so we have between 100-300 concurrent users at all times on the frontend during a typical day now. So we really should look into what happens when infura fails. Once we announce this, people will flock to their history page

mihaisc:
But these calls are made on client right? So it has no impact on our infura account

PR #448, opened 2020-11-11
update for purgatory

mihaisc:
Added `isInPurgatory` and `purgatoryData` to the DDO


src/ddo/DDO.ts
```diff
@@ -55,6 +56,10 @@ export class DDO {
 
   public price: BestPrice
 
+  public isInPurgatory: boolean
```
kremalicious, line 59: just checking, so every DDO will have this key by default, right?

mihaisc: that's what I'm thinking, if it's cached in aquarius it will have `isInPurgatory` `true` or `false` , while we can ignore setting `purgatoryData` if `false`

PR #455, opened 2020-11-16
fullfill promises

alexcos20:


kremalicious (approved):

src/balancer/OceanPool.ts
```diff
@@ -630,7 +631,7 @@ export class OceanPool extends Pool {
       this.logger.error('ERROR: OCEAN approve failed')
       return null
     }
-    return this.swapExactAmountIn(
+    const tx = this.swapExactAmountIn(
```
kremalicious, line 634: interesting that `this.swapExactAmountIn` does not need `await` to work in my local test

alexcos20: it should work, but never the less..

kremalicious: but typings say they are promises, so they should get an `await`

alexcos20: fun fact is that swapExactAmountIn returns a promise, so as long as you do await BuyDT, you are resolving the promise

mihaisc: same here, i can sell a dt, whilst on prod i can't

PR #462, opened 2020-11-18
Add Ocean.js CLI tool

alexcos20:
add cli on top of ocean.js

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -8,7 +8,8 @@
     "project": [
       "./tsconfig.json",
       "./test/unit/tsconfig.json",
-      "./test/integration/tsconfig.json"
+      "./test/integration/tsconfig.json",
+      "./example/tsconfig.json"
     ]
   },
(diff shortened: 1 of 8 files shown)


0x3bfc (approved):
LGTM

kremalicious:
I would advise against merging it in this state. It will drag down the Maintainability score from a bad C to an even worse D on CodeClimate. And CodeClimate is right, it highly increases the amount of maintainability needed to keep the codebase just updated, with double the packages to maintain and update. How do we make sure to update the example on every code change? What tests are there to ensure the example works after code changes? How do external devs know their code changes do not break the example?

An example within this project should make use and import the files directly from the root project, without introducing a whole bunch of new dependencies only used for example. If the example is a CLI then well, maybe there should be a CLI repository and the example should be in there?

alexcos20:
do not merge or delete this, I'm going to move it to another repo

PR #468, opened 2020-11-23
Handle assets with low liquidity

mihaisc:
`isConsumable` is : 
- 'true' when pool/exchange created and enough datatokens are available ( can we check for exchange?)
- 'false' when pool/exchange created and not enough datatokens are avaialbe
- '' when pool/exchange not created


alexcos20 (approved):
lgtm

PR #492, opened 2020-12-09
add ropsten

alexcos20:


kremalicious (approved):
It should just work™ over in `market` once we bump @oceanprotocol/contracts there too

kremalicious:
or not :-) https://github.com/oceanprotocol/market/pull/294

And ocean.js loads the contracts so all good from that side.

alexcos20:
I'm going to merge it once we have aqua & provider deployed.

PR #493, opened 2020-12-09
CheckURL created.

mariacarmina:
Implemented a function which communicates through HTTP calls with the provider (with the checkURL endpoint) and returns the status fetched by the provider.


calina-c:
Duplicated by #483.

PR #531, opened 2021-01-08
Datokens: check cap before mint

### LINKED ISSUE #408: Datokens: check cap before mint and return an error if mintAmount>totalcap
in the current code, the transaction will fail

bogdanfazakas:
compared cap with amount entered to mint
if exceeds an error is thrown

closes #408

### DIFF
--- a/src/datatokens/Datatokens.ts
+++ b/src/datatokens/Datatokens.ts
@@ -176,15 +176,12 @@ export class DataTokens {
     })
     const capAvailble = await this.getCap(dataTokenAddress)
     if (parseFloat(capAvailble) >= parseFloat(amount)) {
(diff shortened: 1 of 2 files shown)


kremalicious:
should work™, bonus points if we could have a unit test for this maybe in https://github.com/oceanprotocol/ocean.js/blob/main/test/unit/Datatokens.test.ts

bogdanfazakas:
> should work™, bonus points if we could have a unit test for this maybe in https://github.com/oceanprotocol/ocean.js/blob/main/test/unit/Datatokens.test.ts, along the lines of `should fail when mint amount exceeds cap`

sure 👍   i'll take care of that.

kremalicious (approved):

PR #537, opened 2021-01-12
fix download for large files

mihaisc:
Fix for downloading large files in the market (browser).
There is no test for the new function because we would need another library to generate a document object and it would be used just in this isolated context.


alexcos20 (approved):
lgtm

PR #540, opened 2021-01-13
[WIP] Adds checkURL function to use provider endpoint.

### LINKED ISSUE #525: Add a function to check URL using the provider checkURL endpoint
The provider has a new endpoint to check the URL contentType and contentLength before proceeding. If the URL is missing, wrong, non-confoormant, does not support GET, or does not return anything, the checkURL endpoint returns a 400 response. This issue adds the possibility to call the checkURL endpoint from ocean.js.

calina-c:
Supersedes #483. Closes #525.


alexcos20 (approved):
lgtm

PR #542, opened 2021-01-13
checkURL interface

alexcos20:


calina-c (approved):
nice :D

src/provider/Provider.ts
```diff
@@ -11,6 +11,11 @@ import { DDO } from '../ddo/DDO'
 
 const apiPath = '/api/v1/services'
 
+export interface urlDetails {
```
kremalicious, line 14: types and interfaces should always start uppercase

kremalicious:
So is it possible to get that method without initializing the whole `ocean` object? Otherwise we would need to block adding any files in UI as long as user does not have any wallet connected

src/provider/Provider.ts
```diff
@@ -73,20 +78,23 @@ export class Provider extends Instantiable {
     }
   }
 
-  public async checkURL(url: string): Promise<Record<string, string>> {
```
kremalicious, line 76: should be named `checkUrl`

kremalicious:
not sure yet why https://github.com/oceanprotocol/provider/pull/57 introduces `/checkUrl` &  `/assetInfo` and what is best to use here

src/provider/Provider.ts
alexcos20: it was indented to mimic https://github.com/oceanprotocol/market/blob/main/src/utils/index.ts#L45
I'm going to update it

alexcos20:
> So is it possible to get that method without initializing the whole `ocean` object? Otherwise we would need to block adding any files in UI as long as user does not have any wallet connected

You need a provider class.  That should be possible without web3

alexcos20:
> not sure yet why [oceanprotocol/provider#57](https://github.com/oceanprotocol/provider/pull/57) introduces `/checkUrl` & `/assetInfo` and what is best to use here

checkUrl takes an URL (http:///www.example.com) and check if your selected provider is able to access it and gets file information (length & type)
/assetInfo takes a DID with some files (you don't have the URLs) and will send back the same interface, but with updated values.

Let's say that you published www.example.com as DID1.  and you are updating the content, but in the DDO you still have the old length of the file.   By calling /assetInfo, provider can decrypt the URLs , check if they are still valid, get the file info and then you can update your DDO with the new info

alexcos20:
superseded by #544

PR #544, opened 2021-01-14
add updateServiceTimeout and edit ddo updates

### LINKED ISSUE #536: Edit timeout in ocean.js
Allow service.timeout updates

alexcos20:
Breaking changes in editMetadata  definition !

closes #536 and #525

Please make sure that you call ocean.OnChainMetadataCache.update after using the helpers, otherwise your asset it's not going to be updated on chain

### DIFF
--- a/src/ocean/Assets.ts
+++ b/src/ocean/Assets.ts
@@ -246,82 +246,37 @@ export class Assets extends Instantiable {
 
   /**
    * Edit Metadata for a DID.
-   * @param  {did} string DID. You can leave this empty if you already have the DDO
    * @param  {ddo} DDO if empty, will trigger a retrieve
    * @param  {newMetadata}  EditableMetadata Metadata fields & new values.
-   * @param  {Account} account Ethereum account of owner to sign and prove the ownership.
    * @return {Promise<DDO>} the new DDO
    */
-  public async editMetadata(
-    did: string,
-    ddo: DDO,
-    newMetadata: EditableMetadata
-  ): Promise<DDO> {
-    if (!ddo) ddo = await this.ocean.metadatacache.retrieveDDO(did)
+  public async editMetadata(ddo: DDO, newMetadata: EditableMetadata): Promise<DDO> {
     if (!ddo) return null
     for (let i = 0; i < ddo.service.length; i++) {
-      if (ddo.service[i].type === 'metadata') {
-        if (newMetadata.title) ddo.service[i].attributes.main.name = newMetadata.title
-        if (!ddo.service[i].attributes.additionalInformation)
-          ddo.service[i].attributes.additionalInformation = Object()
-        if (newMetadata.description)
-          ddo.service[i].attributes.additionalInformation.description =
-            newMetadata.description
-        if (newMetadata.links)
-          ddo.service[i].attributes.additionalInformation.links = newMetadata.links
-      }
+      if (ddo.service[i].type !== 'metadata') continue
+      if (newMetadata.title) ddo.service[i].attributes.main.name = newMetadata.title
+      if (!ddo.service[i].attributes.additionalInformation)
+        ddo.service[i].attributes.additionalInformation = Object()
+      if (newMetadata.description)
+        ddo.service[i].attributes.additionalInformation.description = newMetadata.description
+      if (newMetadata.links)
+        ddo.service[i].attributes.additionalInformation.links = newMetadata.links
     }
     return ddo
   }
 
   /**
-   * Update Compute Privacy
-   * @param  {did} string DID. You can leave this empty if you already have the DDO
-   * @param  {ddo} DDO if empty, will trigger a retrieve
-   * @param  {number} serviceIndex Index of the compute service in the DDO. If -1, will try to find it
-   * @param  {ServiceComputePrivacy} computePrivacy ComputePrivacy fields & new values.
-   * @param  {Account} account Ethereum account of owner to sign and prove the ownership.
-   * @return {Promise<DDO>}
-   */
-  public async updateComputePrivacy(
-    did: string,
-    ddo: DDO,
-    serviceIndex: number,
-    computePrivacy: ServiceComputePrivacy
-  ): Promise<DDO> {
-    if (!ddo) ddo = await this.ocean.metadatacache.retrieveDDO(did)
-    if (!ddo) return null
-    if (serviceIndex === -1) {
-      const service = ddo.findServiceByType('compute')
-      if (!service) return null
-      serviceIndex = service.index
-    }
-    if (typeof ddo.service[serviceIndex] === 'undefined') return null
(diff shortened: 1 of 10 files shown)


alexcos20:
waiting for https://github.com/oceanprotocol/ocean.js/pull/542 , so I can rebase, that's the only test that is failing no.

src/ocean/Assets.ts
```diff
@@ -239,82 +239,96 @@ export class Assets extends Instantiable {
...
-    if (oldDdo.service[serviceIndex].type !== 'compute') return null
-    oldDdo.service[serviceIndex].attributes.main.privacy.allowRawAlgorithm =
+    if (!ddo) ddo = await this.ocean.metadatacache.retrieveDDO(did)
+    if (!ddo) return null
+    if (serviceIndex === -1) {
+      const service = ddo.findServiceByType('compute')
+      if (!service) return null
+      serviceIndex = service.index
+    }
+    if (typeof ddo.service[serviceIndex] === 'undefined') return null
```
kremalicious, line 299: all these null returns could be done in one if statement in line 293 instead of sprinkling the conditions throughout multiple lines:

```
if (
  !ddo ||
  ddo.service[serviceIndex].type !== 'compute' ||
  typeof ddo.service[serviceIndex] === 'undefined'
)
  return null
```

src/ocean/Assets.ts
```diff
@@ -239,82 +239,96 @@ export class Assets extends Instantiable {
...
-    let i
-    for (i = 0; i < oldDdo.service.length; i++) {
-      if (oldDdo.service[i].type === 'metadata') {
-        if (newMetadata.title) oldDdo.service[i].attributes.main.name = newMetadata.title
-        if (!oldDdo.service[i].attributes.additionalInformation)
-          oldDdo.service[i].attributes.additionalInformation = Object()
+    if (!ddo) ddo = await this.ocean.metadatacache.retrieveDDO(did)
+    if (!ddo) return null
+    for (let i = 0; i < ddo.service.length; i++) {
+      if (ddo.service[i].type === 'metadata') {
```
kremalicious, line 263: to prevent excessive nesting of whole blocks like this, it should turn around the logic so the whole block does not need to be wrapped:

```
if (ddo.service[i].type !== 'metadata') return null
```

src/ocean/Assets.ts
```diff
@@ -239,82 +239,96 @@ export class Assets extends Instantiable {
...
+    newMetadata: EditableMetadata
   ): Promise<DDO> {
-    const oldDdo = await this.ocean.metadatacache.retrieveDDO(did)
-    let i
-    for (i = 0; i < oldDdo.service.length; i++) {
-      if (oldDdo.service[i].type === 'metadata') {
-        if (newMetadata.title) oldDdo.service[i].attributes.main.name = newMetadata.title
-        if (!oldDdo.service[i].attributes.additionalInformation)
-          oldDdo.service[i].attributes.additionalInformation = Object()
+    if (!ddo) ddo = await this.ocean.metadatacache.retrieveDDO(did)
```
kremalicious, line 260: we should decide if this method receives DID or DDO, not allow both. I would stick to only DID since it prevents accidental wrong DDO retrieval's by the user which might get passed. This would then also reduce the logic within the method a bit.

src/ocean/Assets.ts
```diff
@@ -239,82 +239,96 @@ export class Assets extends Instantiable {
     return (await this.ocean.metadatacache.queryMetadata(searchQuery)).results
   }
 
+  /**    Metadata updates
+   *  Don't forget to call ocean.OnChainMetadataCache.update after using this functions
```
kremalicious, line 243: and why does this method here not do that for me?

kremalicious: got it, because of the method split-up. I know conceptually updating timeout is not updating the metadata service but maybe it makes sense to handle that within `editMetadata` too is since from user perspective it's also just metadata.

So `EditableMetadata` could have:

```ts
timeout?: {
  serviceIndex: number // could also ask for `serviceName` which would be a bit more user friendly
  value: number
}
```

and if passed we handle that in that method (where `updateServiceTimeout` can be left as individual method, but called here), allowing us to also call `ocean.OnChainMetadataCache.update` at the end. Thoughts?

src/ocean/Assets.ts
alexcos20: actually, the logic is:
 - try to get the ddo if we only have the DID (if we have the DDO already, avoid extra call to aqua)
 - if serviceIndex == -1 -> try to search for compute service
 - else check if serviceIndex is a valid compute service (maybe it's not)

src/ocean/Assets.ts
alexcos20: no, it just needs to loop the services and update the metadata.
if somebody is going to publish a ddo with access service first and then metadata, using return will just exit and you end up with un-updated metadata..

src/ocean/Assets.ts
alexcos20: if we have the DID, then we fetch the DDO from aqua.  but it we have the ddo, then use that in order to avoid an unnecessary call to aqua (most likely market already has the ddo)

src/ocean/Assets.ts
alexcos20: I'm afraid that we are going to end up with a huge function that does all..  also, don't forget that we have compute updates as well, and who knows what kind of updates are we going to have in the future..
That's why I thought that it makes sense to have update helpers (functions that are going to update parts of the ddo) and once you are done, you only have 1 call that updates the ddo on chain.
What if you want to change computePrivacy, timeout and asset's name ?

kremalicious: guess I'm arguing for a helper method doing all then, calling the sub-functions. Since we know which fields are editable we should provide a helper so devs do not need to figure out for themselves which parts of DDO are editable and then which ones require which method to update.

This helper method handling all editable fields could simply call the individual sub-functions. Then it also stays flexible, advanced users can use the individual methods, but most will use the main helper method with a bunch of optional params. And then also stay consistent in the naming to group the methods, like:

- `editMetadata()`
  - `editMetadataFields()`
  - `editTimeout()`
  - `editComputePrivacy()`

If we decide to keep as is then at least solve this mixup of `edit` & `update` and stick to one

src/ocean/Assets.ts
alexcos20: solved, using ddo only

mihaisc:
So after we call `editMetadata` and/or `editServiceTimeout` we should call `ocean.OnChainMetadataCache.update` ? If yes, why is the update in `OnChainMetadataCache` ? It has nothing to do with the cache. Considering where it is, this function should update the cache.

alexcos20:
actually we have another class for that, it's called MetadataCache :)

mihaisc (approved):

Issue #284, opened 2020-09-16
FixedPriceExchange BuyDT MetaMask error on Rinkeby


mihaisc:
When I buy a token from a fixed price exchange MetaMask throws the following error in the browser console. All the flow is working ok: I buy the dt, order asset, download asset. So basically the only issue is that there is an unknown error in the console and someone may notice is and think that something broke. This is a minor issue.

![image](https://user-images.githubusercontent.com/26000280/93313403-1b0a1700-f811-11ea-93b7-1a627cceaed9.png)

marcojrfurtado:
I am getting this pretty consistently on Rinkeby for larger transactions

PR #557, opened 2021-01-21
fix metamask bug

### LINKED ISSUE #543: OnChainMetadata estimate gas always fails
Both 
https://github.com/oceanprotocol/ocean.js/blob/main/src/metadatacache/OnChainMetaDataCache.ts#L114  and https://github.com/oceanprotocol/ocean.js/blob/main/src/metadatacache/OnChainMetaDataCache.ts#L156  
always fails and we end up with the default gas limit, although transaction always goes through.

It's not a major bug, but has 2 major inconveniences:
1 - users are scared by the high cost (real gas usage for transaction is only a fraction)
2 - in future, for very large ddos, our default could not be enough.

alexcos20:
Fixes #543 and #284 

Changes proposed in this PR:
 all estimateGas functions should have the from parameter
- 
- 
-

### DIFF
--- a/src/balancer/Pool.ts
+++ b/src/balancer/Pool.ts
@@ -83,7 +83,7 @@ export class Pool extends PoolFactory {
           baseTokenWeight,
           swapFee
         )
-        .estimateGas((err, estGas) => (err ? gasLimitDefault : estGas))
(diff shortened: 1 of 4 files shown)


bogdanfazakas (approved):

ssallam (approved):
LGTM.

PR #558, opened 2021-01-21
Dynamic provider endpoints

### LINKED ISSUE #538: Dynamic provider endpoints
Blocked by https://github.com/oceanprotocol/provider/issues/36

Right now all endpoints are hardcoded (https://github.com/oceanprotocol/ocean.js/blob/main/src/provider/Provider.ts#L264)

Ocean.js should read the provider root endpoint (/ ) and create the required URLs based on that information.
Also errors are going to be thrown if a service that is not exposed is being accessed.

mariacarmina:
Fixes #538.
Finalizes and closes https://github.com/oceanprotocol/multi-repo-issue/issues/57

Changes proposed in this PR:

- ocean.js reads from the provider serviceEndpoints.


package.json
```diff
@@ -48,6 +48,7 @@
     "lzma": "^2.3.2",
     "node-fetch": "^2.6.1",
     "save-file": "^2.3.1",
+    "test": "^0.6.0",
```
kremalicious, line 51: do we specifically need to add this dependency? Cause we use `mocha` as a test runner and this package seems to be another test runner

src/provider/Provider.ts
```diff
@@ -10,7 +10,11 @@ import { Response } from 'node-fetch'
 import { DDO } from '../ddo/DDO'
 import DID from '../ocean/DID'
 
-const apiPath = '/api/v1/services'
+export interface EndpointInformation {
```
kremalicious, line 13: would it be more clear if we call this interface simply `ServiceEndpoint`?

src/provider/Provider.ts
```diff
@@ -21,20 +25,56 @@ const apiPath = '/api/v1/services'
 export class Provider extends Instantiable {
   public nonce: string
   private baseUrl: string
+  public servicesEndpoints: any
```
kremalicious, line 28: and then we can type the `any` here, so e.g. 

```ts
public servicesEndpoints: ServiceEndpoint[]
```

package-lock.json
```diff
@@ -9519,6 +9524,14 @@
       "integrity": "sha512-wK0Ri4fOGjv/XPy8SBHZChl8CM7uMc5VML7SqiQ0zG7+J5Vr+RMQDoHa2CNT6KHUnTGIXH34UDMkPzAUyapBZg==",
       "dev": true
     },
+    "test": {
```
kremalicious, line 9527: after removing `test` from `package.json` this lock file has to be regenerated to remove the change completely, with `npm install`

kremalicious (approved):
perfect! Tests seem to cover accessing the provider during fileinfo/consume so should work™. We do not have any custom provider implementation so we can't really test this

PR #570, opened 2021-01-26
refactor typedoc generation

kremalicious:
TypeDoc v0.20.0 came with breaking changes, but simplifies a lot of things so we do not have to create our own TypeDoc app to generate the json anymore.

Also results in a way smaller json file so we can highly reduce how we read that out over in the docs. Part fo solving https://github.com/oceanprotocol/docs/issues/527

- refactor typedoc.js to generate initial json with default command, spawned as child process
- merge `src/lib.ts` & `src/index.ts`, leaving only `src/index.ts`


kremalicious:
new format generated with this change confirmed to be working in https://github.com/oceanprotocol/docs/pull/546: https://docs-git-fix-typedoc.oceanprotocol.vercel.app/references/ocean.js

0x3bfc (approved):
LGTM

0x3bfc:
sorry didn't see it is in draft

kremalicious:
not gonna happen now but can inform how we do it in new ocean.js, if at all

PR #585, opened 2021-02-04
remove barge defaults overwrites

kremalicious:
Run CI against latest defaults of Barge. In Barge, `DEPLOY_CONTRACTS` is `true` by default and CI should run against whatever is defined in Barge for the contracts


alexcos20 (approved):

kremalicious:
that was simple, thanks!

PR #587, opened 2021-02-04
Add nativeSearch in SearchQuery

### LINKED ISSUE #453: update SearchQuery typings for new nativeSearch
See:
- https://github.com/oceanprotocol/market/blob/35ec06b75c898585a6bd3e5edf84cd1133b20450/src/components/pages/Home.tsx#L23-L57
- https://github.com/oceanprotocol/market/blob/08965070972b35b34cbf9c837b3dde90128c7fdc/src/components/templates/Search/utils.ts#L18-L35

bogdanfazakas:
Fixes #453 .

Changes proposed in this PR:

- SearchQuery typing update
- update MetadataCache and Assets query's to use new nativeSearch

### DIFF
--- a/src/metadatacache/MetadataCache.ts
+++ b/src/metadatacache/MetadataCache.ts
@@ -84,7 +84,6 @@ export class MetadataCache {
    * @return {Promise<QueryResult>}
    */
   public async queryMetadata(query: SearchQuery): Promise<QueryResult> {
-    this.logger.log('--- queryMetadata querry', query)
     const result: QueryResult = await this.fetch
       .post(`${this.url}${apiPath}/query`, JSON.stringify(query))
       .then((response: Response) => {
--- a/src/metadatacache/MetadataCache.ts
+++ b/src/metadatacache/MetadataCache.ts
@@ -18,7 +18,13 @@ export interface SearchQuery {
   text?: string
   offset?: number
   page?: number
-  query: { [property: string]: string | number | string[] | number[] }
+  query: {
+    nativeSearch?: number
+    // eslint-disable-next-line camelcase
+    query_string: {
+      [property: string]: string | number | string[] | number[]
+    }
+  }
   sort?: { [jsonPath: string]: number }
 }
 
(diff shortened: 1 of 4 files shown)


kremalicious:
thanks for going through this! This is great so far, and covers use cases where people use the `nativeSearch`. But what about everybody using the "old way" of searching? With this change we break it for them, although the functionality is still there.

To me the proper solution has always been to stick to only one way of searching instead of adding 2 different ways. 

So we either:
1. figure typings allowing old & new way of search queries 
2. remove old way from the core

For 2. this PR is then already perfect

kremalicious:
core meaning Aquarius, cause it has 2 ways of searching:
- https://docs.oceanprotocol.com/references/aquarius/#assetsddoesquery
- https://docs.oceanprotocol.com/references/aquarius/#assetsddoquery

This is of course confusing but also not sure how much work it would be to remove the old way of searching in Aquarius. Could be fine deprecating the old way starting in ocean.js, and default to `nativeSearch`.

mihaisc:
I think we should deprecate the old way, we don't plan to maintain the es drivers and whatever is needed

alexcos20:
I agree.  we will only use ES with Aqua from now on

kremalicious:
ok, cool. Which would then make a key like `nativeSearch` redundant if we update Aquarius.

So we would end up in here with only:

```ts
query: {
  query_string: {
    [property: string]: string | number | string[] | number[]
  }
}
```

On the other hand, does Aquarius support all full text queries from here, or just `query_string`? https://www.elastic.co/guide/en/elasticsearch/reference/current/full-text-queries.html

If it supports all of them then in this PR we might want to find some external package which has typings for Elasticsearch queries and rather use that

alexcos20:
Since it's a pass-through, they should all work

kremalicious:
So we could wait for https://github.com/oceanprotocol/aquarius/issues/372 to be able to finalize this here.

Or we just set `nativeSearch` in here by default when hitting `/query`, so users of library do not have to pass that. Then ocean.js would already deprecate all other ways, and once Aquarius is updated we remove passing that `nativeSearch` key

bogdanfazakas:
So I don't know exactly how can we use elasticsearchjs client typings (https://github.com/elastic/elasticsearch-js) in our SearchQuery interface and keep pagination related parameters, but I've added match and simple_query_string text queries.
Also the nativeSearch param is now optional in the SearchQuery interface and if not set will be set in our queryMetadata

kremalicious:
unblocked. New Aquarius has been released where `nativeSearch` does not have to be passed anymore https://github.com/oceanprotocol/aquarius/pull/394

Meaning all those queries are supported under the `query` object https://www.elastic.co/guide/en/elasticsearch/reference/current/full-text-queries.html

alexcos20:
superseeded by #662

PR #599, opened 2021-02-10
speed up previous order

### LINKED ISSUE #602: Consumption is slow
While ordering  a new asset, we have to check for previous valid orders.
Right now, we are scanning the entire chain

alexcos20:
Some code refactor:
 - if there is a timeout, no need to check beyond lastBlock - timeout (and this is generous, because we are assuming 1 block/sec)
 - if timeout = 0 , no need to check beyond startBlock

closes #602


mihaisc (approved):

kremalicious (approved):
nice, PRs which start with "speed up..." are insta-approved

PR #617, opened 2021-02-17
Feature/travis dockerhub auth

alexcos20:
We need to test if this affects PR's opened from forked repos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,7 @@ after_script:
   - ./cc-test-reporter format-coverage -t lcov -o coverage/codeclimate.unit.json coverage/unit/lcov.info # Format unit test coverage
   - ./cc-test-reporter format-coverage -t lcov -o coverage/codeclimate.integration.json coverage/integration/lcov.info # Format integration test coverage
   - ./cc-test-reporter sum-coverage coverage/codeclimate.*.json -p 2 # Sum both coverage parts into coverage/codeclimate.json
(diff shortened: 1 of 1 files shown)


alexcos20:
all fine, I added a check condition

kremalicious:
Usual tactic to make fork builds succeed is to condition ally run those jobs/commands which require env vars. As for barge, pulling images should still succeed for forks within the anonymous tier on DockerHub, right? Unless we have many fork builds at once I guess.

If we're into optimizing for forks, then another candidate is `CC_TEST_REPORTER_ID` where when it is not present no coverage reporting should be tried to send, so this line: https://github.com/oceanprotocol/ocean.js/blob/feature/travis_dockerhub_auth/.travis.yml#L44

alexcos20:
that why we have the env check before docker login:
if [[ -n "$DOCKERHUB_PASSWORD" && -n "$DOCKERHUB_USERNAME" ]]; then echo "logining in to docker";echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin;fi

if DOCKERHUB_USERNAME & DOCKERHUB_PASSWORD are not exported by travis  (they are not exported for external PR), we will not do any auth and barge will download the images as anonymous

alexcos20:
added a check for  CC_TEST_REPORTER_ID too

kremalicious (approved):
[that warning sounds frightening](https://travis-ci.com/github/oceanprotocol/ocean.js/builds/217471344#L284) but I guess we can live with that

PR #627, opened 2021-02-23
Add polygon network

mihaisc:
Related to https://github.com/oceanprotocol/multi-repo-issue/issues/77

Seems to be working on the polygon network. Not sure why the test fails

### DIFF
--- a/src/utils/ConfigHelper.ts
+++ b/src/utils/ConfigHelper.ts
@@ -88,7 +88,7 @@ const configs: ConfigHelperConfig[] = [
     poolFactoryAddress: null,
     fixedRateExchangeAddress: null,
     metadataContractAddress: null,
-    startBlock: 11105522
+    startBlock: 11105459
   },
   {
     networkId: 137,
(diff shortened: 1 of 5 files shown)


src/utils/ConfigHelper.ts
```diff
@@ -89,6 +89,20 @@ const configs: ConfigHelperConfig[] = [
...
+    nodeUri: 'https://rpc-mainnet.maticvigil.com/',
+    metadataCacheUri: 'https://aquarius.polygon.oceanprotocol.com',
+    providerUri: 'https://provider.polygon.oceanprotocol.com',
+    subgraphUri: 'https://subgraph.polygon.oceanprotocol.com',
+    oceanTokenAddress: null,
+    factoryAddress: null,
+    poolFactoryAddress: null,
+    fixedRateExchangeAddress: null,
+    metadataContractAddress: null,
+    startBlock: 0
```
alexcos20, line 105:
```suggestion
    startBlock: 11005222
```

alexcos20:
Seems that the provider is failing from time to time, cause tests failure all around the stack.  We already have a issue for that, we just need to restart the build if it fails

kremalicious:
failing tests on `main` are not an option, so we need to figure this out first with #628. There seems to be more required to make the tests not fail randomly for unknown reasons

alexcos20 (approved):

PR #630, opened 2021-02-24
More integration tests

### LINKED ISSUE #597: Token get deducted but not able to download a file
**Describe the bug**
Token get deducted but not able to download a file

**To Reproduce**
Consume a problematic asset, example invalid provider URL

**Expected behavior**
Check on the provider endpoint is valid before data token exchange during make order

**Additional context**
Example asset:
https://market.oceanprotocol.com/asset/did:op:1aFd29f088baE005fC4B00D8066358b2ceD0683a

alexcos20:
Adding more tests.

closes #597

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ before_script:
   # Barge setup
   - git clone https://github.com/oceanprotocol/barge
   - cd barge
-  - git checkout feature/improve_provider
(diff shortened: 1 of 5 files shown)


kremalicious (approved):
nice, also increases measured coverage, we didn't have this for some time on PRs on here 🥳

PR #662, opened 2021-03-15
add encrypt ddo support

alexcos20:
Breaking changes:
  - ocean.assets.create is not calling ocean.onChainMetadata.publish anymore. It will just return the DDO and you have to call it after to publish your asset

Fixes:
 - query support for aquarius 2.2.5
 - add encrypted ddo support (siloed market)

### DIFF
--- a/src/metadatacache/MetadataCache.ts
+++ b/src/metadatacache/MetadataCache.ts
@@ -95,7 +95,6 @@ export class MetadataCache {
    * @return {Promise<QueryResult>}
    */
   public async queryMetadata(query: SearchQuery): Promise<QueryResult> {
-    if (!query.query.nativeSearch) query.query.nativeSearch = 1
     const result: QueryResult = await this.fetch
       .post(`${this.url}${apiPath}/query`, JSON.stringify(query))
       .then((response: Response) => {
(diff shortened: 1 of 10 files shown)


src/metadatacache/MetadataCache.ts
```diff
@@ -18,7 +18,24 @@ export interface SearchQuery {
   text?: string
   offset?: number
   page?: number
-  query: { [property: string]: string | number | string[] | number[] }
+  query: {
+    nativeSearch?: number
```
bogdanfazakas, line 22: I think we can remove `nativeSearch?: number` completely from the search since it's no longer required or used in Aquarius. This can require some market updates but not sure about that some of the search queries in the market are no longer using search typings from ocean js and are using any as type.

src/ocean/Assets.ts
```diff
@@ -338,11 +343,12 @@ export class Assets extends Instantiable {
...
       page: 1,
       offset: 100,
       query: {
-        value: 1
+        query_string: {
+          query: text
+        }
       },
       sort: {
         value: 1
```
bogdanfazakas, line 354: Shouldn't this be `price.value: 1` if it point to the price value? if it's something else please ignore this comment 😃 . If it point to price value than we need to change also line 240 accordingly.

test/unit/ocean/Assets.test.ts
```diff
@@ -28,7 +28,10 @@ describe('Assets', () => {
         offset: 100,
         page: 1,
         query: {
-          text: 'Office'
+          nativeSearch: 1,
```
bogdanfazakas, line 31: if we remove `nativeSearch` from typings we should remove it also here.

mihaisc:
Tested market locally on ropsten, all good

mihaisc (approved):

PR #683, opened 2021-03-23
don't sleep, wait for aqua instead

alexcos20:
When creating assets, don't sleep, but check aqua from time to time until we find the asset


test/integration/ComputeFlow.test.ts
```diff
@@ -16,6 +16,7 @@ import { Account, DDO, Metadata } from '../../src/lib'
 import { Cluster, Container, Server } from '../../src/ocean/Compute'
 import { LoggerInstance } from '../../src/utils'
 import { ComputeInput } from '../../src/ocean/interfaces/ComputeInput'
+const fetch = require('node-fetch')
```
kremalicious, line 19: no mixing of `require` & `import`. This should be:

```suggestion
import fetch from 'node-fetch'
```

kremalicious:
I think this is the approximate time difference, this looks great!

Screen Shot 2021-03-24 at 11 50 42

kremalicious (approved):
should be good, but no clue why suddenly one Travis run starts to fail. I just restarted that test run

alexcos20:
looks like a provider change.  I will update this

test/integration/ComputeFlow.test.ts
```diff
@@ -16,6 +16,7 @@ import { Account, DDO, Metadata } from '../../src/lib'
 import { Cluster, Container, Server } from '../../src/ocean/Compute'
 import { LoggerInstance } from '../../src/utils'
 import { ComputeInput } from '../../src/ocean/interfaces/ComputeInput'
+import fetch from 'node-fetch'
```
jamiehewitt15, line 19: I noticed that you've added in `node-fetch` here which we have been removing in the other pull request. It might be better to use `cross-fetch` here instead?

jamiehewitt15 (changes requested):

alexcos20:
refactored in #691

PR #685, opened 2021-03-23
Add remove ocean liquidity function

### LINKED ISSUE #684: Remove ocean liquidity
We need another function similar to `removeOceanLiquidity` with the following params :
``` 
account: string,
poolAddress: string,
poolShares: string,
minimumOceanReceived: string
```
This function probably needs to use `exitswapPoolAmountIn` instead of `exitswapPoolAmountOut`

mihaisc:
Closes #684 
Suggestions for the function name are appreciated

### DIFF
--- a/src/balancer/OceanPool.ts
+++ b/src/balancer/OceanPool.ts
@@ -832,6 +832,7 @@ export class OceanPool extends Pool {
    * @param {String} account
    * @param {String} poolAddress
    * @param {String} amount Ocean Token amount in OCEAN
+   * @param {String} minOcean minimum amount of OCEAN received
    * @return {TransactionReceipt}
    */
   public async removeOceanLiquidityWithMinimum(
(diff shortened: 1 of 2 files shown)


kremalicious:
not fully sure about the implications but feels like we could make `removeOceanLiquidityWithMinimum` the new `removeOceanLiquidity`. I don't see a reason why `removePoolLiquidity` & `removeOceanLiquidity` should be so different, and seems what is in `removeOceanLiquidityWithMinimum` would make it consistent.

If we go with additional method, then we should Mae distinction between `removePoolLiquidity`/`removeOceanLiquidity`/`removeOceanLiquidityWithMinimum` also more clear in the JSDoc description, e.g. add `..., based on minimum amount of OCEAN to receive.`

mihaisc:
I agree that the current one doesn't suit any need. I opt for replacing it. @alexcos20 will this impact anything? From my knowledge this was made for the market more or less.

alexcos20:
Only market is using that, as far as I know.  Unless we have some other devs using ocean.js..

mihaisc:
So... do we all agree to change `removeOceanLiquidity` and `removeDTLiquidity` to accept `minOcean` and `minDt` instead of `maximumPoolShares` ?

alexcos20:
I would keep them both, because they have different usage

alexcos20 (approved):

Issue #678, opened 2021-03-22
Move from node-fetch to cross-fetch


alexcos20:
**Describe the bug**
Some of the methods exposed by node-fetch are not available in browser (most important, response.buffer())

Let's investigate if we can mitigate this by using cross-fetch (https://github.com/lquixada/cross-fetch)

jamiehewitt15:
Cross fetch looks like a good option for this, it is recommended by node-fetch for isomorphic usage. It also seems to be the most popular cross platform option. 

One question about this though, we are currently importing types from `node-fetch` (`BodyInit, RequestInit, Response`) and `cross-fetch` doesn't provide any types. Do we want to continue importing the types from `node-fetch` or shall we define them ourselves?

PR #687, opened 2021-03-24
Replacing node-fetch with cross-fetch

jamiehewitt15:
Closes: #678 

Changes proposed in this PR:

- Changing `node-fetch` to `cross-fetch`


package.json
```diff
@@ -44,14 +44,15 @@
...
+    "cross-fetch": "^2.2.3",
     "crypto-js": "^4.0.0",
     "decimal.js": "^10.2.1",
     "fs": "0.0.1-security",
     "lzma": "^2.3.2",
     "node-fetch": "^2.6.1",
     "save-file": "^2.3.1",
     "uuid": "^8.3.2",
-    "web3": "^1.3.4",
+    "web3": "^1.2.3",
```
alexcos20, line 55: any reason for downgrading web3?

jamiehewitt15: That was a mistake, I'll correct that

alexcos20 (approved):

alexcos20:
cool, looks good.

@mihaisc - can you try running this with the market? just to be sure that be didn't broke any browser thing.

jamiehewitt15:
One question about this though, we are currently importing types from `node-fetch` (`BodyInit, RequestInit, Response`) and `cross-fetch` doesn't provide any types. Do we want to continue importing the types from `node-fetch` or shall we define them ourselves and remove the `node-fetch` dependency completely?

alexcos20:
@kremalicious, @mihaisc  ?  - i would say to get rid of node-fetch completely , but not sure if it's used in market

kremalicious:
there should be only one fetch library. It looks like `cross-fetch` is typed so we should be able to replace all used typings with theirs. Like https://github.com/lquixada/cross-fetch/blob/main/lib.fetch.d.ts

Also agree we need to test build process of market with this change, cause the SSR build is usually where the problems with node/browser fetch appear. But in theory it should work™, since dealing with differences of Node.js/Browser context is what those libraries are for

kremalicious:
dang, we really need to deal with those npm v6/v7 differences. Using npm v7 will upgrade the `lockfileVersion` to v2 in the `package-lock.json`, as has been done in this PR.

If we upgrade the `lockfileVersion` then we most likely have to adapt our `npm install` commands with `--legacy-peer-deps`. For now, if we do not want to deal with upgrading the lock file we all need to make sure to modify the `package-lock.json` only with npm v6 which comes with Node.js v14. For this PR this would mean doing an `npm install` with Node.js v14/npm v6 which should downgrade the `lockfileVersion` agin

jamiehewitt15:
I edited the pull request to import the types from 'cross-fetch/lib.fetch' but I'm getting various issues with that:

```
src/metadatacache/MetadataCache.ts(154,27): error TS2339: Property 'buffer' does not exist on type 'Response'.

src/ocean/utils/WebServiceConnector.ts(43,9): error TS2345: Argument of type '{ method: string; body: BodyInit; headers: any; timeout: number; }' is not assignable to parameter of type 'RequestInit'.

  Object literal may only specify known properties, and 'timeout' does not exist in type 'RequestInit'.

src/ocean/utils/WebServiceConnector.ts(48,9): error TS2345: Argument of type '{ method: string; timeout: number; }' is not assignable to parameter of type 'RequestInit'.

  Object literal may only specify known properties, and 'timeout' does not exist in type 'RequestInit'.

src/ocean/utils/WebServiceConnector.ts(59,7): error TS2345: Argument of type '{ method: string; headers: { 'Content-type': string; }; timeout: number; }' is not assignable to parameter of type 'RequestInit'.

  Object literal may only specify known properties, and 'timeout' does not exist in type 'RequestInit'.

src/ocean/utils/WebServiceConnector.ts(71,9): error TS2345: Argument of type '{ method: string; body: BodyInit; headers: { 'Content-type': string; }; timeout: number; }' is not assignable to parameter of type 'RequestInit'.

  Object literal may only specify known properties, and 'timeout' does not exist in type 'RequestInit'.

src/ocean/utils/WebServiceConnector.ts(79,9): error TS2345: Argument of type '{ method: string; headers: { 'Content-type': string; }; timeout: number; }' is not assignable to parameter of type 'RequestInit'.

  Object literal may only specify known properties, and 'timeout' does not exist in type 'RequestInit'.

src/ocean/utils/WebServiceConnector.ts(92,9): error TS2345: Argument of type '{ method: string; body: BodyInit; headers: { 'Content-type': string; }; timeout: number; }' is not assignable to parameter of type 'RequestInit'.

  Object literal may only specify known properties, and 'timeout' does not exist in type 'RequestInit'.

src/ocean/utils/WebServiceConnector.ts(100,9): error TS2345: Argument of type '{ method: string; headers: { 'Content-type': string; }; timeout: number; }' is not assignable to parameter of type 'RequestInit'.

  Object literal may only specify known properties, and 'timeout' does not exist in type 'RequestInit'.

src/ocean/utils/WebServiceConnector.ts(132,23): error TS2339: Property 'pipe' does not exist on type 'ReadableStream<Uint8Array>'.

src/ocean/utils/WebServiceConnector.ts(133,23): error TS2339: Property 'on' does not exist on type 'ReadableStream<Uint8Array>'.
```

jamiehewitt15:
I'm not exactly sure how to proceed with this, there are several issues that I'm getting with the types from `cross-fetch/lib.fetch`

1. The timeout parameter is not available. I could potentially remove the timeout parameters but I realise that would introduce a vulnerability. Another option would be to build our own timeout function.   
2. `.buffer()` isn't available. I can alternatively use `.arraryBuffer()`
3. Property `pipe` does not exist on type `ReadableStream<Uint8Array>`. Line 132 in src/ocean/utils/WebServiceConnector.ts
4. Property `on` does not exist on type `ReadableStream<Uint8Array>` Line 133 in src/ocean/utils/WebServiceConnector.ts

kremalicious:
for 1., not sure how we were able to use `timeout` but it has always been non-standard and by now is removed from `node-fetch`. At same time `cross-fetch` claims to being a proxy not modifying any underlying spec:

- https://github.com/node-fetch/node-fetch/blob/master/docs/v3-UPGRADE-GUIDE.md#the-timeout-option-was-removed
- https://github.com/lquixada/cross-fetch/issues/62

So [`AbortSignal`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) seems to be what we are looking for instead, above migration guide has code example with https://github.com/node-fetch/timeout-signal

For 3. & 4. not sure yet why those would not be available when `cross-fetch` claims to wrap around `node-fetch`, but there might be other ways to write a stream to a file

jamiehewitt15:
Ok thanks, I'll take a look at AboutSignal

kremalicious:
for 3., some other `pipe` thingies seem to be available:

Screen Shot 2021-03-25 at 10 29 38

jamiehewitt15:
Yeah, I had noticed that. They both still give errors though

Screenshot from 2021-03-25 12-34-32

kremalicious:
wait, still into code archeology, but `node-fetch` is not at v3 for its latest version and `cross-fetch` also installs that pre-v3 version https://github.com/lquixada/cross-fetch/blob/main/package.json#L69. So that upgrade guide I mentioned does not apply, although switching timeout could still be valid.

Still a bit baffling to me why `cross-fetch` seems to expose a different interface from `node-fetch`, like for `Response`. Because the functionality is actually there so it might be only a typing problem. Maybe the original idea of using the typings from node-fetch would prevent us from having to refactor some critical functionality

kremalicious:
Some more findings:

We can remove all typings imports for `Request`, `Response`, `BodyInit`, and everything else which is part of the Fetch API. Cause Node.js knows about them already

The `downloadFile` in `WebServiceConnector` seems to be the only problem we are having and I did this to workaround it and not having to refactor or continue figuring out why typings are different:

```bash
npm i -D @types/node-fetch
```

and in `WebServiceConnector` then:

```js
import { Response } from 'node-fetch'
```

jamiehewitt15:
Ok yeah, that's a good point. I agree it seems to be a problem with types. Using the types from `node-fetch` gives us a quick solution to these issues, as it was passing all the tests when I had implemented it that way. It also avoids having to refactor the code which could bring other issues.

src/ocean/utils/WebServiceConnector.ts
```diff
@@ -1,11 +1,12 @@
-import { BodyInit, RequestInit, Response } from 'node-fetch'
+import { BodyInit, RequestInit, Response } from 'cross-fetch/lib.fetch'
 import fs from 'fs'
 import { Logger } from '../../utils'
 import save from 'save-file'
 // import { createWriteStream } from 'streamsaver'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const fetch = require('node-fetch')
+const fetch = require('cross-fetch')
```
kremalicious, line 8: can't recall the exact reason why we specifically did a `require` here, but it was done a long time ago in squid.js. So we can go back to our used `import` pattern. So this should be instead:

```js
import { fetch } from 'cross-fetch'
```

Same for the `timeout-signal` import

jamiehewitt15: Ok sure

src/provider/Provider.ts
```diff
@@ -7,12 +7,12 @@ import { ComputeInput } from '../ocean/interfaces/ComputeInput'
...
 import { MetadataAlgorithm } from '../ddo/interfaces/MetadataAlgorithm'
 import { Versions } from '../ocean/Versions'
-import { Response } from 'node-fetch'
 import { DDO } from '../ddo/DDO'
+import { Response } from 'cross-fetch'
 import DID from '../ocean/DID'
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
-const fetch = require('node-fetch')
+const fetch = require('cross-fetch')
```
kremalicious, line 15: can be removed, there is no direct usage of `fetch` in this file

src/ocean/utils/WebServiceConnector.ts
jamiehewitt15: The `import` pattern doesn't work for `timeoutSignal`:
 `'timeoutSignal' can only be imported by using 'import timeoutSignal = require("timeout-signal")' or a default import.`

kremalicious:
as a side note, often typings from `@types/` packages can't be trusted. They are all interim solutions for libraries which do not include their typings in their main package. So `@types/` packages are always externally managed in the `DefinitelyTyped` repo, and can get out of sync with the actual code base https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node-fetch

Once node-fetch v3 is included in a new cross-fetch release we can revisit since they then include their own typings in the package

jamiehewitt15:
Ok yeah, that makes sense

src/ocean/utils/WebServiceConnector.ts
kremalicious: > ...or a default import.

yup, let's pick that :-) Library has only one default export so can only be used with a default import (instead of single import with `{ }`):

```js
import timeoutSignal from 'timeout-signal'
```

jamiehewitt15: The `timeoutSignal` is working fine as a default import. However,  when I used `import fetch from 'cross-fetch'` it gave me the following error:
```
Type 'Response' is missing the following properties from type 'Response': buffer, size, textConverted, timeout
```

I have changed it on the branch now to `const fetch = require('cross-fetch')` and it's currently working & passing the tests.

kremalicious: well, we might have found a reason for that `require` usage here, seems to be the only way to get typings right without having to refactor the `WebServiceConnector` or further investigate why the typings between node-fetch and cross-fetch are so different

jamiehewitt15: Yeah, it seems like it

mihaisc:
Implemented local version of timeout. The imported one was throwing `TypeError: AbortController is not a constructor`.  I believe it's from `const { AbortController } = require("abort-controller")`. In the local version we are using the native `AbortController`

mihaisc (approved):

jamiehewitt15:
Looks like it's all working now

PR #691, opened 2021-03-26
improve tests speed

alexcos20:
refactor of #683 

improve tests speed

### DIFF
--- a/test/integration/ComputeFlow.test.ts
+++ b/test/integration/ComputeFlow.test.ts
@@ -17,11 +17,29 @@ import { Cluster, Container, Server } from '../../src/ocean/Compute'
 import { LoggerInstance } from '../../src/utils'
 import { ComputeInput } from '../../src/ocean/interfaces/ComputeInput'
 const web3 = new Web3('http://127.0.0.1:8545')
+const fetch = require('cross-fetch')
 
 function sleep(ms: number) {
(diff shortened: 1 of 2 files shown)


test/integration/ComputeFlow.test.ts
```diff
@@ -267,7 +285,7 @@ describe('Compute flow', () => {
     assert(ddo.dataToken === tokenAddress, 'ddo.dataToken !== tokenAddress')
     const storeTx = await ocean.onChainMetadata.publish(ddo.id, ddo, alice.getId())
     assert(storeTx)
-    await sleep(aquaSleep)
+    await waitForAqua(ocean, ddo.id)
```
0x3bfc, line 288: Why we need to wait for aqua after the assert ? specially that nothing won't happen after waiting for aqua.

jamiehewitt15 (approved):
This has been a big improvement. Looks like the time for the tests has reduced from about 22mins to about 15mins

test/integration/ComputeFlow.test.ts
alexcos20: after publishing an asset, you are going to use it somehow.  But aqua is scanning the chain every 15-20 secs, so first we need to wait, to make sure that aqua has cached the asset

PR #694, opened 2021-03-29
use blob instead of buffer

alexcos20:
This will fix issues with encryptedDDO when running ocean.js in browser


kremalicious (approved):
not tested, but specific use case we do not user in our product so should be good

alexcos20:
tested with @mihaisc , looks good on browser

PR #697, opened 2021-03-30
Feature/issues 452

krisliew:
Fixes Market Issue [link 452](https://github.com/oceanprotocol/market/issues/452)
Related PR at market [link 463](https://github.com/oceanprotocol/market/pull/463)

Changes proposed in this PR:

- Added a new variable in DDO `isDisable`
- The default value of `isDisable` is set as `false` to ensure new published dataset can be consumed
- Updated `EditMetadata()` to include `isDisable` changes to the new DDO


kremalicious:
it's a change in the DDO structure itself which implies changes are first needed in aquarius, provider (if the disable should be actually "real", not just via UI), possibly [plecos](https://github.com/oceanprotocol/plecos) (DDO validation scheme). After that it can be added to the libraries, but then needs to be added for both, ocean.js & ocean.py. After this is all released and deployed, it can be implemented in market.

kremalicious:
and editing metadata can only be done for values within the metadata service in the DDO. So `ddo.isConsumeDisable` as the plan most likely will not work, as users can only update the actual metadata service. Whose exact structure is defined in OEP-08, which would need update too as this is the main spec for all library implementations: https://github.com/oceanprotocol/OEPs/tree/master/8/v0.5

As I said, small feature for only a small subset of users in the UI, but we have to change our whole stack first.

kremalicious:
this is the confusing PR. Could we get a merge or rebase with current `main`? And what's supposed to be different in this PR than https://github.com/oceanprotocol/ocean.js/pull/786 which seems to add the described changes from here too?

krisliew:
> this is the confusing PR. Could we get a merge or rebase with current `main`? And what's supposed to be different in this PR than #786 which seems to add the described changes from here too?

This is an older PR before #786 were proposed. I'll close this PR as it is a duplicate.

Issue #246, opened 2020-09-02
New build process


kremalicious:
Currently `tsc` is used to output only one big CommonJS ES5 file. This works but prevents projects using the library to do their own optimizations like tree shaking during their build process, creating a huge lib-js dependency file. To solve that we need multiple outputs:

- CommonJS module
- ES module
- UMD

Over on `react` we use [`microbundle`](https://github.com/developit/microbundle) to achieve that with minimal config.

Should also produce a file usable via unpkg.com

kremalicious:
In production, ocean.js bundle size minified is 6MB!

jamiehewitt15:
@kremalicious Do you suggest using microbundle for ocean.js as well or is there any reason why it wouldn't work?

jamiehewitt15:
I have implemented this with `microbundle` here: https://github.com/oceanprotocol/ocean.js/pull/698

kremalicious:
nice, yes, this would have been my first suggestion too, use `microbundle` and see what happens :-)

PR #698, opened 2021-03-30
Building CommonJS, ES and UMD modules with microbundle

jamiehewitt15:
Closes: https://github.com/oceanprotocol/ocean.js/issues/246

Changes proposed in this PR:
- Implementing a new build process with microbundle


alexcos20:
awesome !
@kremalicious @mihaisc - can we check the frontend ?

mihaisc:
Will do

jamiehewitt15:
By the way, `microbundle` gives the option of optimizing the build for either a node or web environment (the default is web), so I have also added in an extra script `build:node` for optimizing for a node environment

mihaisc:
I keep getting these, not sure yet what the problem is. In theory `utils` and `ConfigHelper` are exported 
image

jamiehewitt15:
Hmm ok that's odd, how can I replicated that? It all seems to work when running `npm run build` in the terminal

jamiehewitt15:
Thanks

jamiehewitt15:
@mihaisc @kremalicious do you have any idea what the problem here could be? I have played around quite a bit with the settings of microbundle (there aren't that many) any it doesn't seem to make any difference. I'm thinking I'll try with a different bundler

jamiehewitt15:
I've started looking into doing this with [rollup](https://rollupjs.org/) on a separate branch but so far there are still issues: https://github.com/oceanprotocol/ocean.js/pull/703

jamiehewitt15:
I figure out the issue with this, there are two problems:
1. `microbundle` bundles everything into one index.js file. While, within the market, there are many imports from the files within the `/utils` folder. 
2. `microbundle` only bundles up the modules that are actually used within ocean.js. So  for example, `ConfigHelperConfig` is defined and exported but never actually used anywhere so it gets ignored by `microbundle` and isn't included in the bundled up index.js file. 

If we want to use `microbundle` on it's own we would need to refactor the code in ocean.js and ocean market (changing the imports from `'@oceanprotocol/lib/dist/node/utils/'` to `'@oceanprotocol/lib'`). 

For the moment, I have got this working with ocean market by using both `tsc` and `microbundle`

jamiehewitt15:
I have refactored the market code so that there are no imports from the utils folder: https://github.com/oceanprotocol/market/pull/481

There is still the issue that some exports are not included in the bundle:

>     2. `microbundle` only bundles up the modules that are actually used within ocean.js. So  for example, `ConfigHelperConfig` is defined and exported but never actually used anywhere so it gets ignored by `microbundle` and isn't included in the bundled up index.js file.

This issue applies to:
`ConfigHelperConfig`
`ConfigHelperNetworkId`
`ConfigHelperNetworkName`
`LogLevel`

I'll work next on refactoring the ocean.js code to make sure all of the exports are included in the bundle.

jamiehewitt15:
This needs to be merged at the same time as the refactoring of the `/utils` imports in Ocean Market: https://github.com/oceanprotocol/market/pull/481

I have tested this in Ocean Market (using `npm link` and then building and running it) and it seems to all wok ok

mihaisc:
Tested it with the market pr and it works ok

mihaisc (approved):

PR #710, opened 2021-04-05
[KianYee] #459 enhance file info function

TSS-LauKY:
Requirement for https://github.com/oceanprotocol/market/issues/459 & https://github.com/oceanprotocol/market/pull/479.

Changes proposed in this PR:

- if arg is DID type, get provider endpoint from DDO
- if arg is string type, get provider endpoint from getFileinfoEndpoint()

### DIFF
--- a/src/provider/Provider.ts
+++ b/src/provider/Provider.ts
@@ -131,11 +131,22 @@ export class Provider extends Instantiable {
    */
   public async fileinfo(url: string | DID): Promise<File[]> {
     let args
+    let path
     const files: File[] = []
     if (url instanceof DID) {
+      const ddo = await this.ocean.assets.resolve(url.getDid())
+      const computeService = ddo?.findServiceByType('compute')
+      const acessService = ddo?.findServiceByType('access')
       args = { did: url.getDid() }
-    } else args = { url }
-    const path = this.getFileinfoEndpoint() ? this.getFileinfoEndpoint().urlPath : null
+      path = computeService
+        ? computeService.serviceEndpoint
+        : acessService.serviceEndpoint
+      path = path + '/api/v1/services/fileinfo'
+    } else {
+      args = { url }
+      path = this.getFileinfoEndpoint() ? this.getFileinfoEndpoint().urlPath : null
+    }
+
     if (!path) return null
     try {
       const response = await this.ocean.utils.fetch.post(path, JSON.stringify(args))
--- a/src/provider/Provider.ts
+++ b/src/provider/Provider.ts
@@ -495,6 +495,7 @@ export class Provider extends Instantiable {
 
   private async getServiceEndpoint(did: string): Promise<string> {
     const ddo = await this.ocean.assets.resolve(did)
+    if (!ddo) return
     const computeService = ddo?.findServiceByType('compute')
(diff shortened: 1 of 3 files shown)


alexcos20:
cool.  can you add some unit tests as well ?

TSS-LauKY:
I added a negative test, since we have cover positive test.

src/provider/Provider.ts
```diff
@@ -494,4 +492,23 @@ export class Provider extends Instantiable {
       return false
     }
   }
+
+  private async getServiceEndpoint(did: string): Promise<string> {
```
mihaisc, line 496: Why do we need this? We already have `getServiceEndpoints` that returns all the endpoints. This function seems to return the compute service if found, and if not return the access service. But what happens if you have both? This is kind of hardcoding. A better approach would be to add `serviceIndex` as a function argument and then check that service. The one who uses this function needs to decide what service he wants to use.

TSS-LauKY: Yes, we can change to serviceIndex. 

getServiceEndpoints will return the endpoint base on config, there is no issue if we want to check fileinfo by passing url.
If we need to check the fileinfo for existing DID, and the provider from DDO is not the same with endpoint from config (wallet connection), we are checking the wrong endpoint.

mihaisc: > getServiceEndpoints will return the endpoint base on config, there is no issue if we want to check fileinfo by passing url.

Yes, my bad.

I'm a little confused about the argument name `url`, isn't that a did? Or can you give me some examples of what url are you refering to?

TSS-LauKY: the argument url can be use as:

- file url: 

- mostly we are using for data publish

image

- DID:

- provider connectivity check once DDO is crated. in this case we need to get the provider from DDO instead for config.

image

TSS-LauKY: I'm not sure is it better to split this function into 2

- filelinfo: pass url as argument and check the connectivity base on endpoint from config and return File response
- isFileconsumable: pass did as argument and check the connectivity base on endpoint from DDO and return Boolean

mihaisc: sounds good, less confusing and more usability

TSS-LauKY: I'm going to create isFileConsumable then.
Is it better have 2 argument: did and serviceIndex?

test/integration/Marketplaceflow.test.ts
```diff
@@ -400,10 +400,12 @@ describe('Marketplace flow', () => {
   })
 
   it('Alice should check if her asset has valid url(s)', async () => {
```
mihaisc, line 402: update test case to something like '.... check if her asset is consumable' and the it's good to go

mihaisc (approved):
looks good

kremalicious:
worth noting that on `market` we do not use `ocean.provider.fileinfo()` but have our own helper method to hit the provider directly:
https://github.com/oceanprotocol/market/blob/main/src/utils/provider.ts

Using `ocean.provider.fileinfo()` will only work when user has a wallet connected which is a restriction because of ocean.js architecture we work around on market.

PR #711, opened 2021-04-05
add compute dataset/compute algo

alexcos20:
add one more compute test:
 - dataset with compute service
 - algo with compute service

(should be published by the same provider)


bogdanfazakas (approved):
LGTM

PR #717, opened 2021-04-06
Optimize compute job status calls

mihaisc:
- optimize status calls by passing ddo or service so ddo is not fetched multiple times.
- includes a small update needed after https://github.com/oceanprotocol/provider/issues/126  ( so this pr is blocked until )

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3778,9 +3778,9 @@
       }
     },
     "cross-fetch": {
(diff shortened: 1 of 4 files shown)


kremalicious (approved):
Great, should work better. Only improvement could be to throw errors instead of null returns but we're in a hurry. Not sure of side effects, and that's a problem we have in existing code base

PR #731, opened 2021-04-12
refactor compute

alexcos20:
WARNING! This is a breaking change!

- Refactor compute functions to use a new interface (ComputeAlgorithm) instead of passing 4-5 params to different functions.
- add new test to prevent running a computeJob is both asset/algo are having compute services and they are not served from the same provider


bogdanfazakas (approved):
LGTM, also tested a local build with the market and seems everything is ok with the proper changes made in market

kremalicious:
cool, only 1 merge conflict after the big merging spree. So PR needs a merge or rebase against `main` to resolve conflict

PR #751, opened 2021-04-16
add moonbeamalpha

alexcos20:


alexcos20:
@md00ux - give the OK signal once we have the backend deployed

alexcos20:
@kremalicious @mihaisc  - ready to go

mihaisc:
Will try and test it locally

mihaisc:
We need a new release of contracts first

jamiehewitt15:
Is there anything still blocking this?

kremalicious:
the main blocker is that it conflicts with our Compute-to-Data rollout because for that to go smoothly we need to lock backend components until that is rolled out, yet backend component changes are needed to progress further with moonbeam.

For now the next steps are:
1. the aquarius update required for this feature [has broken the price creation flow in market](https://github.com/oceanprotocol/market/issues/546), so we might have to roll back to a former version, which would make aquarius the blocker for this PR
2. if we fix everything with a new aquarius, we need new ocean.js release with this PR
3. on market, we need to find and modify all the places where we check for a specific network, which we do in some places for polygon. So all those places have to be adapted to be either more universal, or specifically include moonbeam too
4. finally, once a market PR is ready all flows need to be tested on moonbeam, from publishing, to price creation, to downloading, to compute, in all its variations. All the pool interactions too, basically everything where blockchain transactions are happening. This step of course could be jumped over for the demo alone, and the PR preview could be used

jamiehewitt15:
Thank you for explaining @kremalicious, it sounds like there is still quite a lot left to do.  Do you think it's possible it will be done for the run-through with the moonbeam team tomorrow?

mihaisc:
Yes, i was already on it

mihaisc:
> on market, we need to find and modify all the places where we check for a specific network, which we do in some places for polygon. So all those places have to be adapted to be either more universal, or specifically include moonbeam too

true, but don't actually care about those, since this wasn't in our actual planning, and this is a highly test network

mihaisc (approved):

mihaisc:
Tested, working, bugs/issues that were found are not related to this pr. Were related to other components

alexcos20:
so we can merge and release?

PR #781, opened 2021-04-30
lock ocean-contracts to 0.5.16

alexcos20:


kremalicious (approved):
Verified with fresh local install that indeed `v0.5.16` gets installed

PR #786, opened 2021-05-05
Add ocean.assets.isConsumable function

### LINKED ISSUE #759: Add ocean.assets.isConsumable
This function should check if the asset is consumable by :
- check isDisabled flag (it exists & true -> asset is not consumable)
- check file conectivity
- check file length (optional)
- check credentials

Return should be a interface, containing a status and a status_message: Eg
   0, All good
   1, Asset is disabled
   2. Asset is empty
   3, Connectivity check failure
   4, Credential missing from allow list
   5, Credential found on deny list

krisliew:
Fixes #759 

The function should check the data asset ddo if is consumable by:

- check isOrderDisabled flag in `status` object of ddo (if exists & true -> asset is not consumable)

Return is an interface, containing a status and a status_message: Eg
- 0, All good
- 1, Ordering this asset has been temporarily disabled by the publisher.

### DIFF
--- a/src/ddo/interfaces/Status.ts
+++ b/src/ddo/interfaces/Status.ts
@@ -1,0 +1,26 @@
+/**
+ * Status attributes of Assets Metadata.
+ * @see https://github.com/oceanprotocol/OEPs/tree/master/8
+ */
+export interface Status {
+  /**
+   * Use to flag unsuitable content. True by default. If it's false, the content must not be returned.
+   * @type {boolean}
+   * @example true
+   */
+  isListed?: boolean
+
+  /**
+   * Flag retired content. False by default. If it's true, the content may either not be returned, or returned with a note about retirement.
+   * @type {boolean}
+   * @example false
+   */
+  isRetired?: boolean
+
+  /**
+   * For temporarily disabling ordering assets, e.g. when file host is in maintenance. False by default. If it's true, no ordering of assets for download or compute should be allowed.
+   * @type {boolean}
+   * @example false
+   */
+  isOrderDisabled?: boolean
+}
--- a/src/ocean/Assets.ts
+++ b/src/ocean/Assets.ts
@@ -644,7 +644,6 @@ export class Assets extends Instantiable {
   /**
    *
    * @param {String} did
-   * @param {String} consumerAddress // Optional, to check allow/deny list
    * @return {Promise<Consumable>}
    */
   public async isConsumable(did: string, consumerAddress?: string): Promise<Consumable> {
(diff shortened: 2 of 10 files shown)


alexcos20 (changes requested):
Missing:
- order should check this flag and reject if needed
- tests

alexcos20 (changes requested):
Please add the following tests:
  - Alice should update her asset and set isDisable = true
  - Bob should no be able to consume Alice's dataset after that

kremalicious:
There is currently a proposal to make `isDisabled` into `metadata.curation.isOrderDisabled`:

- https://github.com/oceanprotocol/OEPs/pull/236
- https://github.com/oceanprotocol/ocean.js/pull/789

src/ocean/Assets.ts
```diff
@@ -638,4 +643,30 @@ export class Assets extends Instantiable {
...
+   * @return {Promise<Consumable>}
+   */
+  public async isConsumable(did: string): Promise<Consumable> {
+    const ddo = await this.resolve(did)
+    if (!ddo) return null
+
+    if (ddo.isDisable)
+      return {
+        status: 1,
+        message: 'Asset is disabled'
```
kremalicious, line 659: we should be more clear what exactly is disabled, and by whom. So maybe like so:

```suggestion
        message: 'Ordering this asset has been temporarily disabled by the publisher.'
```

src/ocean/Assets.ts
```diff
@@ -638,4 +643,30 @@ export class Assets extends Instantiable {
...
     return results
   }
+
+  /**
+   *
+   * @param {String} did
+   * @return {Promise<Consumable>}
+   */
+  public async isConsumable(did: string): Promise<Consumable> {
+    const ddo = await this.resolve(did)
```
kremalicious, line 653: from the client side, for the use case for disabling the buy button when user arrives at asset details page, it would help to not fetch the DDO here again cause in almost all use case we already have the DDO. So should either accept a DDO or DID. And when DDO is passed, it should not refetch the DDO.

krisliew:
Will need a rebase / merge against `main` after #789 is merged into `main`.

alexcos20 (approved):

alexcos20:
@kremalicious - this supersedes #789, so we can close that

still blocked by stable aqua, so don't merge it yet

src/ddo/interfaces/Status.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Status attributes of Assets Metadata.
+ * @see https://github.com/oceanprotocol/OEPs/tree/master/8
+ */
+export interface Status {
+  /**
+   * Use to flag unsuitable content. True by default. If it's false, the content must not be returned.
+   * @type {boolean}
+   * @example true
+   */
+  isListed?: boolean
```
mihaisc, line 11: from the comment this looks like `isInPurgatory`, or what is the purpose? @kremalicious @alexcos20

alexcos20: we decided for now that we are going to use only isOrderDisabled

PR #787, opened 2021-05-05
Feature/credentials support in asset

kianyee:
Part of https://github.com/oceanprotocol/multi-repo-issue/issues/98

Changes proposed in this PR:

- to support credentials based access

- Added Credential interface in DDO
- Added a update function in asset class with follow feature:
- add new credential detail object if credential type not exist
- update credential detail object
- remove credential detail object when passing empty list


alexcos20 (changes requested):
1. Where are the helper functions?
2. Where are the integration tests?

kremalicious:
so far PR does not include anything regarding credentials, but deals with `isDisabled` which is not mentioned in the linked issue it is claiming to solve, or in the PR description. If the goal of this PR is to add support for all the different issues at once please link them up, otherwise hard to review.

In general missing connecting the dots, which would become more clear with actual flow tests as Alex pointed out. What's the use case of that `isConsumable` method? Is it supposed to do all checks which can influence the consume status? Will it also check for low liquidity of attached pools which would also influence consume status? Where is it supposed to be used in the end? Within ocean.js? From outside by clients? This all has implications on its implementation, e.g. if method should fetch the DDO yet again as it does right now and so on. If the method only deals with credentials & `isDisabled` it seems more feasible to keep that separated, or at least deal with credentials in another single method.

Based on linked issue and PR description it should not affect this PR but since the `isDisabled` is added here too, worth noting there is currently a proposal to make `isDisabled` into `metadata.curation.isOrderDisabled`:

- https://github.com/oceanprotocol/OEPs/pull/236
- https://github.com/oceanprotocol/ocean.js/pull/789

alexcos20:
I'm going to do some updates

src/ddo/DDO.ts
```diff
@@ -62,6 +63,8 @@ export class DDO {
 
   public dataTokenInfo?: DataTokenInfo
 
+  public credential?: Credential
```
kremalicious, line 66:
```suggestion
  public credentials?: Credentials
```

kremalicious (changes requested):
first findings, there are a lot of naming typos and inconsistencies in this PR:

- when referring to the feature it should always be plural: "credentials"
- a single credential is just that, a "credential" (not a "CredentialDetail")
- further combinations like in method names "removeAllowCredentailDetail" should be named more clear and without typos e.g. "removeCredential" where allow & deny can be a param so 4 methods become only 2

alexcos20:
> first findings, there are a lot of naming typos and inconsistencies in this PR:
> 
> * when referring to the feature it should always be plural: "credentials" (although the feature seems to be access lists, I only see account Ids which are not credentials like a username/password)
> * a single credential is just that, a "credential" (not a "CredentialDetail")
> * further combinations like in method names "removeAllowCredentailDetail" should be named more clear and without typos e.g. "removeCredential" where allow & deny can be a param so 4 methods become only 2

yes, I'm going to fix all of that

also, order functions should check if you have access.  otherwise this PR is useless

alexcos20 (approved):

alexcos20:
closes #758

src/ddo/interfaces/Credentials.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export type CredentialAction = 'allow' | 'deny'
+
+export interface Credential {
+  type: CredentialType
+  value: string[]
+}
+
+export interface Credentials {
+  allow?: Credential[]
```
mihaisc, line 14: I don't really get how this works. You can have a credential in `allow` that has `CredentialType` `deny`.

alexcos20: no.  in allow you can have address | 3box.
same for deny

src/ocean/Assets.ts
```diff
@@ -274,6 +280,63 @@ export class Assets extends Instantiable {
     return ddo
   }
 
+  /**
+   * Update Credentail attribute in DDO
```
mihaisc, line 284: `Credential` not `Credentail`

src/ocean/Assets.ts
```diff
@@ -274,6 +280,63 @@ export class Assets extends Instantiable {
...
+   * @param  {ddo} DDO
+   * @param {credentialType} CredentialType e.g. address / credentail3Box
```
mihaisc, line 286: credentail3Box

src/ocean/AssetsCredential.ts
```diff
@@ -0,0 +1,169 @@
...
+  credentialAction: CredentialAction
+): DDO {
+  const exists = this.checkCredentialExist(
+    ddo.credentials,
+    credentialType,
+    credentialAction
+  )
+  if (credentialAction === 'allow') {
+    if (exists) {
+      ddo.credentials.allow.find((credentail) => {
```
mihaisc, line 106: a lot of tails,  credentail

mihaisc (approved):

PR #789, opened 2021-05-07
add new status structure

kremalicious:
This implements OEP-8 changes in data structure for the `curation` object.

See https://github.com/oceanprotocol/OEPs/pull/236


kremalicious:
as far as I see we also need to update `plecos` https://github.com/oceanprotocol/plecos/blob/develop/plecos/schemas/metadata_remote_v0_5.json#L467-L517

Because `curation` now becomes what we always called local metadata, meaning editable by user/publisher as opposed to remote meaning updated by Aquarius alone.

alexcos20:
issue already there:  https://github.com/oceanprotocol/plecos/issues/45

jamiehewitt15:
From @alexcos20's message on the OEP issue, `curation` needs to be changed to `status` 

https://github.com/oceanprotocol/OEPs/pull/236#issuecomment-836580582

jamiehewitt15:
@kremalicious I have changed 'Curation' to 'Status' according to this: https://github.com/oceanprotocol/OEPs/pull/236#issuecomment-836580582 

Is there much else that needs to be done on this? It is currently blocking:  https://github.com/oceanprotocol/ocean.js/pull/786

mihaisc:
looks good

jamiehewitt15:
Is this still blocked?

alexcos20:
@kremalicious - all changes are already in #786, so we can close this

PR #803, opened 2021-05-19
Fix/ Order method should throw error messages

### LINKED ISSUE #762: Order method should return error messages
Issue:
Order method returns `null` whenever an error occurs.

Fix:
Change the order method to return an `error message` when it occurs instead of `null` value, so that we know what fails wherever we use it.

bogdanfazakas:
Fixes #762 .

Changes proposed in this PR:

- throw error messages when order methods fail
- update integration tests


test/integration/ComputeFlow.test.ts
```diff
@@ -940,15 +940,18 @@ describe('Compute flow', () => {
...
+      await ocean.compute.orderAsset(
+        bob.getId(),
+        datasetNoRawAlgo.id,
+        service1.index,
+        algoDefinition,
+        null, // no marketplace fee
+        computeAddress // CtD is the consumer of the dataset
+      )
+    } catch (error) {
+      assert(error != null, 'Order should be fail.')
```
alexcos20, line 953:
```suggestion
      assert(error != null, 'Order should not fail.')
```

test/integration/ComputeFlow.test.ts
```diff
@@ -971,15 +974,18 @@ describe('Compute flow', () => {
...
+      await ocean.compute.orderAsset(
+        bob.getId(),
+        datasetWithTrustedAlgo.id,
+        service1.index,
+        algoDefinition,
+        null, // no marketplace fee
+        computeAddress // CtD is the consumer of the dataset
+      )
+    } catch (error) {
+      assert(error != null, 'Order should be fail.')
```
alexcos20, line 987:
```suggestion
      assert(error != null, 'Order should not fail.')
```

test/integration/ComputeFlow.test.ts
```diff
@@ -996,15 +1002,19 @@ describe('Compute flow', () => {
...
+      await ocean.compute.orderAsset(
+        bob.getId(),
+        ddo.id,
+        computeService.index,
+        algoDefinition,
+        null, // no marketplace fee
+        computeAddress // CtD is the consumer of the dataset
+      )
+    } catch (error) {
+      assert(error != null, 'Order should be fail.')
```
alexcos20, line 1016:
```suggestion
      assert(error != null, 'Order should not fail.')
```

test/integration/ComputeFlow.test.ts
```diff
@@ -1091,15 +1101,18 @@ describe('Compute flow', () => {
...
+      await ocean.compute.orderAsset(
+        bob.getId(),
+        ddo.id,
+        computeService.index,
+        algoDefinition,
+        null, // no marketplace fee
+        computeAddress // CtD is the consumer of the dataset
+      )
+    } catch (error) {
+      assert(error != null, 'Order should be fail.')
```
alexcos20, line 1114:
```suggestion
      assert(error != null, 'Order should not fail.')
```

test/integration/Marketplaceflow.test.ts
```diff
@@ -561,12 +561,12 @@ describe('Marketplace flow', () => {
...
-    const txid = await ocean.assets.order(
-      ddoWithBadUrl.id,
-      accessService.type,
-      bob.getId()
-    )
-    assert(txid === null)
+    try {
+      await ocean.assets.order(ddoWithBadUrl.id, accessService.type, bob.getId())
+    } catch (error) {
+      assert(error != null)
```
alexcos20, line 567: if order succeeds, your catch block is not going to be executed, so you will have no assertions failed.

alexcos20 (changes requested):

test/integration/Marketplaceflow.test.ts
bogdanfazakas: good point 👍  thanks, going to update this

test/integration/ComputeFlow.test.ts
bogdanfazakas: I think the message should still indicate that the order `'Order should fail.'` in this test case, but as with the test in the MarketplaceFlow i will add inside try block an assert to catch if the order succeeds and have an assertion failed.

alexcos20 (approved):

PR #805, opened 2021-05-20
remove all ddo price tests

alexcos20:
Integrate with upcoming stable aqua


bogdanfazakas (approved):
Looks good to me, wondering if we still need to have the tests for creating and updating pool and fre prices at all ?

alexcos20:
Yeah, it's nice to have them because they are testing internal ocean.js functions

Issue #788, opened 2021-05-06
Normalize all floating number, so conversion to wei will not fail anymore


alexcos20:
**Describe the bug**
Sometimes, we get from the market floating numbers with too many decimals, and ToWei fails.

**Screenshots**
[Screenshot](https://ibb.co/88wGvT9)

**Expected behavior**
We should fix the precision, so conversion to Wei always succeeds

AnaLoznianu:
Closing this issue as we have a major ocean.js refactor in pipe that will fix this.

PR #807, opened 2021-05-21
Reducing all floating number decimals to 18

jamiehewitt15:
Closes: https://github.com/oceanprotocol/ocean.js/issues/788

Changes proposed in this PR:

- 
- 
-


src/utils/Decimals.ts
```diff
@@ -0,0 +1,3 @@
+export default function reduceDecimals(number) {
+  return String(parseFloat(number).toFixed(18))
```
mihaisc, line 2: This is not ok, please check what we did in https://github.com/oceanprotocol/market/pull/586 . We should be consistent  all over the place. Of course , improvements are welcomed.

src/utils/Decimals.ts
```diff
@@ -0,0 +1,5 @@
+import Decimal from 'decimal.js'
+export default function reduceDecimals(number: string) {
```
alexcos20, line 2: Please add an extra param which defaults to 18.
It's going to be super useful in the future, when dealing with tokens that do not use 18 decimals

alexcos20 (changes requested):

PR #822, opened 2021-05-28
fix ddo encrypt for browsers

alexcos20:
Warning:  Needs aquarius >=2.2.12

### DIFF
--- a/src/metadatacache/MetadataCache.ts
+++ b/src/metadatacache/MetadataCache.ts
@@ -146,12 +146,12 @@ export class MetadataCache {
    * @return {Promise<String>} Hex encoded encrypted DDO.
    */
   public async encryptDDO(ddo: any): Promise<any> {
-    const fullUrl = `${this.url}/api/v1/aquarius/assets/ddo/encrypt`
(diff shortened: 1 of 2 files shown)


bogdanfazakas (approved):

src/metadatacache/MetadataCache.ts
```diff
@@ -146,12 +146,12 @@ export class MetadataCache {
    * @return {Promise<String>} Hex encoded encrypted DDO.
    */
   public async encryptDDO(ddo: any): Promise<any> {
-    const fullUrl = `${this.url}/api/v1/aquarius/assets/ddo/encrypt`
+    const fullUrl = `${this.url}/api/v1/aquarius/assets/ddo/encryptashex `
```
kremalicious, line 149: Ideally we we stick to one form of encryption with one `encrypt` endpoint in Aquarius instead of adding multiple endpoints for one use case. If we want to support multiple types of encryptions then we should stick to REST principles and use the same endpoint, but with e.g. a body param

But I guess it's yet another hack for one use case. At least we should type what this methods returns which seems to be now `Promise<string>`.

And what's the implications for users who have used this `encryptDDO` method before, is this a breaking change for them?

PR #828, opened 2021-06-03
 Add also DDO object param where DID param is available 

### LINKED ISSUE #777: All functions expecting a DID as param should allow a DDO object as well
Way too many assets.resolve calls when we usually have the ddo already

bogdanfazakas:
Fixes #777  .

Changes proposed in this PR:

- added DDO object param to methods that expect did and call  asset.resolve


alexcos20:
@bogdanfazakas - what if we are going to use the following flow:

1. Create a util function that takes asset DDO|string as param and returns did and ddo
```
async function assetResolve(asset: DDO | string){
  if asset is DDO
           return { asset.did, asset}
  else{
        ddo = await ocean.assets.resolve(asset)
        return(ddo.id,ddo)
  }
}
```

and everywhere in ocean.js where we need a ddo, we can do:
```
function someOceanJsFunction(asset: DDO|string , .......) {
        const {did, ddo}  = await ocean.assets.assetResolve(asset)
       // use did & ddo 
}
```

So, all ocean.js functions will accept asset: DDO | string and use that helper to get both DID and DDO

bogdanfazakas:
> async function assetResolve(asset: DDO | string){
>   if asset is DDO
>            return { asset.did, asset}
>   else{
>         ddo = await ocean.assets.resolve(asset)
>         return(ddo.id,ddo)
>   }
> }

Yep sure we can have this

mihaisc:
should we check for `DID` as well? we don't use it often but  it is used in some functions and people use it. Or if we don't plan on using `DID` i think maybe we should just remove it

alexcos20 (approved):

PR #847, opened 2021-06-15
Replace Travis with GitHub Actions

kremalicious:
- kick out Travis, move to GitHub Actions workflows
- new workflow: `CI`
  - runs lint, build, unit tests, integration tests jobs, all independent of each other
- new workflow: `publish`, triggered on tag creation independent of `CI` run
- test build against all OS, fix building on Windows
- separate unit & integration tests again. Rule of thumb: if test needs more than Ganache, it's an integration test
- total CI run time down from ~21 min. to ~14 min.

Screen Shot 2021-06-21 at 13 17 57


jamiehewitt15:
I noticed that the CI tests failed in https://github.com/oceanprotocol/ocean.js/pull/842, I guess that's related to what you are testing/fixing here?

kremalicious:
yup, Travis is getting on my nerves failing for no particular reason so quickly doing sprint to figure out how quickly I can switch to GitHub Actions :-)

mihaisc:
Btw there are multiple tests failing https://travis-ci.com/github/oceanprotocol/ocean.js/builds/229036382 , not sure what is the culprit here

kremalicious:
yeah it seems just random, we restarted runs multiple times and sometimes they succeeded and sometimes not. GitHub Actions would have great advantages including speed, made it so that publishing the npm package should be down from ~25min to ~3min and it's independent now

But for whatever reason the unit tests fail already on creating a datatoken https://github.com/oceanprotocol/ocean.js/runs/2838811916?check_suite_focus=true#step:11:175

Which made me believe the contracts in `address.json` can't be read out or are wrong but by now pretty sure they are written out as can be seen here https://github.com/oceanprotocol/ocean.js/runs/2838811916?check_suite_focus=true#step:9:966

alexcos20 (approved):
let's merge https://github.com/oceanprotocol/ocean.js/pull/855 into this and we are good to go

kremalicious:
interesting, so looks like the barge job does not "detach" if there's not something like `& cd ..` at the end. That was the advantage of using pure docker command as that has a detach option which barge does not

alexcos20:
Yeah, that is strange. but it works now :).

PR #855, opened 2021-06-22
use barge instead of ganache

alexcos20:
For consistency, let's use barge for unit tests as well
You can say that it's over engineered, but if we need to change the mnemonic, or activate/deactivate EIPS, we are going to do that in one repo (barge)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,30 +46,15 @@ jobs:
           key: ${{ runner.os }}-test-unit-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
           restore-keys: ${{ runner.os }}-test-unit-${{ env.cache-name }}-
 
-      # Env var expansion workaround
-      # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
-      - name: Set ADDRESS_FILE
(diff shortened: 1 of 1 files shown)


kremalicious:
please no. Please keep them separated so we do not wait for ages for every test. There is a reason I split them up and NOT using barge for the unit tests. If you need more than ganache for a test, it's an integration test,

alexcos20:
instead of running ganache:latest with some extra param, we run barge with ganache and without aqua/provider/es, etc..

.github/workflows/ci.yml
```diff
@@ -45,15 +45,34 @@ jobs:
           path: ~/.npm
           key: ${{ runner.os }}-test-unit-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
           restore-keys: ${{ runner.os }}-test-unit-${{ env.cache-name }}-
-      - run: npm ci
-      - run: npm run build:metadata
 
-      - name: Run Ganache
+      # Env var expansion workaround
+      # https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable
+      - name: Set ADDRESS_FILE
```
kremalicious, line 51: let's remove the `ADDRESS_FILE` job as it's not needed. Most likely the DcokerHub login won't be needed either

kremalicious:
alright, so Ganache with Barge starts up faster than Ganache directly from DockerHub so let's go with that

kremalicious (approved):

PR #919, opened 2021-07-21
add single metadataCacheUri for all networks

kremalicious:
* requires Aquarius [`v3.0.0+`](https://github.com/oceanprotocol/aquarius/releases/tag/v3.0.0)
* addressing https://github.com/oceanprotocol/multi-repo-issue/issues/67

This is a breaking change as it will automatically switch clients to use the new Aquarius

### DIFF
--- a/src/utils/ConfigHelper.ts
+++ b/src/utils/ConfigHelper.ts
@@ -16,8 +16,8 @@ export interface ConfigHelperConfig extends Config {
 const configHelperNetworksBase: ConfigHelperConfig = {
   networkId: null,
   network: 'unknown',
+  metadataCacheUri: 'https://aquarius.oceanprotocol.com',
   nodeUri: 'http://localhost:8545',
-  metadataCacheUri: 'http://127.0.0.1:5000',
   providerUri: 'http://127.0.0.1:8030',
   subgraphUri: null,
   explorerUri: null,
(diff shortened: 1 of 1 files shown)


src/utils/ConfigHelper.ts
```diff
@@ -102,7 +98,6 @@ export const configHelperNetworks: ConfigHelperConfig[] = [
     networkId: 2021000,
     network: 'gaiaxtestnet',
     nodeUri: 'https://gaia-x.rpc',
-    metadataCacheUri: 'https://aquarius.gaiaxtestnet.oceanprotocol.com',
```
mihaisc, line 105: not sure if gx testnet will be in main aqua as well, @alexcos20 can you confirm this?

alexcos20: not for the moment..
probably we will run a separate aquarius for this

kremalicious: so we can leave as is and deal with it once we know

kremalicious:
still undecided if we should use these values here over in market, which would define if we need this change here in a new ocean.js release before releasing multi-network in market. 

Current state is: over in https://github.com/oceanprotocol/market/pull/628 the `metadataCacheUri` is part of the app config and we assume it is not changing during the user session. So the source of truth for `metadataCacheUri` is now the market app config, and not ocean.js

kremalicious:
After some days playing with multi network PR codebase it's definitely way more comfortable on market to leave it as an app config value so it can be easily accessed everywhere. Needs to be revisited though once we know how other metadata cache URIs should behave.

KatunaNorbert (approved):

PR #944, opened 2021-07-31
Custom user parameters

### LINKED ISSUE #908: Pass through of algouserdata to provider
During C2D, some algorithms may require additional parameters from the consumer

After user input, those fields are going to look like an array of key => values, which has to be passed to the provider

### LINKED ISSUE #909: Pass through of userdata to provider
We need to be able to support:

- assets that require some input data from the consumer
- during C2D, some algorithms may require additional parameters from the consumer

After user input, those fields are going to look like an array of key => values

For each dataset, that array has to be passed to the provider

alexcos20:
Add creation of `userCustomParameters` & `algoCustomParameters` and passthrough of `userCustomParameters` to provider.

Closes #908
Closes #909 

Breaking changes !!!!
 - `ocean.assets.createAccessServiceAttributes`  accepts a new parameter (requiredData)
 - `ocean.compute.createComputeServiceAttributes`  accepts a new parameter (requiredData)
 - `ocean.assets.order` has an extra param, breaking compatibility !
 - `ocean.compute.orderAsset` has an extra param, breaking compatibility !
 - `ocean.compute.orderAlgorithm` has an extra param, breaking compatibility !
 - on market the UI needs to catch and send `UserCustomParameters` along with `order()`, and consume/compute UI outputs from `ServiceCustomParametersRequired`
 - deal with pre-releases in publish GitHub Action


src/ddo/interfaces/Service.ts
```diff
@@ -1,12 +1,27 @@
...
+  name: string
+  type: string
+  label: string
+  required: boolean
+  options?: any
+  description: string
+}
 
+export interface RequiredData {
+  userdata?: CustomData[]
```
kremalicious, line 14: camelCase consistency, so should be `userData` & `algoData`

src/ddo/interfaces/Service.ts
```diff
@@ -1,12 +1,27 @@
...
+  algodata?: CustomData[]
+}
+
+export type ServiceType = 'authorization' | 'metadata' | 'access' | 'compute'
 export interface ServiceCommonAttributes {
   main: { [key: string]: any }
   additionalInformation?: { [key: string]: any }
   status?: Status
+  userdata?: CustomData[]
```
kremalicious, line 23: duplication, `ServiceCommonAttributes` can extend `RequiredData` so we do not have to repeat this here:

```ts
export interface ServiceCommonAttributes extends RequiredData {
  ...
}
```

src/ddo/interfaces/Service.ts
kremalicious: also `data` is very generic in multiple places in this PR, can we come up with something more precise. Otherwise we end calling everything `data` even if we can be more precise like `parameters`, `options`, `metadata`, `user input` and so on

src/ocean/Assets.ts
```diff
@@ -517,6 +523,11 @@ export class Assets extends Instantiable {
         }
       }
     }
+    if (requiredData && requiredData.userdata)
```
kremalicious, line 526: we started to like that optional chaining in some other places, so this is the same but shorter:

```js
if (requiredData?.userdata)
```

src/ocean/Assets.ts
```diff
@@ -782,4 +801,25 @@ export class Assets extends Instantiable {
...
+   */
+  public async isCustomDataValid(
+    customDDOData: CustomData[],
+    userData?: { [key: string]: any }
+  ): Promise<boolean> {
+    if (customDDOData)
+      for (const data of customDDOData) {
+        const keyname = data.name
+        if (!userData || !userData[keyname]) {
+          console.log('Missing key: ' + keyname + ' from customData')
```
kremalicious, line 819: Should use our `Logger` instead

test/integration/ComputeFlow.test.ts
```diff
@@ -826,6 +930,13 @@ describe('Compute flow', () => {
         const balance = await datatoken.balance(tokenAddressAdditional2, bob.getId())
         assert(balance.toString() === dTamount.toString())
       })
+
+    await datatoken
+      .transfer(tokenAddressWithCustomData, bob.getId(), dTamount, alice.getId())
+      .then(async () => {
```
kremalicious, line 936: not fully sure this has any real implication, but the mix of async/await with promise.then seems unnecessary. This is the same following the established async/await syntax in this file:

```js
await datatoken.transfer(tokenAddressWithCustomData, bob.getId(), dTamount, alice.getId())
const balance = await datatoken.balance(tokenAddressWithCustomData, bob.getId())
assert(balance.toString() === dTamount.toString())
```

test/integration/Marketplaceflow.test.ts
```diff
@@ -480,6 +555,12 @@ describe('Marketplace flow', () => {
         const balance = await datatoken.balance(tokenAddressForBadUrlAsset, bob.getId())
         assert(balance.toString() === dTamount.toString())
       })
+    await datatoken
+      .transfer(tokenAddressWithUserData, bob.getId(), dTamount, alice.getId())
+      .then(async () => {
```
kremalicious, line 560: same here with async/await

src/ddo/interfaces/Service.ts
mihaisc: `userCustomParameters` would be more precise i think, since this is what their purpose is.

kremalicious (approved):

mihaisc (approved):

PR #983, opened 2021-08-23
add compute getResult

### LINKED ISSUE #982: Secure C2D Outputs
Right now, the result of a C2D data (outputs/algo logs) are sent as a URL to the consumer.
Which means that anyone who has those URL has access to them, even if he/she is not the actual consumer.

Proposal:

- GET /compute    (get job results) hides the urls, exposing only an array of output file names, size & type
```json
[
  {
    "agreementId": "0x111111",
    "algoDID": "did:op:87bdaabb33354d2eb014af5091c604fb4b0f67dc6cca4d18a96547bffdc27bcf",
    "dateCreated": 1629281470.08931,
    "dateFinished": 1629281491.05138,
    "inputDID": [
      "did:op:1384941e6f0b46299b6e515723df3d8e8e5d1fb175554467a1cb7bc613f5c72e"
    ],
    "jobId": "6951125396154f2da719f097ae758bb5",
    "owner": "0xC41808BBef371AD5CFc76466dDF9dEe228d2BdAA",
    "removed": 0,
    "results": [
      {
        "filename": "output.log",
        "filesize": 85,
        "type": "output"
      },
      {
        "filename": "algorithm.log",
        "filesize": 127,
        "type": "algorithmLog"
      }
    ],
    "status": 70,
    "statusText": "Job finished",
    "stopreq": 0
  }
]
```

Provider exposes new endpoint:   GET /computeResult, which retrieves an output from a job (index based)
```
parameters:
      - name: consumerAddress
        in: query
        description: The consumer address.
        required: true
        type: string
      - name: jobId
        in: query
        description: JobId
        required: true
        type: string
      - name: index
        in: query
        description: Result index
        required: true
      - name: signature
        in: query
        description: Signature of (jobId+index+consumerAddress+nonce) to verify that the consumer has rights to download result
    """
(see https://github.com/oceanprotocol/provider/pull/201)


Ocean.js should use that endpoint

### PR DESCRIPTION
**Pull request by alexcos20**
Closes #982

1. Removes signature parameter of ocean.compute.status  (not used anymore)
2. Add ocean.compute.getResult to get the file content of index X
3. ocean.provider exposes a computeLimits object defined as:

export interface ComputeLimits {
  algoTimeLimit?: string
  storageExpiry?: string
}
(both expressed in seconds)

this allows the market to expose computeLimits to users

### CODE REVIEW

**Comment by alexcos20** [comment]
Please DO NOT merge, needs updated versions of provider/C2D components.
After approval, we are going to deploy all new components at a specific time

**Comment by mihaisc** [comment]
what is `algoTimeLimit` ?

**Comment by alexcos20** [comment]
> what is `algoTimeLimit` ?

Maximum amount of time that an algorithm is allowed to run in a C2d env.  If service timeout > algoTimeLimit, then algoTimeLimit is enforced

File: src/provider/Provider.ts [hunk 1]
```diff
@@ -447,6 +443,40 @@ export class Provider extends Instantiable {
...
+    try {
+      !destination
+        ? await this.ocean.utils.fetch.downloadFileBrowser(consumeUrl)
+        : await this.ocean.utils.fetch.downloadFile(consumeUrl, destination, index)
+    } catch (e) {
+      this.logger.error('Error getting job result')
+      this.logger.error(e)
+      throw e
+    }
+    return destination
```
mihaisc, line 477: what's the point of returning this? You have destination as a parameter and then it's never reassigned

test/integration/ComputeFlow.test.ts
```diff
@@ -917,6 +914,33 @@ describe('Compute flow', () => {
...
+      0,
+      './node_modules/jobs-results/' + jobId + '/',
+      ddo.id,
+      undefined,
+      undefined
+    )
+  })
   it('Bob should stop compute job', async () => {
     assert(jobId != null, 'Jobid is null')
     await ocean.compute.stop(bob, ddo, jobId)
```
mihaisc, line 946: i just saw that we have a stop function. I didn't find any resume function, what is the point of stop?

src/provider/Provider.ts
alexcos20: yeah, it's overkill.  any result !== null means that the download succeeded

test/integration/ComputeFlow.test.ts
alexcos20: There is no resume function. And stop could be used if, as a datascientist, you realize that your algo, which takes 3 days to complete, has some errors and there is no point of running it.
In the future, we will have real time logs

mihaisc (approved):

src/ocean/Compute.ts
```diff
@@ -244,39 +243,60 @@ export class Compute extends Instantiable {
...
+   * Downloads compute job result
    * @param  {Account} consumerAccount The account of the consumer ordering the service.
-   * @param  {DDO|string} asset DID Descriptor Object containing all the data related to an asset or a Decentralized identifier.
-   * @param  {string} jobId The ID of the compute job to be stopped.
-   * @return {Promise<ComputeJob>} Returns the DDO of the result asset.
+   * @param  {string} jobId JobId
+   * @param  {number} index Compute Result index
+   * @param  {string} destination: destination folder
+   * @param  {string} did Decentralized identifier.
+   * @param  {DDO} ddo If undefined then the ddo will be fetched by did, this is just to optimize network calls
```
kremalicious, line 259: so sounds like a param like `asset: DDO | DID | string` and then check what you get passed would be nicer pattern here to limit parameters

src/ocean/Compute.ts
```diff
@@ -244,39 +243,60 @@ export class Compute extends Instantiable {
...
+   * @param  {ServiceCompute} service If undefined then we get the service from the ddo
```
kremalicious, line 260: where else would we get it from? Why can another compute service be passed to this method which seems to be then different from the DDO compute service? What's the use case for this?

src/ocean/Compute.ts
```diff
@@ -244,39 +243,60 @@ export class Compute extends Instantiable {
...
   /**
-   * Returns the final result of a specific compute job published as an asset.
+   * Downloads compute job result
    * @param  {Account} consumerAccount The account of the consumer ordering the service.
-   * @param  {DDO|string} asset DID Descriptor Object containing all the data related to an asset or a Decentralized identifier.
-   * @param  {string} jobId The ID of the compute job to be stopped.
-   * @return {Promise<ComputeJob>} Returns the DDO of the result asset.
+   * @param  {string} jobId JobId
+   * @param  {number} index Compute Result index
+   * @param  {string} destination: destination folder
```
kremalicious, line 257: What destination folder is meant here? Where is this folder created? We should have a default here, or have this not as a param at all.

src/provider/Provider.ts
```diff
@@ -460,6 +456,40 @@ export class Provider extends Instantiable {
     }
   }
 
+  public async computeResult(
+    jobId: string,
+    index: number,
+    destination: string,
+    account: Account
+  ): Promise<any> {
```
kremalicious, line 464: based on typings and underlying code this method returns a `string`

src/ocean/Compute.ts
```diff
@@ -244,39 +243,60 @@ export class Compute extends Instantiable {
...
-    const computeJobsList = await provider.computeStatus(
-      did,
-      consumerAccount,
+    jobId: string,
+    index: number,
+    destination: string,
+    did?: string,
+    ddo?: DDO,
+    service?: ServiceCompute
+  ): Promise<any> {
```
kremalicious, line 270: based on underlying typing and code of `provider.computeResult()` this returns a `Promise<string>`

kremalicious:
So main use case to me would be: "I have a `jobId` and a `ddo` and I want to get all results for it." How do I know how many `index` I have to iterate through?

src/ocean/Compute.ts
mihaisc: it's for use cases where you already have the service, but not the ddo. This was actually from the market

src/ocean/Compute.ts
mihaisc: it's for when you use it on a server. This is how it's always been, no need to refactor now just for the sake of refactoring.

src/ocean/Compute.ts
```diff
@@ -244,39 +243,60 @@ export class Compute extends Instantiable {
...
-  ): Promise<ComputeJob> {
-    const { did, ddo } = await assetResolve(asset, this.ocean)
-    const service = ddo.findServiceByType('compute')
-    const { serviceEndpoint } = service
-    const provider = await Provider.getInstance(this.instanceConfig)
-    await provider.setBaseUrl(serviceEndpoint)
-    const computeJobsList = await provider.computeStatus(
-      did,
-      consumerAccount,
+    jobId: string,
```
mihaisc, line 264: we don't need `jobId` here, and the returned value here should be an array right?

mihaisc:
@kremalicious  the flow is the following 
- call [`status`](https://github.com/oceanprotocol/ocean.js/blob/ee75240b7a15c50917700345dbdf0ae64ec84fc5/src/ocean/Compute.ts#L213) get a list of `ComputeJob` 
- display list of jobs like we do now, but you don't need to sign to get this list ( you currently do)
image
- when you click on a result you call [`getResult`](https://github.com/oceanprotocol/ocean.js/blob/ee75240b7a15c50917700345dbdf0ae64ec84fc5/src/ocean/Compute.ts#L262) with the jobId that you have and did  (or ddo , or service) . You will need to sign and after that the download will begin.

kremalicious:
perfect, thanks for explanation! So that's already good guidance on what needs to be done over in market for this

mihaisc:
Already linked this in the issue

src/ocean/Compute.ts
kremalicious: If I see this right, we always have full DDOs in this `getComputeJobs` method:
https://github.com/oceanprotocol/market/blob/main/src/utils/compute.ts#L257

we just don't pass it as part of `ComputeJobMetaData` but we actually could:
https://github.com/oceanprotocol/market/blob/main/src/utils/compute.ts#L188

Cause then we could remove that passing only a service use case from here. By always requiring passing of an `asset` which then is either a `DDO` or just `string/DID` we have made 3 parameters into 1 too

PR #989, opened 2021-08-27
Clean up transferOwnership

kremalicious:
Reduce all `transferOwnership` methods to a single one.

- `ocean.assets.transferOwnership` as main interface
- shortcut to `ocean.OnChainMetadata.transferOwnership` which updates the owner field in DDO on-chain, creating a `MetadataUpdate` event in the process
- add an integration test for the whole thing (WIP)
- based on findings in https://github.com/oceanprotocol/privateissues/issues/36


kremalicious:
not worth the effort before v4, and #1030 already removed the confusing methods

PR #1012, opened 2021-09-09
Fix token amount - token allowance comparison

### LINKED ISSUE #1008: Wrong token amount conversion
There is a problem in converting the token `amount` to Wei using `this.web3.utils.toWei()` function in `addOceanLiquidity()` and `addDTLiquidity()` when sending it as a parameter to the `approve()` method.  

Inside the `approve()` function there is a check if the number of allowed tokens is grater then the amount of tokens to add as liquidity and although the `amount` is for example: `4` or `1` after conversion it ends up as `4000000000000000000` or `1000000000000000000` resulting a wrong comparison result.

An easy fix for this could be to rather than remove the `amount` conversion to Wei in all the methods, to convert the `allowance` value to Wei inside the `approve(`) method

bogdanfazakas:
Fixes #1008  .

Changes proposed in this PR:

- fixed allowance check in approve 
- added allowance check in buyDT/buyDTWithExactOcean/addDTLiquidity/addOceanLiquidity methods so we do not call approve if not necessary


src/balancer/OceanPool.ts
```diff
@@ -800,15 +810,18 @@ export class OceanPool extends Pool {
...
-      account,
-      this.oceanAddress,
-      poolAddress,
-      this.web3.utils.toWei(amount)
-    )
-    if (!txid) {
-      this.logger.error('ERROR: OCEAN approve failed')
-      return null
+    const oceanAllowance = await this.allowance(this.oceanAddress, account, poolAddress)
+    if (new Decimal(oceanAllowance).lt(amount)) {
```
KatunaNorbert, line 814: Not sure if we need to make this check here as well, isn't the check inside the approve method enough? Now when there are not enough tokens approved we double check

alexcos20:
I will clean it up

src/balancer/OceanPool.ts
bogdanfazakas: The need of checking twice came up because the approve method returns `null` if the allowance is already bigger that the amount needed, and in the upper methods we if we call approve we are expecting a `txid` so we can proceed with the logic. The approve method also returns a null response in case the approve fails and we check if the txid is received from the approve method. So we check the allowance in the upper methods also and call approve only when necessary. In my opinion we can remove the check in approve method or we can return something different than null if the allowance is bigger than the amount.

alexcos20 (approved):
lgtm

PR #1030, opened 2021-09-20
remove aqua queries

### LINKED ISSUE #1033: Make ocean.js compatible with Aquarius 3.1.x
Query format has been changed, so we need to update ocean.js
Also, do some cleanup on metadatacache part

alexcos20:
Closes #1033 

Changes proposed in this PR:

- remove interface QueryResult 
- update interface SearchQuery 
- remove ocean.metadataCache.transferOwnership  (no longer used)
- remove ocean.metadataCache.getOwnerAssets (never used)
- remove ocean.metadataCache.retire  (no longer used)
- remove ocean.assets.ownerAssets
- remove ocean.assets.resolveByDTAddress
- remove ocean.assets.query
- remove ocean.assets.search
- move ocean.metadatacache tests from unit to integration

Labeled: Breaking because needs Aqua 3.1.0 or later.  

PS: And since we are using Aqua 3.1.0, which has the events object in the ddo, a lot of improvements in the tests speed were possible


src/metadatacache/MetadataCache.ts
```diff
@@ -1,23 +1,16 @@
 import { DDO } from '../ddo/DDO'
 import DID from '../ocean/DID'
-import { EditableMetadata } from '../ddo/interfaces/EditableMetadata'
 import { Logger, isDdo } from '../utils'
 import { WebServiceConnector } from '../ocean/utils/WebServiceConnector'
 import { Response } from 'node-fetch'
 import { Metadata, ValidateMetadata } from '../ddo/interfaces'
+const fetchLibrary = require('cross-fetch')
```
kremalicious, line 7: was there a specific reason why we mix `import` and old `require` here?

src/metadatacache/MetadataCache.ts
```diff
@@ -338,19 +207,39 @@ export class MetadataCache {
...
-    return {
-      results: (results || []).map((ddo: DDO) => new DDO(ddo as DDO)),
-      page,
-      totalPages,
-      totalResults
-    }
+  /**
+   * Simple blocking sleep function
+   */
+  public sleep(ms: number) {
```
kremalicious, line 213: what's the purpose of this, how are users supposed to use this? If it's only for testing purposes it should not exist here but in some test helper file. Same for `waitForAqua()`

.github/workflows/ci.yml
```diff
@@ -106,6 +106,8 @@ jobs:
 
       - name: Run Barge
         working-directory: ${{ github.workspace }}/barge
+        env:
+          AQUARIUS_VERSION: v3.1.1
```
kremalicious, line 110: ideally we have this as default in barge and not hardcode specific versions in here before merging

kremalicious:
if we remove all the `ocean.assets.*` methods then even less clear what the difference between `ocean.assets.*` & `ocean.metadataCache.*` is, and we should either decide on one, or use one of it as a shortcut. E.g. over in #989 I use `ocean.assets.*` as a shortcut for methods https://github.com/oceanprotocol/ocean.js/pull/989/files#diff-193f6dfec7a03d2b356809b8e23c7ccd3790c854365aef4a58b96bdcb05b4c30R788-R808

So, what is the purpose of `ocean.assets.*` now?

src/metadatacache/MetadataCache.ts
```diff
@@ -338,19 +207,39 @@ export class MetadataCache {
...
+    })
+  }
+
+  /**
+   * Blocks until Aqua will cache the did (or the update for that did) or timeouts
+   * @param  {string} did DID of the asset.
+   * @param  {string} txid used when the did exists and we expect an update with that txid.
+   * @return {Promise<DDO>} DDO of the asset.
+   */
+  public async waitForAqua(did: string, txid?: string) {
```
mihaisc, line 225: why do you have this here? looks like a function used just in testing.

src/metadatacache/MetadataCache.ts
```diff
@@ -1,23 +1,16 @@
...
-import { EditableMetadata } from '../ddo/interfaces/EditableMetadata'
 import { Logger, isDdo } from '../utils'
 import { WebServiceConnector } from '../ocean/utils/WebServiceConnector'
 import { Response } from 'node-fetch'
 import { Metadata, ValidateMetadata } from '../ddo/interfaces'
+const fetchLibrary = require('cross-fetch')
 
 const apiPath = '/api/v1/aquarius/assets/ddo'
 
-export interface QueryResult {
```
kremalicious, line 11: so, how do we now know outside of ocean.js how the search result is structured? What's the replacement for `QueryResult` in market then?

src/metadatacache/MetadataCache.ts
```diff
@@ -93,49 +86,22 @@ export class MetadataCache {
    * @param  {SearchQuery} query Query to filter the DDOs.
    * @return {Promise<QueryResult>}
    */
-  public async queryMetadata(query: SearchQuery): Promise<QueryResult> {
-    const result: QueryResult = await this.fetch
-      .post(`${this.url}${apiPath}/query`, JSON.stringify(query))
+  public async queryMetadata(query: SearchQuery): Promise<any> {
```
kremalicious, line 89: We really don't know what data structure Aquarius will return on queries anymore?

src/metadatacache/MetadataCache.ts
```diff
@@ -93,49 +86,22 @@ export class MetadataCache {
...
+    const result = await this.fetch
+      .post(`${this.url}/api/v1/aquarius/assets/query`, JSON.stringify(query))
```
kremalicious, line 91: should work with the `apiPath` const defined at top of file, otherwise defining that as const becomes redundant if not used whenever we have an API path

src/metadatacache/MetadataCache.ts
mihaisc: it's elasticseach result. not sure if we should do anything regarding it. That function is more or less useless anyway, people will just directly call aquarius.

mihaisc:
Agree having `MetadataCache` and `Assets` is confusing and looking at the functions in them not sure exactly what handles each.

src/metadatacache/MetadataCache.ts
mihaisc: https://github.com/oceanprotocol/market/issues/838 + more discussion on the thread bellow , so we can keep it in one place since it's the same topic.

src/metadatacache/MetadataCache.ts
alexcos20: we could use waitForAqua in the market as well, when publishing, instead of just doing a pause..

.github/workflows/ci.yml
alexcos20: yes, once this is merged, we will update barge and remove it
If we update barge, all ocean.js tests will fail now

src/metadatacache/MetadataCache.ts
alexcos20: we could use waitForAqua in the market as well, when publishing, instead of just doing a pause..

src/metadatacache/MetadataCache.ts
alexcos20: yeah, somebody thought than instead of /api/v1/aquarius/assets/ddo/query to use /api/v1/aquarius/assets/query :(

alexcos20:
> Agree having `MetadataCache` and `Assets` is confusing and looking at the functions in them not sure exactly what handles each.

I will refactor a bit

alexcos20:
> if we remove all the `ocean.assets.*` methods then even less clear what the difference between `ocean.assets.*` & `ocean.metadataCache.*` is, and we should either decide on one, or use one of it as a shortcut. E.g. over in #989 I use `ocean.assets.*` as a shortcut for methods https://github.com/oceanprotocol/ocean.js/pull/989/files#diff-193f6dfec7a03d2b356809b8e23c7ccd3790c854365aef4a58b96bdcb05b4c30R788-R808 but this is kinda only done because we have this weird split-up.
> 
> So, what is the purpose of `ocean.assets.*` now? Also `ocean.assets.*` sounds way more user friendly and we should rather work with that as main interface, e.g. `ocean.assets.query()` is way better than `ocean.metadataCache.queryMetadata()`, and it mirrors the actual API path, which always has `assets` in there

ocean.assets.query is back

alexcos20:
@kremalicious @mihaisc - any more thoughts?

mihaisc (approved):
good enough, not worth going in too deep at this point

anudit:
Hey @kremalicious, how do I get owner assets now that `ownerAssets()` is removed ?

PR #1054, opened 2021-10-01
bump contracts & add more networks

### LINKED ISSUE #1053: Add support for Moonriver & Energy Web
Contracts were deployed to new networks with https://github.com/oceanprotocol/contracts/pull/351 & https://github.com/oceanprotocol/contracts/pull/352:

- Moonriver (production network)
- Energy Web (production network)
- Celo Alfajores (test network)

Add all of them to the `ConfigHelper`, Celo is optional as it's a test network but if we are deployed to it then we should include it.

alexcos20:
Closes #1053


mihaisc:
but why the need to change bases for some random reason? can we be consistent? will the next chain id be in base64?

alexcos20:
> please use the same base, it is very confusing to change bases randomly

fixed

package.json
```diff
@@ -42,11 +42,11 @@
   },
   "homepage": "https://github.com/oceanprotocol/ocean.js#readme",
   "peerDependencies": {
-    "web3": ">=1.3.5"
+    "web3": "^1.6.0"
```
kremalicious, line 45: Something went wrong on this PR with the code comments. So this change for me was shown within our `dependencies` Where I then said it’s too adventurous. For the `peerDependencies` This range should stay as is though, so we do not force everybody to upgrade their web3.js, that’s why we used `>=` for this which is fine, but only for `peerDependencies`

src/utils/ConfigHelper.ts
```diff
@@ -129,6 +129,33 @@ export const configHelperNetworks: ConfigHelperConfig[] = [
...
+    nodeUri: 'https://rpc.energyweb.org',
+    providerUri: 'https://provider.energyweb.oceanprotocol.com',
+    subgraphUri: 'https://subgraph.energyweb.oceanprotocol.com',
+    explorerUri: 'https://explorer.energyweb.org'
+  },
+  {
+    ...configHelperNetworksBase,
+    networkId: 1285,
+    network: 'moonriver',
+    nodeUri: 'https://moonriver.api.onfinality.io/public',
```
kremalicious, line 155: That's a 3rd party service. Does it work without providing an API key or does market have to inject that in end just like with Infura? If so which account and plan are we using for this>

And from their docs:

> Our API Service is currently in beta. We've got big plans for it, but please treat it as such and avoid running production applications on it.

https://support.onfinality.io/hc/en-us/articles/900004222183-API-Service-Getting-Started

alexcos20: it's one of the official RPCs:  https://docs.moonbeam.network/builders/get-started/moonriver/

alexcos20:
superseded by #1059

Issue #1057, opened 2021-10-04
Missing type declaration: Event.d.ts not included in build


jamiehewitt15:
`Event.d.ts` is not included in the ocean.js build.

```
node_modules/@oceanprotocol/lib/dist/node/ddo/DDO.d.ts:10:23 - error TS2307: Cannot find module './interfaces/Event' or its corresponding type declarations.

10 import { Event } from './interfaces/Event';
```

PR #1058, opened 2021-10-04
Changing files name event.d.ts -> event.ts

jamiehewitt15:
Closes: #1057

Changes proposed in this PR:
- Changing files name from `event.d.ts` to `event.ts`


kremalicious:
Not sure this is the right solution as files which only include typings really should be named `*.d.ts`. To me it looks like the exports are not correct, this was always a mess in ocean.js, but just when quickly looking through that `src/ddo/interfaces` folder shows that its `incex.ts` does not export the event typings. Have you tried adding the export to there?

jamiehewitt15:
Yeah, good point the Event typings should also be exported from `src/ddo/interfaces/index.ts`, I have added that now. 

The build fails if the Event typings are still named `Event.d.ts` when it is exported from that file:

```
Error: Could not resolve './Event' from src/ddo/interfaces/index.ts
```

Naming it `Event.ts` solves this build error. Also, I see that the other files  which only include typings in the same folder have all been named `*.ts` rather than `*.d.ts`

kremalicious:
Sure, then whatever works which seems to be combination of renaming and exporting from that index file. All good if that works, no need to further investigate as ocean.js will be rewritten completely

kremalicious (approved):

PR #1059, opened 2021-10-04
bump contracts + add networks

### LINKED ISSUE #1053: Add support for Moonriver & Energy Web

alexcos20:
Closes #1053


kremalicious (approved):
All good if we are not running into any rate limits for that public Moonriver RPC node

PR #1063, opened 2021-10-06
Delete parameters.md

kremalicious:
Just copying over files from ocean.py as is won't make them magically work in ocean.js. This parameter file is useless for ocean.js as we do not have any env vars here, and the whole doc just talks about a config.ini which we also don't have. It's a confusing document so remove it


trentmc:
cc @jamiehewitt15 for visibility here

PR #1073, opened 2021-10-14
Feature/issue 1060 change credential type

kianyee:
Fixes # . Support extend RBAC authentication

Changes proposed in this PR:

- change credential type from enum to string to support any credential type that control in the market
- new EventAccessControl class to handle RBAC related api call
- isConsumable to call RBAC api to validate event access


jamiehewitt15:
The only issue here seems to be that allow and deny list tests are failing in the marketplace flow, otherwise I'm happy to approve. Do you already know what the cause of the failed tests is?

kianyee:
isConsumable require barge to start up latest version of RBAC server

jamiehewitt15 (approved):
Great that you got the tests to pass. This will be merged into the V3 version of Ocean.js - the V4 version is being developed on a separate branch and won't have this functionality (as there is no plan for the branches to be merged). We will have to speak separately about how this can be achieved when you migrate to V4

PR #1081, opened 2021-10-21
V4

### LINKED ISSUE #1046: Ocean.js v4 alpha

bogdanfazakas:
Fixes #1046  .

The v4 main branch with the new version for the ocean.js lib where all other v4 PR's will be merged against. 
The will never be merged into main and it will become the new main after the v4 release and the current main will became v3 branch.

**In a highly WIP state, the tests and build may fail for the moment.**


mihaisc:
Do we need to prefix stuff with V4  `OceanPoolV4` ?

mihaisc:
```
interface FixedData {
  fixedPriceAddress: string
  addresses: string[]
  uints: (string | number)[]
}
```
Please stop using fields like `uints`, for a js developer it means nothing. This should be an object /struct with the following fields `[baseTokenDecimals,dataTokenDecimals, fixedRate, marketFee, withMint]` (got this from the contracts), so something like : 

```
interface FixedData {
  fixedPriceAddress: string
  addresses: string[]
  baseTokenDecimals: number
  dataTokenDecimals: number
  fixedRate: number
  marketFee: number
  withMint: boolean
}
```

PR #1178, opened 2021-12-29
pass down userCustomParameter to API request

soonhuat:
Fixes for parameterized consume and compute.

Changes proposed in this PR:

- pass userCustomParameter param to API request

PR #1179, opened 2021-12-31
allow custom parameters

alexcos20:
Changes proposed in this PR:

- make sure we pass userCustomParameters to provider download
- make sure we pass userCustomParameters to provider startCompute
- make sure we pass algoCustomParameters  & userCustomParameters for algorithm to provider startCompute
- make sure we pass userCustomParameters for each additionalInput to provider startCompute

(Supersedes #1178)

### DIFF
--- a/src/ocean/interfaces/Compute.ts
+++ b/src/ocean/interfaces/Compute.ts
@@ -1,7 +1,7 @@
 import DID from '../DID'
 import { Metadata } from '../../ddo/interfaces/Metadata'
 import { MetadataAlgorithm } from '../../ddo/interfaces/MetadataAlgorithm'
-
+import { UserCustomParameters } from '../../provider/Provider'
 export type ComputeResultType = 'algorithmLog' | 'output'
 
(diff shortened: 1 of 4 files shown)


src/ocean/interfaces/Compute.ts
```diff
@@ -52,4 +53,5 @@ export interface ComputeAlgorithm {
   transferTxId?: string
   dataToken?: string
   algoCustomParameters?: { [key: string]: any }
+  userCustomParameters?: UserCustomParameters
```
soonhuat, line 56: shouldn't userCustomParameters have the same type/structure of { [key: string]: any }, which is the key-value pair

alexcos20: it is.  it's imported from provider.ts at the beginning of the code

soonhuat: you are right, mistaken that to another var of similar naming

soonhuat (approved):

PR #1201, opened 2022-01-06
Get asset urls V4

### LINKED ISSUE #1136: Integrate new exposed asset URL(s) (V4) from Provider
Linked Provider Issue: https://github.com/oceanprotocol/provider/issues/282

bogdanfazakas:
Fixes #1136  .

Changes proposed in this PR:

- added suport for assetUrls method in provider

### DIFF
--- a/src/provider/Provider.ts
+++ b/src/provider/Provider.ts
@@ -214,7 +214,7 @@ export class Provider {
 
   /** Lists files from an asset
    * @param {String} documentId The ID of the asset/document (the DID)
-   * @param {number} serviceId ServiceId for the asset access service.
+   * @param {string} serviceId ServiceId for the asset access service.
    * @param {String} publisherAddress The publisher address
    * @param {Web3} web3 web3 instance
    * @param {String} providerUri uri of the provider that will be used
@@ -223,7 +223,7 @@ export class Provider {
    */
   public async getAssetUrls(
     documentId: string,
-    serviceId: number,
+    serviceId: string,
     publisherAddress: string,
     web3: Web3,
     providerUri: string,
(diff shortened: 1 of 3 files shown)


src/provider/Provider.ts
```diff
@@ -213,6 +213,59 @@ export class Provider {
...
+  /** Lists files from an asset
+   * @param {String} documentId The ID of the asset/document (the DID)
+   * @param {string} serviceId ServiceId for the asset access service.
+   * @param {String} publisherAddress The publisher address
+   * @param {Web3} web3 web3 instance
+   * @param {String} providerUri uri of the provider that will be used
+   * @param {any} getMethod get method client wants to use instance
+   * @return {Promise<FileMetadata[]>} returns the list of files for a documentId
+   */
+  public async getAssetUrls(
```
kremalicious, line 225: small naming thing, but in our terms an "asset"  is data set/algorithm metadata + files. So being more specific throughout this PR regarding what exactly the methods retrieve would be more clear. Also we seem to retrieve the full `FileMetadata` so not just the URLs? In that case this method here could be named `getFiles()`.

And just double checking cause not entirely sure about the intended usage of this method. Isn't this exposing the decrypted file URLs in the browser, like in the response? If so then I assume only the publisher can call this method?

bogdanfazakas: You are right its exposing the decrypted files and only publisher can call the method, and it will be used for the v3 to v4 migration

bogdanfazakas: Hmm got your point with the naming since in the future we may support other file types that may not have an url. I'll go with `getFiles()`

AnaLoznianu:
After some discussions, we decided this functionality is no longer needed. Having the decrypted URL available opens the door for security vulnerabilities

PR #1207, opened 2022-01-10
Fix/pool_approve

mihaisc:


bogdanfazakas (approved):

src/pools/balancer/Pool.ts
```diff
@@ -138,7 +138,7 @@ export class Pool {
 
     try {
       result = await token.methods
-        .approve(spender, new BigNumber(await this.amountToUnits(tokenAddress, amount)))
+        .approve(spender, Web3.utils.toWei(amountFormatted))
```
bogdanfazakas, line 141: i think amountToUnits needs some adjustments also if sending before convering to wei

mihaisc: ah, yes, my bad, this is not done, there are other issues as well :)

PR #1208, opened 2022-01-10
Add ocean-contracts artifacts in root dir

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- add artifacts folder with the used artifacts from ocean-contracts in the root folder
- update imports artifacts import  
- removed not used dependancies


bogdanfazakas:
Moved the artifacts in the src folder, if i have them in root folder the build fails because we have tsconfig rootDir pointing to src folder. Trying fix that locally and update the pr later if i succed but wanted to have this green till then

mihaisc:
i don't think we need the .dbg files , am i wrong @alexcos20

bogdanfazakas:
> i don't think we need the .dbg files , am i wrong @alexcos20

we don't use them so i guess we do not need them

bogdanfazakas:
> Moved the artifacts in the src folder, if i have them in root folder the build fails because we have tsconfig rootDir pointing to src folder. Trying fix that locally and update the pr later if i succed but wanted to have this green till then

should be fine now updated the config file

bogdanfazakas:
> Please move https://github.com/oceanprotocol/contracts/blob/v4main_postaudit/addresses/address.json here as well, it can be somewhere in src

it was added in the artifacts folder, just updated it with latest changes made by @alexcos20 , should i move it to src folder or we can have it here in the artifacts folder?

mihaisc:
leave it

alexcos20:
> i don't think we need the .dbg files , am i wrong @alexcos20

no, .dgb files are only for hardhat debugs

alexcos20 (approved):
nice to have dbg files removed before merge

mihaisc (approved):

Issue #1210, opened 2022-01-11
remove native Node.js module usage, create proper browser build


kremalicious:
Nobody knows by now why ocean.js is a Node.js build by default, yet it's majority of actual usage happens in the browser. And looks like only the `ConfigHelper` is responsible for using 2 modules (`fs` & `os`) just so it can load a potential `address.json` from somewhere on user's disk.

Refactor to remove Node.js module usage, or replace with package which works in browser. Would highly question that `ADDRESS_FILE` use case though so best to remove it.

E.g. the `homedir/os` usage can be removed right away as this logic does not make sense cause that homedir fallback would never fire:
https://github.com/oceanprotocol/ocean.js/blob/4bbdd521119634625accb1a80f13aeb3728fa896/src/utils/ConfigHelper.ts#L183-L191

Likewise, final build should be then a browser build by default with changes like these:

package.json

<img width="693" alt="Screen Shot 2022-01-11 at 15 19 04" src="https://user-images.githubusercontent.com/90316/148971889-0d37ec81-b685-487a-9d08-c5a8313b576b.png">

tsconfig.json
<img width="405" alt="Screen Shot 2022-01-11 at 15 20 10" src="https://user-images.githubusercontent.com/90316/148971902-c13eb008-1f13-4532-ad7d-e8b4168e1d9b.png">

Addresses #1155

PR #1211, opened 2022-01-11
Build tweaks

### LINKED ISSUE #1209: typing index file is missing in npm package
Typings are not picked up in external projects cause the referenced `./dist/node/index.d.ts` does not exist in final package:

Screen Shot 2022-01-11 at 14 04 24

kremalicious:
After tinkering a bit with different options, I was able to just run with the default provided values .

- output browser target instead of Node.js target, ref #1210
- move artifacts folder under /src, so we can ditch specifying the root directory in `tsconfig`, which always comes with more consequences
- remove `os` usage from `ConfigHelper`, ref #1210
- point `typings` to existing typings index file, closes #1209
- build tool tweaks:
  - replicate micro bundle config from https://github.com/developit/microbundle#specifying-filenames-in-packagejson
  - run with suggestions for tsconfig `module` & `target`
  - not sure if `esm` instead of `es` as output is best choice or if there is a difference within `microbundle`
- test tool tweaks:
  - compile test files individually with custom `tsconfig`
  - single `tsconfig` & `mocharc` for both test runs
  - new `npm run mocha` command so we can set some defaults, making it pick up individual `tsconfig`

With all that, these are the only remaining warnings I get during the build (so no warnings about Node.js module usage so remaining `fs` usage might be not problematic at all) and the final file sizes kinda become tolerable:

Screen Shot 2022-01-11 at 16 43 14


kremalicious:
oh boy, guess now we are own this rabbit hole of figuring out how to transpire only the test files 🥃

test/integration/tsconfig.json
```diff
@@ -1,9 +1,11 @@
...
-      "allowSyntheticDefaultImports": true
-    }
-  }
\ No newline at end of file
+  "compilerOptions": {
+    "resolveJsonModule": true,
+    "lib": ["es6", "es7"],
+    "module": "CommonJS",
+    "target": "ES5",
+    "noUnusedLocals": true,
```
bogdanfazakas, line 7: you were a bit strict here 😄  it now return errors on issues that are most likely warning for eg:
Error: test/integration/ComputeFlow.test.ts(1,1): error TS6133: 'config' is declared but its value is never read.
but we can handle those and delete unused vars

kremalicious: yeah, figured there's no need for all those individual configs and merged them all into one place in fdcf7df. And made `"noUnusedLocals": false` so I hope now everything gets green again.

bogdanfazakas (approved):
LGTM nice tweaks 🚀  and if we are fine having artifacts in src folder we can merge this 👍

kremalicious:
OMG, it's green

giphy

Now, final question would be what package market will pick up but since `v4` over there uses webpack 5 in next.js, I am 95% confident it will pick up the CommonJS module referenced under `lib` in package.json

PR #1214, opened 2022-01-12
Allow users/market to set own gas contract options & configurable web3 contract params

### LINKED ISSUE #1106: Need a flexible way to change some default options of web3.eth.Contract
The default value for [transactionBlockTimeout](https://web3js.readthedocs.io/en/v1.2.4/web3-eth-contract.html#transactionblocktimeout) is 50. Hence, we need a flexible way to change the `options` when we are on network with faster block time to prevent catching the error below too quickly when there is a spike in gwei price.

image

### LINKED ISSUE #1107: Option for user/market specify their own gas price
Make multiplier value a parameter at: 
https://github.com/oceanprotocol/ocean.js/blob/e07a7cb6ecea12b39ed96f994b4abe37806799a1/src/utils/GasUtils.ts#L4-L6

bogdanfazakas:
Fixes #1106  
Fixes #1107 

Changes proposed in this PR:

- added multiplier gas value as parameter in config and use it in the get fair gas price
- added transactionPollingTimeout, transactionConfirmationBlocks & transactionBlockTimeout as parameter in config
- updated codebase to use and set the default options in a similar way we had them in v3


src/models/Config.ts
```diff
@@ -170,6 +170,30 @@ export class Config {
...
+  transactionConfirmationBlocks: number
+
+  /**
+   * Specify the transaction Polling Blocks Timeout
+   * @type {number}
+   */
+  transactionPollingTimeout: number
+
+  /**
+   * Specify the multiplaier for the gas fee
```
alexcos20, line 193:
```suggestion
   * Specify the multiplier for the gas fee
```

src/utils/ContractUtils.ts
```diff
@@ -1,6 +1,28 @@
...
 import Web3 from 'web3'
+import BigNumber from 'bignumber.js'
+import { Contract } from 'web3-eth-contract'
 import { generateDtName } from './DatatokenName'
+import { Erc20CreateParams, FreCreationParams, PoolCreationParams } from '../interfaces'
+import { Config } from '../models'
+
+export function setContractDefaults(contract: Contract, config: Config): Contract {
+  if (config) {
+    contract.transactionBlockTimeout = config.transactionBlockTimeout
```
alexcos20, line 10:
```suggestion
    if(config.transactionBlockTimeout) contract.transactionBlockTimeout = config.transactionBlockTimeout
```

src/utils/ContractUtils.ts
```diff
@@ -1,6 +1,28 @@
...
+    contract.transactionConfirmationBlocks = config.transactionConfirmationBlocks
```
alexcos20, line 11: if (config.transactionConfirmationBlocks) contract.transactionConfirmationBlocks = config.transactionConfirmationBlocks

src/utils/ContractUtils.ts
```diff
@@ -1,6 +1,28 @@
...
+    contract.transactionPollingTimeout = config.transactionPollingTimeout
```
alexcos20, line 12: if (config.transactionPollingTimeout) contract.transactionPollingTimeout = config.transactionPollingTimeout

alexcos20 (approved):

src/models/Config.ts
bogdanfazakas: oops missed that 😄

PR #1234, opened 2022-01-20
Add setMetaDataAndTokenURI method

### LINKED ISSUE #1228: Set the metaData and the tokenURI in a single call

bogdanfazakas:
Fixes #1228  .

Changes proposed in this PR:

- add support for setMetaDataAndTokenURI function
- added unit test

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "test:pool": "npm run mocha -- 'test/unit/pools/balancer/Pool.test.ts'",
     "test:dispenser": "npm run mocha -- 'test/unit/pools/dispenser/Dispenser.test.ts'",
     "test:dt": "npm run mocha -- 'test/unit/tokens/Datatoken.test.ts'",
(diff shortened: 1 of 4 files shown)


kremalicious:
should be good, just the naming. We mirror that typo from the contracts, as _metadata_ is one word so the method should be `setMetadataAndTokenURI`. As you already corrected that with e.g. `setMetadata()` method we should stay consistent and also correct this in all methods in this PR

bogdanfazakas:
> should be good, just the naming. We mirror that typo from the contracts, as _metadata_ is one word so the method should be `setMetadataAndTokenURI`. As you already corrected that with e.g. `setMetadata()` method we should stay consistent and also correct this in all methods in this PR

sure going to handle it for both methods and interfaces that we have in oceanjs and update the naming and once that's fixed in contracts I'm going to tackle it another pr and update the names of the functions we call

PR #1235, opened 2022-01-21
remove optional fetch method, add abort signal

mihaisc:
Removed optional fetch method
Added optional abort signal

### DIFF
--- a/src/aquarius/Aquarius.ts
+++ b/src/aquarius/Aquarius.ts
@@ -14,9 +14,9 @@ export class Aquarius {
   /** Resolves a DID
    * @param {string} did
    * @param {AbortSignal} signal abort signal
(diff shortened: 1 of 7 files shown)


src/aquarius/Aquarius.ts
```diff
@@ -14,18 +13,22 @@ export class Aquarius {
 
   /** Resolves a DID
    * @param {string} did
-   * @param {string} fetchMethod fetch client instance
+   * @param {AbortSignal} signal abort signal
    * @return {Promise<DDO>} DDO
```
kremalicious, line 17: isn't this returning `Asset` as it's coming from Aquarius?

mihaisc: true, but not the issue

PR #1287, opened 2022-02-14
fix fixed order params

mihaisc:


kremalicious (approved):
I just assume it fixes my fixed price assets ordering so great! 😀

mihaisc:
correct, i am amazed how it worked in my tests earlier this morning, and the ocean.js tests have no issue with this.

PR #1291, opened 2022-02-15
use barge addresses for unit tests

alexcos20:
Use barge addresses for unit tests.

Removed a lot of code, that was way out of scope.
Why do we need router.updateOPCFee in ocean.js unit tests?
Why are we trying to add another NFT template from ocean.js?
........ (list can go on).....

Also, we ARE NOT using any ocean class in the unit tests.  Our unit tests are an copy of ocean-contracts unit tests files..  Why ?


mihaisc (approved):
good as a starting point

PR #1318, opened 2022-03-01
Replace toWei/fromWei when base token conversion

### LINKED ISSUE #1279: Stop using toWei/fromWei from web3
We need to use amountToWei, weiToAmount to be compatible to tokens with various decimals

KatunaNorbert:
Fixes #1279 .

### DIFF
--- a/src/pools/fixedRate/FixedRateExchange.ts
+++ b/src/pools/fixedRate/FixedRateExchange.ts
@@ -366,10 +366,13 @@ export class FixedRateExchange {
     const estGas = await this.estSetRate(
       address,
       exchangeId,
-      await this.amountToUnits(exchange.baseToken, String(newRate))
+      await this.amountToUnits(exchange.baseToken, newRate.toString())
     )
     const trxReceipt = await this.contract.methods
-      .setRate(exchangeId, await this.amountToUnits(exchange.baseToken, String(newRate)))
+      .setRate(
+        exchangeId,
+        await this.amountToUnits(exchange.baseToken, newRate.toString())
+      )
       .send({
         from: address,
         gas: estGas + 1,
@@ -534,7 +537,14 @@ export class FixedRateExchange {
    */
   public async getRate(exchangeId: string): Promise<string> {
     const weiRate = await this.contract.methods.getRate(exchangeId).call()
-    return this.web3.utils.fromWei(weiRate)
+    const rate = await unitsToAmount(
+      this.web3,
+      (
+        await this.getExchange(exchangeId)
+      ).baseToken,
+      weiRate
+    )
+    return rate
   }
 
   /**
--- a/src/pools/fixedRate/FixedRateExchange.ts
+++ b/src/pools/fixedRate/FixedRateExchange.ts
@@ -635,7 +635,7 @@ export class FixedRateExchange {
       .calcBaseOutGivenInDT(
         exchangeId,
         await this.amountToUnits(exchange.datatoken, datatokenAmount),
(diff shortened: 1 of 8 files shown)


src/pools/fixedRate/FixedRateExchange.ts
```diff
@@ -366,10 +366,10 @@ export class FixedRateExchange {
     const estGas = await this.estSetRate(
       address,
       exchangeId,
-      await this.unitsToAmount(exchange.baseToken, String(newRate))
+      await this.amountToUnits(exchange.baseToken, String(newRate))
```
mihaisc, line 369: why noy toString()?

KatunaNorbert: this is what was used above so I used to be consistent

mihaisc: strange, we always use toString, don't know how that stayed there.

KatunaNorbert: I will change it

bogdanfazakas: feel free to update all String(value) in other places if you find more of this 😃

KatunaNorbert: ok, I will

src/pools/fixedRate/FixedRateExchange.ts
```diff
@@ -362,13 +362,17 @@ export class FixedRateExchange {
     exchangeId: string,
     newRate: string
   ): Promise<TransactionReceipt> {
+    const exchange = await this.getExchange(exchangeId)
     const estGas = await this.estSetRate(
       address,
       exchangeId,
-      this.web3.utils.toWei(String(newRate))
+      await this.amountToUnits(exchange.baseToken, newRate.toString())
```
bogdanfazakas, line 369: this conversions should also take place in `estSetRate()` just pass the raw string values to the estimate method

src/pools/fixedRate/FixedRateExchange.ts
```diff
@@ -362,13 +362,17 @@ export class FixedRateExchange {
...
     )
     const trxReceipt = await this.contract.methods
-      .setRate(exchangeId, this.web3.utils.toWei(newRate))
+      .setRate(
+        exchangeId,
+        await this.amountToUnits(exchange.baseToken, newRate.toString())
```
bogdanfazakas, line 374: also should we have setNewRate in wei even if we have have the rate in raw string value in FreCreationParams used to create a new fixed rate for example ?

bogdanfazakas: nevermind we convert it there also here https://github.com/oceanprotocol/ocean.js/blob/v4main/src/utils/ContractUtils.ts#L83

src/pools/fixedRate/FixedRateExchange.ts
KatunaNorbert: right, makes sense

KatunaNorbert:
I run into an issue when using `amountToUnits` inside `createNftErc20WithFixedRate` -> `getFreCreationParams` with USDC base token to transform the fixed rate value.

When sending the rate to the contract it contains 6 zeros so `1000000` but the transaction is reverted because in the contracts we have a `MIN_RATE` check which is `10 ** 10`.
I think `amountToUnits` will not work in this case unless we change the contract.


Screenshot 2022-03-07 at 14 44 01

mihaisc (approved):

Issue #1274, opened 2022-02-10
Check allowance in startOrder


mihaisc:

AnaLoznianu:
@mihaisc Is this issue still relevant?

PR #1320, opened 2022-03-03
check currentAllowance on startOrder

EnzoVezzaro:
Fixes [Check allowance in startOrder #1274](https://github.com/oceanprotocol/ocean.js/issues/1274).

Changes proposed in this PR:

- added check allowance on startOrder


mihaisc:
Not ok, you never need to check allowance for gas fee.
In this case you need to check allowance for  publishMarketFee, consumeMarketFee, providerFees. `publishMarketFee` you need to do a contract call to get it, the other two are in the parameters. Once you have these you need to group them based on token address and do allowance for each token.
For example if all are in ocean you just add then and check for allowance once. If all three are different, you need 3 checks.

EnzoVezzaro:
i'm not sure i'm fully grasping what it's need it here. Based on the information i can retrieve both `providerFees` and `consumeMarketFee` have their fee token as `0x0000000000000000000000000000000000000000` making the allowance function not functional. Another problem i found is that both object for `providerFees` and `consumeMarketFee` have an amount of '0' making the total for the allowance always '0'. I'm sure i'm still missing something, am i getting the right information in the params? is right to be always '0'?

mihaisc:
The `0x0..` address is set when we don't have any fee, so if you detect that then you don't need to do any checks for that fee. If the fee is 0 then you don't need to check for allowance. They have the value 0 because we set it to be 0 because that was the golden path. You can create tests where it's not 0.

EnzoVezzaro:
I'm unsure what is going on [here](https://github.com/oceanprotocol/ocean.js/blob/feature/issue-1274-check-allowance-startOrder/src/tokens/Datatoken.ts#L1043:L1045) though. We're saying that the `currentAllowance` (which has been always '0' so far) must be less than the token's amount, is that right?

src/tokens/Datatoken.ts
```diff
@@ -973,13 +975,73 @@ export class Datatoken {
...
+        return token
+      }
+    })
+
+    try {
+      const allownceTokens = await Promise.all(getCurrentAllownceTokens)
+    } catch (e) {
+      LoggerInstance.error(`ERROR: Failed checking allowance : ${e}`)
+      throw new Error(`Failed checking allowance: ${e}`)
+    }
```
bogdanfazakas, line 1043: Think moving this to a another method maybe under utils would be better and just calling the method here, since it's adding a lot of code to the startOrder method alters method's scope

EnzoVezzaro: got it 👍🏽 you're right. I'll move it into utils as it makes more sense there. Thanks

AnaLoznianu:
Closing this as it was solved with other issues while working on v4

PR #1328, opened 2022-03-09
 Dispenser creation params should not be in wei

### LINKED ISSUE #1302: Dispenser creation params should not be in wei
https://github.com/oceanprotocol/ocean.js/blob/36fa91f1dc9a86d8d93a2dc1f8a9a14f43927ab6/src/factories/NFTFactory.ts#L862

conversion in wei should happen in ocean.js like all the other functions.

KatunaNorbert:
Fixes #1302


kremalicious (approved):
should work based on the tests, ultimately can only be tested in complete flow in `market`

Issue #1246, opened 2022-02-01
Integration tests:  Use confighelper instead of hardcoded endpoints


alexcos20:
See https://github.com/oceanprotocol/ocean.js/blob/v4main/test/integration/ComputeFlow.test.ts#L26-L28

We should use ConfigHelper instead of hardcoding endpoints

Update ConfigHelper with barge endpoints as well

PR #1331, opened 2022-03-11
Issue-#1292: Fix Compute test that is failing due to 2 seconds timeout

### LINKED ISSUE #1292: Fix Compute test that is failing due to 2 seconds timeout

miquelcabot:
Fixes #1292.

Changes proposed in this PR:

- Set correct services URLs in the tests


alexcos20 (changes requested):
Please use config helper (https://github.com/oceanprotocol/ocean.js/blob/v4main/src/utils/ConfigHelper.ts) to get all required endpoints.
See #1246

PR #1336, opened 2022-03-14
Make metadataProofs optional in setMetadataAndTokenURI

### LINKED ISSUE #1307: Change metadataProofs to optional in MetadataAndTokenURI interface
In reference to https://github.com/oceanprotocol/market/issues/996

In other metadata methods (e.g. [setMetadata](https://github.com/oceanprotocol/ocean.js/blob/aadd1458ba7364a709f6eb94b3f8c123de3c189a/src/tokens/NFT.ts#L1059)) `metadataProofs` is an optional parameter.

When using `setMetadataAndTokenURI` the `metadataAndTokenURI` is using the `MetadataAndTokenURI`interface, where `metadataProofs` [is required](https://github.com/oceanprotocol/ocean.js/blob/aadd1458ba7364a709f6eb94b3f8c123de3c189a/src/%40types/Erc721.ts#L12). To be consistent this should probably also be optional in the interface.

I also noticed some inconsistencies in the naming of parameters. Previously all variable names including `metadata` treat metadata as one word [(e.g. metadataState)](https://github.com/oceanprotocol/ocean.js/blob/aadd1458ba7364a709f6eb94b3f8c123de3c189a/src/tokens/NFT.ts#L1053). In the interface, `metaData` is now written in camelcase, for example: [metaDataState](https://github.com/oceanprotocol/ocean.js/blob/aadd1458ba7364a709f6eb94b3f8c123de3c189a/src/%40types/Erc721.ts#L4)

md0x:
Signed-off-by: Pablo Maldonado <pablo@oceanprotocol.com>

Fixes #1307 .

Changes proposed in this PR:

- Make metadataProofs optional in setMetadataAndTokenURI

### DIFF
--- a/src/@types/Erc721.ts
+++ b/src/@types/Erc721.ts
@@ -9,5 +9,5 @@ export interface MetadataAndTokenURI {
   metaDataHash: string
   tokenId: number
   tokenURI: string
-  metadataProofs: MetadataProof[]
+  metadataProofs?: MetadataProof[]
 }
--- a/src/tokens/NFT.ts
+++ b/src/tokens/NFT.ts
@@ -1119,9 +1119,13 @@ export class Nft {
       )
     const gasLimitDefault = this.GASLIMIT_DEFAULT
     let estGas
+    const sanitizedMetadataAndTokenURI = {
+      ...metadataAndTokenURI,
+      metadataProofs: metadataAndTokenURI.metadataProofs || []
+    }
     try {
       estGas = await nftContract.methods
-        .setMetaDataAndTokenURI(metadataAndTokenURI)
+        .setMetaDataAndTokenURI(sanitizedMetadataAndTokenURI)
         .estimateGas({ from: metadataUpdater }, (err, estGas) =>
           err ? gasLimitDefault : estGas
         )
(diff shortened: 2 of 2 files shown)


src/@types/Erc721.ts
```diff
@@ -9,5 +9,5 @@ export interface MetadataAndTokenURI {
   metaDataHash: string
   tokenId: number
   tokenURI: string
-  metadataProofs: MetadataProof[]
+  metadataProofs?: MetadataProof[]
```
bogdanfazakas, line 12: unfortunately adding only this change is not enough, we need some checks in the hole codebase in every place the interface is used and check for `metadataProofs` existence and if not set we need to set an empty string array `= []` to `metadataAndTokenURI.metadataProofs` , otherwise the contract call will fail if metadataProofs is undefined or null. 
So we need a check similar with this https://github.com/oceanprotocol/ocean.js/blob/v4main/src/tokens/NFT.ts#L1065  in all places the `MetadataAndTokenURI` is used

bogdanfazakas (approved):
LGTM 👍

PR #1339, opened 2022-03-15
Issue-#1246: Integration tests: Use confighelper instead of hardcoded endpoints

### LINKED ISSUE #1246: Integration tests:  Use confighelper instead of hardcoded endpoints

miquelcabot:
Fixes #1246.

Changes proposed in this PR:
- Use ConfigHelper instead of hardcoded endpoints in tests

### DIFF
--- a/src/utils/ConfigHelper.ts
+++ b/src/utils/ConfigHelper.ts
@@ -7,7 +7,7 @@ const configHelperNetworksBase: Config = {
   chainId: null,
   network: 'unknown',
   metadataCacheUri: 'https://v4.aquarius.oceanprotocol.com',
-  nodeUri: 'http://localhost:8545',
+  nodeUri: 'http://127.0.0.1:8545',
   providerUri: 'http://127.0.0.1:8030',
   subgraphUri: null,
   explorerUri: null,


test/integration/ComputeFlow.test.ts
```diff
@@ -21,10 +20,12 @@ const data = JSON.parse(
   )
 )
 
+const config = configHelperNetworks[1]
```
alexcos20, line 23: Please use getConfig (https://github.com/oceanprotocol/ocean.js/blob/v4main/src/utils/ConfigHelper.ts#L205)  with web3.chainId to get the correct network.
It's not safe to assume that configHelperNetworks[1] will always be development

miquelcabot: Solved. Now I use  a config file to get these values:
[https://github.com/oceanprotocol/ocean.js/blob/165b4c86809615aada2463f2ece2a3be1d3ee419/test/config.ts](https://github.com/oceanprotocol/ocean.js/blob/165b4c86809615aada2463f2ece2a3be1d3ee419/test/config.ts)

test/TestContractHandler.ts
```diff
@@ -98,17 +96,11 @@ export class TestContractHandler {
   }
 
   public async deployContracts(owner: string, routerABI?: AbiItem | AbiItem[]) {
-    const data = JSON.parse(
-      fs.readFileSync(
-        process.env.ADDRESS_FILE ||
-          `${homedir}/.ocean/ocean-contracts/artifacts/address.json`,
```
bogdanfazakas, line 104: I'm not sure if we should get this out and replaced those addresses since it reads the contract addresses deployed by barge, and I think we should be still using those in our tests. So i think we should focus on replacing endpoints of services like provider Aquarius and web3 for eg in this PR but not the contract addresses

bogdanfazakas: nvm just saw the config logic that reads that

PR #1364, opened 2022-03-22
Solve all typing warnings

kremalicious:
Cause it's a pain having to constantly see them on every PR review.

Please keep warnings clean in the future, never a need to keep unused imports or define unused variables


bogdanfazakas (approved):
Good points 👍

PR #1365, opened 2022-03-23
Small improvements on encrypt and getHash methods

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- removes decodeUri from body the encrypt 
- adds try catch on get hash
- update getNonce JsDoc


bogdanfazakas:
merged in #1386

PR #1374, opened 2022-03-28
default amountToUnits decimals to 18, FRE consumerMarketSwapFee witho…

soonhuat:
…ut basetoken

Fixes # .

Changes proposed in this PR:

- FRE calcBaseInGivenOutDT convert consumeMarketSwapFee (percentage/fraction based value) non-token specific, via web3.toWei to unit before pass to smart contract method (which will divide by Base of 18)

### DIFF
--- a/src/utils/ContractUtils.ts
+++ b/src/utils/ContractUtils.ts
@@ -146,8 +146,8 @@ export async function amountToUnits(
   amount: string
 ): Promise<string> {
   try {
-    const tokenContract = token && new web3.eth.Contract(minAbi, token)
-    let decimals = token ? await tokenContract.methods.decimals().call() : '0'
+    const tokenContract = new web3.eth.Contract(minAbi, token)
+    let decimals = await tokenContract.methods.decimals().call()
     if (decimals === '0') {
       decimals = 18
     }
--- a/src/utils/ContractUtils.ts
+++ b/src/utils/ContractUtils.ts
@@ -146,8 +146,8 @@ export async function amountToUnits(
   amount: string
 ): Promise<string> {
   try {
-    const tokenContract = new web3.eth.Contract(minAbi, token)
-    let decimals = await tokenContract.methods.decimals().call()
+    const tokenContract = token && new web3.eth.Contract(minAbi, token)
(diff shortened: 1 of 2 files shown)


mihaisc:
Don't change `amountToUnits` function, just use Web3.utils.toWei

src/utils/ContractUtils.ts
```diff
@@ -146,8 +146,8 @@ export async function amountToUnits(
   amount: string
 ): Promise<string> {
   try {
-    const tokenContract = new web3.eth.Contract(minAbi, token)
-    let decimals = await tokenContract.methods.decimals().call()
+    const tokenContract = token && new web3.eth.Contract(minAbi, token)
+    let decimals = token ? await tokenContract.methods.decimals().call() : '0'
```
kremalicious, line 150: what weird logic combined with next 3 lines which should be all removed. If `decimals` is set here to `'0'` as part of `else` then no need to check for it again in next lines to set it to something else again. So for `else`, `decimals` here should be just `18` and then no need for the next 3 lines. And then it's a `const` and no `let`

kremalicious:
> amountToUnits with default token decimals to 0

the code does not reflect this as `decimals` is immediately set to `18` as soon as `decimals` is `'0'`. Commented on respective code line above.

And then what Mihai said, I don't have any context on this PR as no issue to be found describing what it tried to solve so no `amountToUnits` usage as it looks like

mihaisc:
Yes it's a simple solution somthing like 
```
.calcBaseInGivenOutDT(
        exchangeId,
        await this.amountToUnits(fixedRateExchange.datatoken, datatokenAmount),
        Web3.utils.toWei(consumeMarketFee)
      )
```

soonhuat:
@mihaisc ya, that's the right way to go, earlier got blinded by the existing conversion method

mihaisc (approved):

kremalicious (changes requested):
`test_integration` needs to succeed, `test_unit` is fine failing if only 1 specific test fails at the moment (Pool unit test - Test a pool with DAI (18 Decimals) - #swapExactAmountIn - should swap)

mihaisc:
Test integration fails because of compute timeoute issue

PR #1378, opened 2022-03-29
Fix typedoc

mihaisc:


kremalicious (approved):
should work, let's hope over on docs where the `list/lib.json` file is used, doesn't require any typed related updates to render it

PR #1386, opened 2022-04-01
Fix get compute results

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- fixed and renamed  ( `getComputeResultUrl` ) compute results method
- removes `decodeUri` from `encrypt` method, adds try catch on `getHash` and updates `getNonce` JsDoc
 
~~Note: **Tests will fails** since the PR uses ocean contracts 1.0.0-alpha.26 version. Used this version to be able to work locally with current assets on market. Will remove the downgrade once all other components will use the current contracts version~~


kremalicious:
does this need changes after we merged #1395?

src/provider/Provider.ts
```diff
@@ -573,23 +571,22 @@ export class Provider {
...
-   * @param {string} destination
-   * @param {Web3} web3
-   * @param {AbortSignal} signal abort signal
+  /** Get compute result url
+   * @param {string} providerUri The URI of the provider we want to query
+   * @param {Web3} web3 Web3 instance
+   * @param {string} consumerAddress The consumer ethereum address
+   * @param {string} jobId The ID of a compute job.
+   * @param {number} index Result index
+   * @param {AbortSignal} signal Abort signal
```
kremalicious, line 580: no `AbortSignal` param on the actual method, and the return type is different from this jsdoc comment too. So most likely comment has to be updated

src/provider/Provider.ts
```diff
@@ -599,38 +596,26 @@ export class Provider {
...
-      } else {
-        return await downloadFile(consumeUrl, index)
-      }
-    } catch (e) {
-      LoggerInstance.error('Error getting job result')
-      LoggerInstance.error(e)
-      throw e
-    }
+    if (!resultUrl) return null
+    return resultUrl
```
kremalicious, line 618: This would do the same thing as those 2 lines and multiple return statements:

```ts
    return resultUrl || null
```

But looks like `resultUrl` will always be defined as we are adding strings to it above so would end up with some string like `null?consumerAddress=&jobId=`. So not sure what the check for empty `resultUrl` would give us here

test/integration/ComputeFlow.test.ts
```diff
@@ -10,9 +9,12 @@ import {
   NftCreateData,
   Datatoken,
   getHash,
-  Nft
+  Nft,
+  downloadFile,
```
kremalicious, line 13: remove unused imports `downloadFile` & `fetchData`

src/provider/Provider.ts
```diff
@@ -11,7 +11,6 @@ import {
 } from '../@types/'
 import { noZeroX } from '../utils/ConversionTypeHelper'
 import fetch from 'cross-fetch'
-import { DownloadResponse } from '../@types/DownloadResponse'
```
kremalicious, line 14: more unused imports above imported `from '../utils'`

src/provider/Provider.ts
bogdanfazakas: missed that 👍  updated the method too many times 😃

src/provider/Provider.ts
kremalicious: GitHub still sees them, just couldn't comment on those lines as they were not changed:

Screen Shot 2022-04-14 at 17 16 38

src/provider/Provider.ts
```diff
@@ -573,23 +571,21 @@ export class Provider {
...
-   * @param {string} destination
-   * @param {Web3} web3
-   * @param {AbortSignal} signal abort signal
+  /** Get compute result url
+   * @param {string} providerUri The URI of the provider we want to query
+   * @param {Web3} web3 Web3 instance
+   * @param {string} consumerAddress The consumer ethereum address
+   * @param {string} jobId The ID of a compute job.
+   * @param {number} index Result index
    * @return {Promise<ComputeJob | ComputeJob[]>}
```
kremalicious, line 580: return type here is different than TypeScript-typed return type `Promise<string>`

kremalicious (approved):

PR #1387, opened 2022-04-01
Add transfer() function to TokenUtils

### LINKED ISSUE #1390: Add transfer() function to TokenUtils
Add `transfer()` function to `TokenUtils` to simplify the call to ERC20 functions in the tests

miquelcabot:
Fixes #1390 

Changes proposed in this PR:
- Add `transfer()` function to `TokenUtils`


src/utils/TokenUtils.ts
```diff
@@ -45,7 +45,7 @@ export async function estApprove(
  * @param {String} account
  * @param {String} tokenAddress
  * @param {String} spender
- * @param {String} amount  (always expressed as wei)
+ * @param {String} amount amount of ERC20 tokens, expressed without decimals (not as wei)
```
mihaisc, line 48: why without decimals?

miquelcabot: @mihaisc I mean "as an integer ERC20 amount". For example: in 1 DAI, we must set `amount` as, `1`, not as `1000000000000000000`

mihaisc: but those are not decimals , without decimals means you can send only integers and that is false.

miquelcabot: @mihaisc Then it can be better `amount of ERC20 tokens (not as wei)`?

miquelcabot: @mihaisc Solved

jamiehewitt15:
Did you look at trying to solve the codeclimate issues?

PR #1389, opened 2022-04-04
default nonce dateNow, include nonce in computeResult params

soonhuat:
Fixes # .

Changes proposed in this PR:

- default getNonce to date.now if not in redis
- include nonce in computeResult params as commented (`The UTC timestamp, used to prevent replay attacks`) in provider endpoint but not in used yet

### DIFF
--- a/src/provider/Provider.ts
+++ b/src/provider/Provider.ts
@@ -100,7 +100,8 @@ export class Provider {
         },
         signal: signal
       })
-      return (await response.json()).nonce.toString()
+      const nonce = (await response.json()).nonce || Date.now()
+      return nonce.toString()
     } catch (e) {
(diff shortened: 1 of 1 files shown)


alexcos20 (dismissed):

bogdanfazakas:
there is already a PR opened #1386 that fixes this for computeResults. We decided to keep that as a draft and merge once we bump everything to use the latest version of contracts

soonhuat:
> there is already a PR opened #1386 that fixes this for computeResults. We decided to keep that as a draft and merge once we bump everything to use the latest version of contracts, till then we use version 1.0.0.alpha-26 in that pr so we can use the build locally in the market

alright, then i just abandon the PR

PR #1394, opened 2022-04-05
Issue-#1391: Add generic estimateGas() function

### LINKED ISSUE #1391: Create a generic `estimateGas()` function
Create a generic `estimateGas()` function that can be used whenever we want to estimate gas in a function.

miquelcabot:
Fixes #1391 

Changes proposed in this PR:
- Add generic `estimateGas()` function
- Call this function from all the estimate gas specific functions
- Call this function from all the functions that need to estimate the gas


src/utils/ContractUtils.ts
```diff
@@ -162,3 +162,29 @@ export async function amountToUnits(
...
 }
+
+/**
+ * Estimates the gas used when a function would be executed on chain
+ * @param {string} from account that calls the function
+ * @param {Function} functionToEstimateGas function that we need to estimate the gas
+ * @param {...any[]} args arguments of the function
+ * @return {Promise<number>} gas cost of the function
+ */
+export async function estimateGas(
```
bogdanfazakas, line 173: Can we somehow write a test for this method ? I'm thinking of having picking one contract method for eg and compare the value returned by this method to the value returned by calling estimate directly on the contract method as we did before ? Or maybe something else if you have a better idea .
By doing this i think we may also not have issues with code climate coverage dropping and the diff coverage

bogdanfazakas:
Looks good in general should work, do you think is worth testing it locally in the market on the estimates we have in publish flow ?

src/utils/ContractUtils.ts
miquelcabot: @bogdanfazakas I can add a `Utils.test` file, for example, to test functions like this one. But I also think that this function is executed in a lot of tests, because all the specific functions used to estimate gas for concrete ABI functions, like `estGasCreateNFT()`, `estGasAddNFTTemplate()`... also calls this function. And this function is also called directly from all the functions that call the smart contract functions, like `createNFT()`, `addNFTTemplate()`...

jamiehewitt15:
> Looks good in general should work, do you think is worth testing it locally in the market on the estimates we have in publish flow ?

Yeah, I agree it would be good to test this with the market before merging. @miquelcabot have you done that?

Also, there seems to be a lot of codeclimate issues - 89. Have you looked into fixing them?

jamiehewitt15 (approved):
Ok, looks good to me

PR #1396, opened 2022-04-07
setPaymentCollector apply same checking as contract

### LINKED ISSUE #1341: As an NFT Owner, I am not able to call `setPaymentCollector` due to the condition checking here
**Issue:**
As an NFT Owner, I am not able to call `setPaymentCollector()` due to the condition checking here

https://github.com/oceanprotocol/ocean.js/blob/d8a7d42a9e75cb36edf47381a29d4418ee7bc249/src/tokens/Datatoken.ts#L788-L790

**Impact:**
- Owner need to set himself as a `paymentManager` first in Ocean.js before able to call `setPaymentCollector()`. 

**Expectation:**
- I should be able to call `setPaymentCollector()` as an `Owner` according to the smart contract:
https://github.com/oceanprotocol/contracts/blob/249c9b2d51c79a95dce5f427e33a1a648c4d8576/contracts/templates/ERC20TemplateEnterprise.sol#L696-L701

**Possible Solution:**
- Add a  `||` OR condition checking to see if the `address` caller is not the `owner`, then throw similar error as in the smart contract: `Caller is not Fee Manager or Owner` 
- Hence, `Owner` will be able to call `setPaymentCollector()` from Ocean.js

soonhuat:
Fixes #1341 

Changes proposed in this PR:

- setPaymentCollector: implement same condition checking as smart contract
- allow if wallet are either payment manager OR **nft owner OR erc20 deployer**

same condition checking as smart contract method below
image

### DIFF
--- a/src/tokens/Datatoken.ts
+++ b/src/tokens/Datatoken.ts
@@ -785,8 +785,8 @@ export class Datatoken {
       new this.web3.eth.Contract(this.datatokensAbi, dtAddress),
       this.config
     )
-    const isPaymentManager =
-      (await this.getDTPermissions(dtAddress, address)).paymentManager === true
+    const isPaymentManager = (await this.getDTPermissions(dtAddress, address))
+      .paymentManager
     const nftAddress = !isPaymentManager && (await this.getNFTAddress(dtAddress))
     const isNftOwner = nftAddress && (await this.nft.getNftOwner(nftAddress)) === address
     const nftPermissions =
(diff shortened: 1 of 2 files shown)


src/tokens/Datatoken.ts
```diff
@@ -785,8 +785,15 @@ export class Datatoken {
       new this.web3.eth.Contract(this.datatokensAbi, dtAddress),
       this.config
     )
-    if ((await this.getDTPermissions(dtAddress, address)).paymentManager !== true) {
-      throw new Error(`Caller is not Fee Manager`)
+    const isPaymentManager =
+      (await this.getDTPermissions(dtAddress, address)).paymentManager === true
```
mihaisc, line 789: if `paymentManager ` is already a bool why do we need the extra `===true` check ?

mihaisc (approved):

PR #1419, opened 2022-04-14
optional config to apply blockTimeout and gasFeeMultiplier

soonhuat:
Fixes # .

Changes proposed in this PR:

- add optional config parameter when init contract
- add optional config for token approve
- both allow market operator to set different transaction and multiple config


alexcos20:
@bogdanfazakas - is this what you need ?

bogdanfazakas:
> @bogdanfazakas - is this what you need ?

Will check

src/pools/ssContracts/SideStaking.ts
```diff
@@ -350,7 +369,11 @@ export class SideStaking {
     contractInstance?: Contract
   ): Promise<number> {
     const sideStaking =
-      contractInstance || new this.web3.eth.Contract(this.ssAbi as AbiItem[], ssAddress)
+      contractInstance ||
+      setContractDefaults(
```
bogdanfazakas, line 373: I think you can use also here the newly added `sideStakingContract()` method

src/pools/ssContracts/SideStaking.ts
```diff
@@ -286,7 +301,11 @@ export class SideStaking {
...
```
bogdanfazakas, line 305: same here as prev comment ☝️

src/pools/ssContracts/SideStaking.ts
```diff
@@ -21,11 +26,18 @@ export class SideStaking {
...
   }
 
+  private sideStakingContract(ssAddress: string) {
+    return setContractDefaults(
+      new this.web3.eth.Contract(this.ssAbi, ssAddress),
+      this.config
+    )
+  }
+
   async amountToUnits(token: string, amount: string): Promise<string> {
```
kremalicious, line 36: there are 3 places in the codebase with exactly the same code. Why? So if we do a change here do we need to change all other places too?

Screen Shot 2022-04-14 at 16 45 13

src/pools/ssContracts/SideStaking.ts
```diff
@@ -40,9 +52,9 @@ export class SideStaking {
 
   async unitsToAmount(token: string, amount: string): Promise<string> {
```
kremalicious, line 53: same here

Screen Shot 2022-04-14 at 16 46 00

src/pools/ssContracts/SideStaking.ts
kremalicious: not fully same code but this is just crazy confusing and there's no way for reviewer to know what the intention here was or why there are methods with exact same name in multiple files.

In any case this is then also populated to CodeClimate:

Screen Shot 2022-04-14 at 16 48 06

src/utils/TokenUtils.ts
```diff
@@ -47,18 +53,23 @@ export async function estApprove(
...
+  config?: Config
 ): Promise<TransactionReceipt | string> {
-  const tokenContract = new web3.eth.Contract(minAbi, tokenAddress)
+  const tokenContract = setContractDefaults(
+    new web3.eth.Contract(minAbi, tokenAddress),
+    config
+  )
   if (!force) {
-    const currentAllowence = await allowance(web3, tokenAddress, account, spender)
+    const currentAllowence = await allowance(web3, tokenAddress, account, spender, config)
```
kremalicious, line 72: typo:

```suggestion
    const currentAllowance = await allowance(web3, tokenAddress, account, spender, config)
```

kremalicious, line 72: typo:

```suggestion
    const currentAllowance = await allowance(web3, tokenAddress, account, spender, config)
```

soonhuat:
> No writeup of problem or intention so hard to review fully. Just from DX perspective it feels weird having to construct a full `Config` object to get like 1-3 values across. Sounds like this should have its own interface of values which actually influence anything in the `setContractDefaults` which seems to be just reassigning values. So looks like a deeper architectural problem and I guess was not intention of this PR to solve, so we just continue to hack in stuff to solve isolated issues. But for developer completely unclear what exact values the full `Config` then should have if I just want to change the `gasFeeMultiplier`.
> 
> Also, is this intended to be merged before contracts 1.0.0-alpha.29? Cause that version then has no vesting anymore but this PR here touches multiple vesting methods

@kremalicious  @bogdanfazakas 
wonder does this look good?
the vesting part, i just apply the same implementation, when alpha-29 in, will auto remove the area of changes right. would be fine either before or after alpha-29 merge in.

as for code climate, looks like mostly is the warning of > 4 parameter, and false positive of redundant code duplication.

AnaLoznianu:
@soonhuat The AMM pool removal PR: https://github.com/oceanprotocol/ocean.js/pull/1568 adds a lot of changes to the ocean.js lib. Once @bogdanfazakas merges that PR, please update this one if it is still relevant. Thank you!

soonhuat:
> @soonhuat The AMM pool removal PR: #1568 adds a lot of changes to the ocean.js lib. Once @bogdanfazakas merges that PR, please update this one if it is still relevant. Thank you!

@AnaLoznianu could abandon this PR, I created another PR from fork on top of eip-1559 working on by Bogdan
https://github.com/oceanprotocol/ocean.js/pull/1581

AnaLoznianu:
All right, thank you for the update. Closing this then

PR #1428, opened 2022-04-20
Creating set publishing market fee function

### LINKED ISSUE #1335: Add missing function `setPublishingMarketFee`
add corresponding function for  https://github.com/oceanprotocol/contracts/blob/cb8d35adf8b80e697c8ff16b3ae4f0ce572862e0/contracts/templates/ERC20TemplateEnterprise.sol#L755

jamiehewitt15:
Fixes #1335

Changes proposed in this PR:

- Writing function for `setPublishingMarketFee`
- Writing function for `getPublishingMarketFee`
- Adding tests for `setPublishingMarketFee`


src/tokens/Datatoken.ts
```diff
@@ -44,6 +44,12 @@ export interface DispenserParams {
   allowedSwapper?: string // only account that can ask tokens. set address(0) if not required
 }
 
+export interface PublishingMarketFee {
```
bogdanfazakas, line 47: I think we can move this to `Erc20.ts` file under `@types` folder where we also have `ConsumeMarketFee` interface

src/tokens/Datatoken.ts
```diff
@@ -1357,4 +1363,105 @@ export class Datatoken {
...
+    datatokenAddress: string,
+    publishMarketFeeAddress: string,
+    publishMarketFeeToken: string,
+    publishMarketFeeAmount: string,
+    address: string
+  ) {
+    const dtContract = new this.web3.eth.Contract(this.datatokensAbi, datatokenAddress, {
+      from: address
+    })
+    const estGas = await this.estGasSetPublishingMarketFee(
```
bogdanfazakas, line 1423: it would be nice to check if the caller has the `publishMarketFeeAddress` role and throw an error if that's not the case. Not 100% sure that is possible, think we did that for `deployERC20` role for some methods

jamiehewitt15: The smart contract already checks this but we could add an additional check here too

bogdanfazakas (approved):

PR #1430, opened 2022-04-20
Issue-#1425: Use proper config in constructors

### LINKED ISSUE #1425: Use proper config in constructors
We should fetch the correct config based on chainId (not the first one) in constructors https://github.com/oceanprotocol/ocean.js/blob/a92ad7115bbb3c94ecbb4f9d59d297b162df9ace/src/pools/balancer/Pool.ts#L47

miquelcabot:
Fixes #1425 

Changes proposed in this PR:
- use `getConfig(network)` function to get proper configuration


mihaisc (approved):
good to go

Issue #1354, opened 2022-03-17
READMEs as integration tests


AnaLoznianu:

AnaLoznianu:
Related ocean.py issue: https://github.com/oceanprotocol/ocean.py/issues/743

jamiehewitt15:
My proposal is to simplify all of our documentation to just two files: the `readme.md` and a `codeExamples.md` file which will be auto-generated from an integration test.

- The readme can be for a general introduction and simple instructions on how to install and import the library. It should not contain any code examples.
- The codeExamples integration test should cover all of our main use cases:
  - [ ] Publishing a fixed price asset
  - [ ] Publishing an asset with a pool
  - [ ] Publishing an asset with free pricing 
  - [ ] Consuming each of those assets
  - [ ] Adding liquidity to a pool
  - [ ] Swapping Ocean tokens for a datatoken
- The `codeExamples.test.ts` file will be converted into markdown with a bash script
- The current docs folder and all of its content will be deleted.

AnaLoznianu:
@jamiehewitt15 I like this approach!  Please keep in mind that even though right now we don't have a C2D readme, we need to have examples of this functionality as well

jamiehewitt15:
Yeah, maybe it will make sense to put the compute to data examples in a separate file. That's a reasonable place to introduce some separation

Issue #1193, opened 2022-01-04
"Simple Flow" README works in v4


AnaLoznianu:
The v4 version of this flow
https://github.com/oceanprotocol/ocean.js#simple-flow should run end-to-end without errors.

Write the README as integration tests, and if needed, underlying components.
We can use some of these examples: https://github.com/akshay-ap/Ocean.js-PlayGround/tree/v4

PR #1438, opened 2022-04-25
Creating code examples file + updates to readme

### LINKED ISSUE #1437: "Marketplace flow" integration test for v4
Write a "Marketplace flow" integration test for v4, that will be also used as part of the READMEs.

miquelcabot:
Fixes: #1437
Fixes: #1354
Fixes: #1193

Changes proposed in this PR:
- Add a "Code Examples" integration test for v4
- Adding a script to convert the integration test to markdown
- Updating main Readme file
- Updating the CI so that it auto-generates the COdeExamples.md file and commits the changes


test/integration/MarketplaceFlow.test.ts
```diff
@@ -0,0 +1,404 @@
...
+  })
+  /// ```
+
+  it('Deploy contracts', async () => {
+    /// ```Typescript
+    contracts = await deployContracts(web3, publisherAccount)
+  })
+  /// ```
+
+  it('Publish a dataset (create NFT + ERC20) with a liquidity pool', async () => {
```
jamiehewitt15, line 187: I would expand this so that you create two more assets. One with fixed pricing and one with free pricing

jamiehewitt15:
It would be great if we ultimately get all of the different user journeys into the same integration test file and then it will end up in the same markdown file. That should help reduce duplication and make it easier to maintain

test/integration/MarketplaceFlow.test.ts
miquelcabot: @jamiehewitt15 Done!

bogdanfazakas (commented):
Looks really good 🙌 , lots of details in the the CodeExamples test file. 
One thing that I'm not really sure about is if we need to comit the resulted markdown file (`CodeExamples.md`), I think it would be better if we can run the script in the github CI  script. This way we won't need to maintain and commit two files when we update the test file. Maybe we can also keep just the createCodeExamples.sh file this way because the markdown will be generated in the CI so we won't need the mac sh file

jamiehewitt15:
> Looks really good raised_hands , lots of details in the the CodeExamples test file. One thing that I'm not really sure about is if we need to comit the resulted markdown file (`CodeExamples.md`), I think it would be better if we can run the script in the github CI script. This way we won't need to maintain and commit two files when we update the test file. Maybe we can also keep just the createCodeExamples.sh file this way because the markdown will be generated in the CI so we won't need the mac sh file

Yeah, good point. We should definitely update the CI so it automates the creation of the markdown file. We can also add CodeExamples.md to the gitignore file so that no one accidentally starts editing it locally

jamiehewitt15:
I'm just thinking this through, if the script is run as part of the CI then we will need another commit after the markdown file is created. The CI could do a commit on its own, is that what you were thinking @bogdanfazakas? If we do it that way, I'm wondering if it should be on the same branch or a different one that can be merged separately

jamiehewitt15:
@bogdanfazakas this is ready for another review. I've written a bash script that compares the checksum of the CodeExamples.md file before and after running `npm run create:guide` and it will only do a commit if there are changes

bogdanfazakas (approved):
Looks good to me 🎊

PR #1448, opened 2022-04-29
bump to contracts alpha 31

alexcos20:
Changes proposed in this PR:

-  remove joinPool & exitPool
- use barge env to set contracts, until barge default contract version will be alpha.31


mihaisc:
It seems there is still a vesting related test

alexcos20:
on it.  strange, because on localhost I have ```
  253 passing (1m)
```

Issue #1474, opened 2022-05-17
Refactoring


mihaisc:
We need a better library. The goal is for js developers to have a good experience. 

Topics: 
- do all those classes make sense?
- we need to restructure the files. We copied the structure from the contracts, we don't want that
- we need better flows/processes in place for building the packages so we can have an small optimized library

miquelcabot:
First proposals for the refactoring:
1. Refactor folder names and file position in folders. `\pools` folder must be renamed
2. Move `ConfigHelper` from to `\utils` to `\config`
3. Move `interface` types to `\@types`
4. Simplify imports in `\src` files
5. Are the ABI parameters necessary in Class constructors? Or can we use the default ABI?
6. Move `getErcCreationParams()` function to `NFTFactory`. It’s only used in this class
7. Move `getFreOrderParams()` function to `Datatoken`. It’s only used in this class
8. Move `getFreCreationParams()` function to `Datatoken`. It’s only used in this class
9. Move `getPoolCreationParams()` function to `Datatoken`. It’s only used in this class
10. Move `getData()` function to `Provider`. It’s only used in this class
11. Move all `\utils\ConversionTypeHelper` functions to `Provider`. They are only used in this class
12. Move all `\utils\PoolHelper` functions to `Pool`. They are only used in this class
13. Remove `generateDid()` functions. It’s not used
14. Remove `fetchData()` functions. It’s not used
15. Remove `downloadFileBrowser()` functions. It’s not used
16. Remove `postWithHeaders()` functions. It’s not used
17. Remove `postData()` functions. It’s not used
18. Move `downloadFile()` function to `/utils/General`
19. Add `decimals()` function to `/utils/TokenUtils.ts`
20. Add `mint()` function to `/utils/TokenUtils.ts`

Perhaps 19 and 20 can be done in a different Pull Request.

mihaisc:
5. Not sure what the use of that was. If one changes the contracts they would need a changed library. @alexcos20 do you remember what's the point of ABI ?
13. It is used in the market
15. Used in the market

miquelcabot:
@mihaisc Do you refer to 6, 7 or 13, 14, 15?
I have reviewed it and yes, `generateDid()`, `fetchData()` and `downloadFileBrowser()` are used in the Market

mihaisc:
Yes, sry, i am refering to 13 to 17, don't need to remove them

miquelcabot:
More proposals:
[...]

PR #1492, opened 2022-06-03
Issue-#1474: Refactoring (1). Restructure file location

miquelcabot:
1st PR for issue 1474 - refactoring.

Changes proposed in this PR:
- Restructure file location


bogdanfazakas:
Think the contracts/pools folder also needs some more restructuring, not all the files inside the folder are related to pools. But if we want just to copy the structure from contracts repo should be fine

bogdanfazakas (approved):

PR #1509, opened 2022-06-13
Issue-#1474: Refactoring (9): Refactor estimategas. Change from function to parameter in real functions

miquelcabot:
9th PR for issue 1474 - refactoring.

Changes proposed in this PR:
- Change the `estimategas()` specific functions. Instead of using a specific function for every task, use a parameter in the functions that let calculate the estimated gas instead of executing the transaction.
For example, with this function signature:
```
public async addNFTTemplate<G extends boolean = false>(
    address: string,
    templateAddress: string,
    estimateGas?: G
  ): Promise<G extends false ? TransactionReceipt : number> {
```
By default, it executes the transaction and returns a `TransactionReceipt` value. But if we set the `estimateGas` parameter to `true`, it returns a number representing the gas cost, without executing the transaction.


.eslintrc
```diff
@@ -16,7 +16,8 @@
   "rules": {
     "no-empty": ["error", { "allowEmptyCatch": true }],
     "prefer-destructuring": ["warn", { "object": true, "array": false }],
-    "no-dupe-class-members": ["warn"],
+    "no-dupe-class-members": "off",
+    "no-redeclare": "off",
```
bogdanfazakas, line 20: not pretty sure why we should add this rule that can lead to confusion to where the variable is declared and initialised

bogdanfazakas (approved):

PR #1510, opened 2022-06-13
update to new schema & compute fixes

alexcos20:
Changes proposed in this PR:

- update all tests with new ddo schema 4.1.0
- fix all known compute issues with ordering, reusing orders , providerFees, paid/free compute envs, etc
- add tests for above


src/@types/File.ts
```diff
@@ -0,0 +1,27 @@
+export interface UrlFiles {
```
kremalicious, line 1: this describes a single object so should not be plural. Guess `File` clashes with Node.js-native `File`?

alexcos20: yeah, we might have some issue later

alexcos20:
I will add more compute tests shortly

src/tokens/Datatoken.ts
```diff
@@ -143,6 +143,36 @@ export class Datatoken {
     return trxReceipt
   }
 
+  /**
+   * ApproveWei
+   * @param {String} dtAddress Datatoken address
+   * @param {String} spender Spender address
```
kremalicious, line 149: What is this method for? We have a generic `approve` method over in `TokenUtils.ts` and it uses wei

alexcos20: providerFeeAmount is always expressed in Wei, so when paying that fee, we need to approve first.
Converting from Wei to Decimal and then back to Wei will most likely cause some rounding errors, which will result in a fail validation in https://github.com/oceanprotocol/contracts/blob/v4main/contracts/templates/ERC20Template.sol#L453 , because the signature is created using the Wei amount

alexcos20: as usage, it's here:  https://github.com/oceanprotocol/ocean.js/blob/feature/ddo_schema_4.1.0/test/integration/ComputeFlow.test.ts#L312

(handleOrder function is going to be most likely duplicated in the market, it's not ocean.js specific, but more app)

alexcos20: about gasEstimate:     Would be nice, but technically speaking,  
     estimateGas (approve( 0x11, 100))  or 
     estimateGas (approve( 0x11, 1000000000000000))
 costs the same, so you don't need Wei and non-Wei, the result will be the same..

mihaisc: This function should be in token utils and indeed the one that already exists has bad comments , it doesn't accept wei, but a number (as string) . It doesn't hurt to have one that takes wei directly just in case.

src/@types/File.ts
```diff
@@ -0,0 +1,27 @@
...
+   * File index.
+   * @type {number}
+   */
+  index?: number
+
+  /**
+   * File URL.
+   * @type {string}
+   */
+  url?: string
```
mihaisc, line 14: why is this nullable? can this exist without an url?

alexcos20: You are not going to get the url when calling provider fileInfo :)

alexcos20:
since this is not needed for our push, let's keep it until I will fix the provider for compute

src/@types/File.ts
mihaisc: @alexcos20 this is for encryption, `FileInfo` is what you get from provider.

alexcos20: fixed, I was confused.  both url & method are required

PR #1524, opened 2022-06-21
Issue-#1474: Refactoring(11): Review try catch clauses

miquelcabot:
11th PR for issue 1474 - refactoring.

Changes proposed in this PR:
- Review `try...catch` clauses


src/contracts/pools/FixedRateExchange.ts
```diff
@@ -0,0 +1,773 @@
...
+      this.contract.methods.buyDT,
+      exchangeId,
+      dtAmountFormatted,
+      maxBtFormatted,
+      consumeMarketAddress,
+      consumeMarketFeeFormatted
+    )
+    if (estimateGas) return estGas
+
+    const trxReceipt = await this.contract.methods
```
bogdanfazakas, line 69: not sure why the try catch clause was removed from here , maybe the error logs would be helpful

src/utils/DdoHelpers.ts
```diff
@@ -1,17 +1,12 @@
...
   return `did:op:${checksum.toString()}`
 }
 
 export function getHash(data: any): string {
-  try {
-    return sha256(data).toString()
-  } catch (e) {
-    LoggerInstance.error('getHash error: ', e.message)
-  }
+  return sha256(data).toString()
```
bogdanfazakas, line 11: since this can fail think if the data send is not well formed, think we should keep the clause and have some error logs at least

src/contracts/tokens/NFT.ts
```diff
@@ -0,0 +1,744 @@
...
+        [this.web3.utils.toWei(cap), this.web3.utils.toWei(feeAmount)],
+        []
+      )
+      .send({
+        from: address,
+        gas: estGas + 1,
+        gasPrice: await this.getFairGasPrice()
+      })
+
+    return trxReceipt.events.TokenCreated.returnValues[0]
```
bogdanfazakas, line 81: worth having optional chaining here also

bogdanfazakas (approved):

PR #1537, opened 2022-06-28
Issue-#1195: "C2D Flow" README works in v4

### LINKED ISSUE #1195: "C2D Flow" README works in v4
The v4 version of this flow
https://github.com/oceanprotocol/ocean.py#compute-to-data-flow

should run end-to-end without errors.

This means updating the README, and if needed, underlying components.

miquelcabot:
Fixes #1195 

Changes proposed in this PR:
- Add `C2DExamples.test.ts` test script
- Generate automatically `C2DExamples.md`markdown file


C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+```bash
+mkdir marketplace-quickstart
+cd marketplace-quickstart
+npm init
+# Answer the questions in the command line prompt
+cat > marketplace.js
+# On linux press CTRL + D to save
+```
+
+## 3. Install dependancies
```
kremalicious, line 52: typo, it's "dependencies". Needs to change throughout whole document

C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+
+Open the package.json file in a text editor and update the dependancies to include the following:
```
kremalicious, line 54: this is not how you install dependencies in npm world, and you are telling people to install outdated dependencies which do not work anymore. This should be replaced with:

```bash
npm i @oceanprotocol/contracts @oceanprotocol/lib crypto-js web3
```

C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+        }
+      }
+    },
+    services: [
+      {
+        id: 'testFakeId',
+        type: 'access',
+        files: '',
+        datatokenAddress: '0x0',
+        serviceEndpoint: 'https://providerv4.rinkeby.oceanprotocol.com',
```
kremalicious, line 262: this endpoint does not exist

C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+      author: 'oceanprotocol-team',
+      license: 'MIT'
+    },
+    services: [
+      {
+        id: 'testFakeId',
+        type: 'compute',
+        files: '',
+        datatokenAddress: '0x0',
+        serviceEndpoint: 'https://providerv4.rinkeby.oceanprotocol.com',
```
kremalicious, line 219: this endpoint does not exist

C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+    consumerAccount = accounts[1]
+```
+Again, lets console log the values so that we can check that they have been saved properly
+```Typescript
+    console.log(`Publisher account address: ${publisherAccount}`)
+    console.log(`Consumer account address: ${consumerAccount}`)
+  
+```
+
+  ### 5.2 Next, lets get the address of the deployed contracts
```
kremalicious, line 297: can we please not mix up headings and instructional texts. Suddenly here all the headings are what should be just the explaining copy. Find actual headings for those

C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+  transfer,
+  ComputeAsset,
+  ComputeAlgorithm,
+  ComputeJob,
+  sleep,
+  ProviderComputeInitialize,
+  ConsumeMarketFee,
+  approveWei
+} from '../../src'
+import { web3, getTestConfig, getAddresses } from '../config'
```
kremalicious, line 100: where is this `../config` file?

scripts/createC2DExamples-mac.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# Create markdown file
+cp test/integration/C2DExamples.test.ts C2DExamples.md
```
kremalicious, line 3: why does this file exist twice, indicating that macOS needs different build process but then both files are the same? There should be no multiple different build processes but rather one script dealing with OS differences inside of it

C2DExamples.md
miquelcabot: @kremalicious We have this endpoint in all of the `ocean.js` tests. Which is the correct one?

package.json
```diff
@@ -43,9 +43,9 @@
     "test:integration": "npm run mocha -- 'test/integration/**/*.test.ts'",
     "test:compute": "npm run mocha -- 'test/integration/ComputeFlow.test.ts'",
     "test:integration:cover": "nyc --report-dir coverage/integration --exclude 'src/@types/**/*' --no-clean npm run test:integration",
-    "create:guide": "chmod +x ./scripts/createCodeExamples.sh && ./scripts/createCodeExamples.sh",
-    "create:guide:mac": "chmod +x ./scripts/createCodeExamples-mac.sh && ./scripts/createCodeExamples-mac.sh",
-    "commit:guide": "chmod +x scripts/commitChanges.sh && scripts/commitChanges.sh"
+    "create:guide": "chmod +x ./scripts/createCodeExamples.sh && ./scripts/createCodeExamples.sh test/integration/CodeExamples.test.ts",
```
kremalicious, line 46: doing `chmod +x` every time this runs seems unnecessary and I doubt this works on Windows. Do this locally once, then commit the file as file attributes are version controlled so in end you commit the executable script

C2DExamples.md
miquelcabot: @kremalicious Solved https://github.com/oceanprotocol/ocean.js/pull/1537/commits/45ee71f71e7e3d46e38284e02f214511897f7bda

C2DExamples.md
miquelcabot: @kremalicious Solved https://github.com/oceanprotocol/ocean.js/pull/1537/commits/45ee71f71e7e3d46e38284e02f214511897f7bda

C2DExamples.md
miquelcabot: @kremalicious Solved. https://github.com/oceanprotocol/ocean.js/pull/1537/commits/6eabc2ecd375789445274295d407523cb26497b0

C2DExamples.md
miquelcabot: @kremalicious Solved https://github.com/oceanprotocol/ocean.js/pull/1537/commits/d8e9d8d450a44edfe9c296d0e8ca845964d52a3c

package.json
miquelcabot: @kremalicious Solved. The +x attributes are already saved. https://github.com/oceanprotocol/ocean.js/pull/1537/commits/c74e993e589a54507005921c685dcbaaa1880a82

C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+  
+```
+
+  ### 5.3 We send some OCEAN to consumer account
+```Typescript
+    transfer(web3, publisherAccount, addresses.Ocean, consumerAccount, '100')
+  
+```
+
+## 6. Publish a dataset (Data NFT and Datatoken)
```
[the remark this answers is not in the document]
miquelcabot: @kremalicious Solved. https://github.com/oceanprotocol/ocean.js/pull/1537/commits/e4131790cbc7ed16e90da817d50b874599773384

C2DExamples.md
```diff
@@ -0,0 +1,616 @@
...
+  Files,
+  transfer,
+  ComputeAsset,
+  ComputeAlgorithm,
+  ComputeJob,
+  sleep,
+  ProviderComputeInitialize,
+  ConsumeMarketFee,
+  approveWei
+} from '../../src'
```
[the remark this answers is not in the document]
miquelcabot: @kremalicious Fixed https://github.com/oceanprotocol/ocean.js/pull/1537/commits/f8dece20108b228242e941ff8be7ff60af19627e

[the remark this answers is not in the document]
miquelcabot: @kremalicious Fixed https://github.com/oceanprotocol/ocean.js/pull/1537/commits/f8dece20108b228242e941ff8be7ff60af19627e

C2DExamples.md
miquelcabot: @kremalicious With @jamiehewitt15 we agree that these titles come from the conversion of the titles of the `it` tests. I have reviewed these titles in https://github.com/oceanprotocol/ocean.js/pull/1537/commits/49a85ff23adc40a9385f8dfe033d54dfc56a4f7f

C2DExamples.md
miquelcabot: @kremalicious I have linked an example of the `../config` file in https://github.com/oceanprotocol/ocean.js/pull/1537/commits/295f94ed1294b483680f3dbf772875af1928754a

idiom-bytes:
@miquelcabot I have updated the workflow so:
- user touches marketplace.json
- sed matches and replace the import config path.

idiom-bytes:
Hi @miquelcabot and @jamiehewitt15 I tagged both of you as I saw your names in the github annotations. I'm reviewing the script for createCodeExamples.sh and I'm continuing to run into issues: 
- Tutorial suggest user to create a .js file but all the inner workings are in .ts
- shell script is using sed to remove chai test functions such as before() and it(), but it fails to properly substitute for an async function
- There is nothing in the documentation that tells the user he needs to compile from .ts => .js

I understand the overall drive to want to go from an internal .ts test => clean documentation, but this seems like more effort than it's worth and perhaps fragile too. I.E. Why not just have the documentation.md updated?

Let me know, cheers!

idiom-bytes:
Here is another issues.

The following import is local to tests. Meaning that the config helper functions are not bundled/exported with the module.
```
import { getAddresses, getTestConfig, web3 } from '../config'
```

Downstream, when the user is going through the marketplace or C2D flow, he will try to import this from @oceanprotocol/lib, they will be unable to.

C2DExamples.md
```diff
@@ -0,0 +1,605 @@
...
+## 2. Create a new node.js project
+
+Start by creating a new Node.js project. Open a new terminal and enter the following commands:
+
+```bash
+mkdir marketplace-quickstart
+cd marketplace-quickstart
+npm init
+# Answer the questions in the command line prompt
+touch marketplace.ts
```
jamiehewitt15, line 48: If we are now advising the user to write this in typescript we need to add an extra step for creating a `tsconfig.json` file + we need to provide a command for compiling and running the code.

CodeExamples.md
```diff
@@ -55,32 +55,21 @@ mkdir marketplace-quickstart
 cd marketplace-quickstart
 npm init
 # Answer the questions in the command line prompt
-cat > marketplace.js
+touch marketplace.ts
```
jamiehewitt15, line 58: Again, we need to inform the user about how to compile and run the code if we are advising them to use typescript

idiom-bytes: I mentioned that above too, amongst a variety of other problems.
```
There is nothing in the documentation that tells what the user needs to compile from .ts => .js, and run it. If the user tries to use node marketplace.js he will receive ES6 import-related errors.
```

**I believe the meta is the issue here is that this bit of code, is trying to become a generic exporter that goes from .ts => .md** Having a pipeline that goes from test.ts +> tutorial.md, is neat, but seems like a complete distraction.

**My recommendation:**
Forget building the script that generates the markdown. Just update the markdown/tutorial as needed so it can remain functional.

jamiehewitt15: In retrospect, it would have been easier if the test was written in js rather than ts as that just adds an additional step. The PR has improved a lot since you did your review above though so I don't think we should abandon it at this point

jamiehewitt15:
@miquelcabot do you have an idea of why the tests are failing?

kremalicious (dismissed):

kremalicious (changes requested):
what is not the best solution is that we now start mixing ES5 js with TypeScript which is a pain to maintain. The code examples and c2d examples have been now written in ES5 but within the same folder structure as all the rest. This is a TypeScript project so all source files need to be written in TypeScript. Compilation to ES5 then is another step but writing the source in ES5 will lead to crazy maintenance problems in future.

By now tending towards @idiom-bytes's suggestion of simply writing stuff in the markdown files and be done with it.

It's a noble goal of autogenerating this all and have it tested, but this PR is not finalized for this. If we want to have this, the source needs to be in TypeScript and then figured out how to transpile this. Would also argue that the code examples need to be in ES6 syntax as most users will have a build process. It's a weird mix we would introduce with current state of PR

jamiehewitt15:
> By now tending towards @idiom-bytes's suggestion of simply writing stuff in the markdown files and be done with it.

If that's the direction we go in then we could just abandon this altogether and then make sure we're covering everything properly in the docs. The readme here can be updated with links to the relevant tutorials, similar to how we have done it in the [market](https://github.com/oceanprotocol/market#-forking)

bogdanfazakas:
Closing this since it is handled in #1615

PR #1555, opened 2022-07-08
get files checksum using fileInfo helper

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- add withChecksum param to fileinfo provider request
- add checksum attribute to FIleInfo type

### DIFF
--- a/src/@types/FileInfo.ts
+++ b/src/@types/FileInfo.ts
@@ -37,6 +37,12 @@ export interface FileInfo {
   method?: string
 
   /**
+   * Computed file checksum
+   * @type {string}
+   */
+  checksum?: string
+
+  /**
    * check if file exists
    * @type {boolean}
    */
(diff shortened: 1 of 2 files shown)


src/@types/FileInfo.ts
```diff
@@ -36,6 +36,12 @@ export interface FileInfo {
    */
   method?: string
 
+  /**
+   * Computed file checksum
+   * @type {string}
+   */
+  checksum?: string
```
kremalicious, line 43: I don't understand why everything in this typing is optional, and now also the `checksum`. Based on the typings this is the only response I should expect:

```js
const { type } = getFileInfo()
```

Because everything else is optional. What exactly am I supposed to do with the `type` alone? Why would all the other fields be empty int he response?

mihaisc (approved):

src/@types/FileInfo.ts
bogdanfazakas: yeah they kind of are optionals based on various responses we receive for provider when we query this. As i understood we will break this into smaller types and work with inheritance once we are going to integrate other storage options 😃

PR #1563, opened 2022-07-20
fix download

mihaisc:
Prevents malicious redirects from custom providers


bogdanfazakas (approved):
LGTM, also tested the build with local version of the market and seems all good

PR #1568, opened 2022-07-28
Remove amm pools

### LINKED ISSUE #1562: Remove AMM pools functionality
As we are moving away from the dynamic pricing schema, we need to remove everything POOL related.
- [ ] Remove POOL related code
- [ ] Update READMEs

bogdanfazakas:
Fixes #1562 

Changes proposed in this PR:

- Removes AMM pools functionality and related code from source files and tests
- Update Readmes


src/contracts/pools/Router.ts
```diff
@@ -115,40 +108,6 @@ export class Router extends SmartContractWithAddress {
     return trxReceipt
   }
 
-  /**
-   * Remove a token from oceanTokens list, pools without baseToken in this list have a opf Fee
-   * @param {String} address
-   * @param {String} tokenAddress address to remove
-   * @return {Promise<TransactionReceipt>}
-   */
-  public async removeApprovedToken<G extends boolean = false>(
```
kremalicious, line 124: are we sure about this removal. Cause there's still `addApprovedToken()`? Not really clear based on code comments or contract method call where exactly the token is added/removed

mihaisc: this is not related to pools. This needs to stay.

kremalicious:
going radical like over in market and using CI for test so things might be broken until I get it green again. No need to keep multiple `pool/` folders cause the contracts have that. Flattened all those folder structures in `contracts/` and ``test/unit/` as if every folder has only 1 file there should be no folder

src/contracts/pools/Router.ts
mihaisc: don't really remove anything from Router.

kremalicious:
all green again, remaining things are all around that `Router`, like finishing the test for it and possibly cleaning up its typings.

Checked if we can remove some dependencies and figured we use bignumber.js mostly for fixed rate contract calculations & decimal.js for the dispenser contract. Most likely we could decide on one library as they should have sufficient overlap.

The above mentioned `addDispenserContract()`, ... seems to be only used in tests so we could look into not exposing them and have it only in tests.

mihaisc:
`addDispenserContract` is used only by the contract owner ( aka us ).The comments are wrong or incomplete for some functions , other than this good progress.

mihaisc:
Does the `contracts/tokens` make sense? `NFTFactory` should not be there so you then have a folder just for erc20 and nft. All could just sit in the `contracts` folder

kremalicious (approved):
green, yay!

PR #1579, opened 2022-08-10
Eip 1559 support

### LINKED ISSUE #1566: Add EIP-1559 support
- [ ] detect if the blockchain supports the EIP-1559 
- [ ] if it is supported, use the new fee structure

bogdanfazakas:
Fixes #1566  .

Changes proposed in this PR:

- added suport for eip 1559
- updated send calls to use a method from contract utils


src/utils/ContractUtils.ts
```diff
@@ -16,6 +16,15 @@ export function setContractDefaults(contract: Contract, config: Config): Contrac
   return contract
 }
 
+export async function networkSupportsEip1559(web3: Web3): Promise<boolean> {
+  const feeHistory = await web3.eth.getFeeHistory(4, 'pending', [25, 50, 75])
+  if (feeHistory && feeHistory.baseFeePerGas) {
```
kremalicious, line 21: way shorter and only one return:

```js
return feeHistory?.baseFeePerGas ? true : false
```

src/utils/ContractUtils.ts
```diff
@@ -82,3 +91,48 @@ export async function calculateEstimatedGas(
     .estimateGas({ from }, (err, estGas) => (err ? GASLIMIT_DEFAULT : estGas))
   return estimatedGas
 }
+
+/**
+ * Send the transation on chain
+ * @param {string} from account that calls the function
+ * @param {any} estGas estimated gas for the transaction
+ * @param {Web3} web3 web3 objcet
+ * @param {Function} functionToEstimateGas function that we need to send
```
kremalicious, line 100: doc & code mismatch (`functionToEstimateGas` vs `functionToSend`)

src/utils/ContractUtils.ts
```diff
@@ -82,3 +91,48 @@ export async function calculateEstimatedGas(
...
+ * @param {...any[]} args arguments of the function
+ * @return {Promise<any>} transaction receipt
+ */
+export async function sendTx(
+  from: string,
+  estGas: any,
+  web3: Web3,
+  config: Config,
+  functionToSend: Function,
+  ...args: any[]
```
kremalicious, line 110: this is not so helpful but not sure we can figure out more precise typings here? As those typings imply I can send whatever after the first params

src/utils/ContractUtils.ts
```diff
@@ -82,3 +91,48 @@ export async function calculateEstimatedGas(
...
+ * @param {string} from account that calls the function
+ * @param {any} estGas estimated gas for the transaction
+ * @param {Web3} web3 web3 objcet
+ * @param {Function} functionToEstimateGas function that we need to send
+ * @param {...any[]} args arguments of the function
+ * @return {Promise<any>} transaction receipt
+ */
+export async function sendTx(
+  from: string,
+  estGas: any,
```
kremalicious, line 106: as far as I see, this is always a number in our code so we should be able to be more precise than `any`

src/utils/ContractUtils.ts
kremalicious: well, they do it like that too, `any` is the solution :-) 

https://github.com/balancer-labs/frontend-v2/blob/develop/src/lib/utils/balancer/web3.ts#L28-L104

src/utils/ContractUtils.ts
```diff
@@ -82,3 +91,48 @@ export async function calculateEstimatedGas(
...
+  web3: Web3,
+  config: Config,
```
kremalicious, line 108: is there anything else used from this object than `config.gasFeeMultiplier`? If not, then this should be the param, and not the whole config object

src/utils/ContractUtils.ts
bogdanfazakas: aa forgot to delete this sorry my bad, the logic is integrated in the `sendTx` indeed

src/utils/ContractUtils.ts
bogdanfazakas: yep thought about having that instead of the whole object, will update this and the calls 😃 , think i was not sure if we need something else from config at the development time

src/utils/ContractUtils.ts
bogdanfazakas: hmm interesting will take a look at their approach maybe we can "inspire" a little and update the `catch` block

src/utils/ContractUtils.ts
bogdanfazakas: true, will update this and also will try to make the methods wrapper methods for estimate and send more restrictive on the return type

src/utils/ContractUtils.ts
bogdanfazakas: So basically our approach is quite different from the balancer one.
 - Firstly they call the send method using EIP1559 params and as a fallback they go back to trying the legacy one. On the other hand we first call web3js lib to get the fee history and call the send method only once. The `getFeeHistory` throws an error if the networks does not support EIP1559 transactions, and so we set the params for the legacy transaction instead of the EIP1559 transaction
 - Also they make https calls to various api's to get the fees and we get them from the web3js lib directly

kremalicious: Good research, thanks! Then it looks like we actually have the better solution 😀

Could be fine assuming if getting fee history fails, network doesn't support EIP 1559. Just thought that we might also get some dedicated response text in that case we can work with, like Balancer does, as they really isolate that error case.

So should be fine for now how we do it, we could do more once anybody actually runs into issues

PR #1614, opened 2022-09-22
add full support for Goerli

kremalicious:
This actually adds Goerli support after #1613 🤦‍♂️ 

- add goerli to ConfigHelper
- removes all usage of Rinkeby resources across codebase
- bump @oceanprotocol/contracts to v1.1.7

### DIFF
--- a/src/config/ConfigHelper.ts
+++ b/src/config/ConfigHelper.ts
@@ -36,6 +36,26 @@ export const configHelperNetworks: Config[] = [
   },
   {
     ...configHelperNetworksBase,
+    chainId: 3,
+    network: 'ropsten',
+    nodeUri: 'https://ropsten.infura.io/v3',
+    providerUri: 'https://v4.provider.ropsten.oceanprotocol.com',
+    subgraphUri: 'https://v4.subgraph.ropsten.oceanprotocol.com',
+    explorerUri: 'https://ropsten.etherscan.io',
+    gasFeeMultiplier: 1.1
+  },
+  {
+    ...configHelperNetworksBase,
+    chainId: 4,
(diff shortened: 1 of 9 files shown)


alexcos20 (approved):
:)

src/config/ConfigHelper.ts
```diff
@@ -36,22 +36,12 @@ export const configHelperNetworks: Config[] = [
...
-    nodeUri: 'https://ropsten.infura.io/v3',
-    providerUri: 'https://v4.provider.ropsten.oceanprotocol.com',
-    subgraphUri: 'https://v4.subgraph.ropsten.oceanprotocol.com',
-    explorerUri: 'https://ropsten.etherscan.io',
-    gasFeeMultiplier: 1.1
-  },
-  {
-    ...configHelperNetworksBase,
-    chainId: 4,
-    network: 'rinkeby',
```
kremalicious, line 50: I removed ropsten & rinkeby support here but not sure we should. It should definitely not be used anymore but this change would break e.g. current ve testing. If we're fine with that then we can keep them out of here.

bogdanfazakas (approved):

src/config/ConfigHelper.ts
alexcos20: wait. let's keep them for 2 more weeks, until 5th of oct

alexcos20 (changes requested):
let's keep rinkeby & ropsten until official remove

alexcos20:
can you bump contracts to 1.1.7 pls?  it was released a min ago

alexcos20 (approved):

PR #1630, opened 2022-10-04
add allocated

mihaisc:


alexcos20 (approved):

src/@types/Asset.ts
```diff
@@ -89,6 +89,20 @@ export interface AssetDatatoken {
...
+   * How often an asset was consumed, meaning how often it was either downloaded or used as part of a compute job.
+   * @type {number}
+   */
+  orders: number
+
+  /**
+   * Total amount of veOCEAN allocated on this asset.
+   * @type {number}
+   */
+  allocated?: number
```
kremalicious, line 103: why is it optional? Is it supposed to be on all assets, or not? Cause if it does't have allocations, it should have `0` as value

bogdanfazakas (approved):

PR #1633, opened 2022-10-10
Add ve addresses to config

mihaisc:


alexcos20 (approved):

kremalicious (approved):

src/config/Config.ts
```diff
@@ -170,4 +170,13 @@ export class Config {
    * @type {number}
    */
   gasFeeMultiplier: number
+
+  veAllocate?: string
```
kremalicious, line 174: should all work out. Just as usual the question why this is all marked as optional?

mihaisc: Yes, because it's not on all networks. It's just on mainnet, goerli , mumbai

PR #1636, opened 2022-10-19
add new nft state

mihaisc:
Added state based on https://github.com/oceanprotocol/market/issues/1747

### DIFF
--- a/src/@types/Asset.ts
+++ b/src/@types/Asset.ts
@@ -34,7 +34,7 @@ export interface AssetNft {
    * 4	Ordering is temporary disabled.
    * @type {number}
    */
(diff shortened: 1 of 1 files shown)


src/@types/Asset.ts
```diff
@@ -34,7 +34,7 @@ export interface AssetNft {
    * 4	Ordering is temporary disabled.
    * @type {number}
    */
-  state: 0 | 1 | 2 | 3 | 4
+  state: 0 | 1 | 2 | 3 | 4 | 5
```
bogdanfazakas, line 37: can you pls include this state in the jsdoc also, to have a short desc about this new state ?

bogdanfazakas (approved):

PR #1642, opened 2022-10-27
bring back provider headers and checksum from older oceanJs function

soonhuat:
Fixes # .

Changes proposed in this PR:

- bring back headers for provider getFileInfo
- bring back checksum flag for provider getFileInfo
- add back withChecksum signature comment

### DIFF
--- a/src/services/Provider.ts
+++ b/src/services/Provider.ts
@@ -150,6 +150,7 @@ export class Provider {
    * @param {string} did did
    * @param {number} serviceId the id of the service for which to check the files
    * @param {string} providerUri uri of the provider that will be used to check the file
+   * @param {boolean} withChecksum if true, will return checksum of files content
    * @param {AbortSignal} signal abort signal
    * @return {Promise<FileInfo[]>} urlDetails
    */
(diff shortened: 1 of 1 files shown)


src/services/Provider.ts
```diff
@@ -192,20 +193,24 @@ export class Provider {
...
+   * @param {any} headers headers key value pairs associated with the asset GET request
    * @return {Promise<FileInfo[]>} urlDetails
    */
   public async getFileInfo(
     file: UrlFile | Arweave,
     providerUri: string,
-    signal?: AbortSignal
+    withChecksum: boolean = false,
+    signal?: AbortSignal,
+    headers?: any
```
bogdanfazakas, line 206: can't we define a interface for this to avoid using `any` , with the desired HTTP get fields needed ?

bogdanfazakas:
it would be also nice to have a test in the `test/integration/Provider.test.ts` file that covers an example of `getFileInfo` using headers

alexcos20:
We should add headers to UrlFile and Graphql types

bogdanfazakas:
handled in #1665

PR #1651, opened 2022-11-14
Updating types to include the price which is sent from Aquarius

jamiehewitt15:
Currently getting typing errors in the market when trying to use the price from Aquarius as the Asset.Stats interface doesn't include the price object which is being sent from Aquarius. This PR is to update the typings. 

Required for: https://github.com/oceanprotocol/market/issues/1654

Changes proposed in this PR:

- Updating Asset type to include the price which is sent from Aquarius in Stats

### DIFF
--- a/src/@types/Asset.ts
+++ b/src/@types/Asset.ts
@@ -90,6 +90,11 @@ export interface AssetDatatoken {
   serviceId: string
 }
 
+export interface AssetPrice {
+  value: number
+  tokenSymbol?: string
+  tokenAddress?: string
+}
 export interface Stats {
   /**
    * How often an asset was consumed, meaning how often it was either downloaded or used as part of a compute job.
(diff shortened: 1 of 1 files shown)


src/@types/Asset.ts
```diff
@@ -90,6 +90,25 @@ export interface AssetDatatoken {
...
+   * The price of the asset expressed as a number. If 0 then the price is FREE.
+   * @type {number}
+   */
+  value: number
+
+  /**
+   * The symbol that the price of the asset is expressed in.
+   * @type {string}
+   */
+  tokenSymbol?: string
```
kremalicious, line 104: why is this optional? And also `tokenAddress` and the whole `price` object? Can I expect it to be present on all responses, or not? Marking it optional means that some requests will have it, and some don't

jamiehewitt15: Currently Aquarius doesn't send `tokenSymbol` or `tokenAddress` when the asset is free. It just sends ` "price": { "value": 0.0}` so they would need to remain optional unless we change the response that Aquarius sends. I'll make the whole price object required though as it should always be there

bogdanfazakas (approved):
LGTM

PR #1663, opened 2022-12-07
Updating test command in readme

jamiehewitt15:
Changes proposed in this PR:

- The readme currently has the wrong test command, this PR updates it


jamiehewitt15:
> Or create a npm script `test` which actually does what the text says, and then we don’t need to change the command here

Yeah, that's better. Presumably it was there before and got deleted by mistake.

PR #1666, opened 2022-12-13
Test new ganache

alexcos20:
Test new ganache from of barge (see https://github.com/oceanprotocol/barge/pull/342)


bogdanfazakas:
@alexcos20 i think we can close this since the ganache pr was merge long time ago in the barge repo

PR #1697, opened 2023-02-09
add macOS readme message

### LINKED ISSUE #1673: Add instructions in the readme's for running ocean.js with barge inside macOS
There are some network contraints preventing users running the integration tests and main flows like publish, consume or compute   out of the box with the configs returned for the barge network

bogdanfazakas:
Fixes #1673 .

Changes proposed in this PR:

- add note about running the integration tests on macOs
- use correct subgraph port

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,8 +128,6 @@ npm run test:integration
 npm run test:integration:cover
 ```
 
-Note: If you are running the integration tests on macOs pls make sure you update the
-
 Note: If you are running the integration tests operating system you should also do same changes to the provider, metadata cache and subgraph urls, since the default barge ip can not be accessed due to some network constraints on macOs. So we should be using the `http://127.0.0.1` for each direct call from the to the mentioned services, but we should keep the internal barge provider url (`http://172.15.0.4:8030`) hardcoded inside all ddo's `serviceEndpoint` and when calling `nft.setMetadata` method.
 
 ## 🛳 Production
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ npm run test:integration
 npm run test:integration:cover
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -128,6 +128,8 @@ npm run test:integration
 npm run test:integration:cover
 ```
 
+Note: If you are running the integration tests operating system you should also do same changes to the provider, metadata cache and subgraph urls, since the default barge ip can not be accessed due to some network constraints on macOs. So we should be using the `http://127.0.0.1` for each direct call from the to the mentioned services, but we should keep the internal barge provider url (`http://172.15.0.4:8030`) hardcoded inside all ddo's `serviceEndpoint` and when calling `nft.setMetadata` method.
```
kremalicious, line 131: let's actually mention macOS?

bogdanfazakas: oh yeah 👍  , i was just updating the text so many times and this part got removed somehow 😞

src/config/ConfigHelper.ts
```diff
@@ -31,9 +31,9 @@ export const configHelperNetworks: Config[] = [
     ...configHelperNetworksBase,
     chainId: 8996,
     network: 'development',
-    metadataCacheUri: 'http://172.15.0.5:5000',
```
jamiehewitt15, line 34: Why did you delete the slash `/` here?

bogdanfazakas: ow just a mistake 🤕

README.md
```diff
@@ -128,6 +128,8 @@ npm run test:integration
 npm run test:integration:cover
 ```
 
+Note: If you are running the integration tests operating system macOS you should also do same changes to the `provider`, `metadataCache` and `subgraph` urls, since the default barge ip can not be accessed due to some network constraints on macOS. So we should be using the `http://127.0.0.1` for each direct call from the to the mentioned services, but we should keep the internal barge provider url (`http://172.15.0.4:8030`) hardcoded inside all ddo's `serviceEndpoint` and when calling `nft.setMetadata` method.
```
kremalicious, line 131:
```suggestion
> Note: On macOS, changes to the `provider`, `metadataCache` and `subgraph` URLs are required, as their default `barge` IPs can not be accessed due to network constraints on macOS. Instead use `http://127.0.0.1` for each direct call to the mentioned services, but keep the internal `provider` URL (`http://172.15.0.4:8030`) hardcoded inside all DDO's `serviceEndpoint`, and when calling `nft.setMetadata()`.
```

jamiehewitt15 (approved):
Looks good to me

kremalicious (approved):

PR #1723, opened 2023-04-25
Using typedoc to autogenerate documentation

jamiehewitt15:
Fixes # .

Changes proposed in this PR:

- Using typedoc to autogenerate documentation
- 
-

### DIFF
--- a/docs/.nojekyll
+++ b/docs/.nojekyll
@@ -1,0 +1,1 @@
+TypeDoc added this file to prevent GitHub Pages from using Jekyll. You can turn off this behavior by setting the `githubPages` option to false.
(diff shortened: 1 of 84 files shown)


bogdanfazakas:
@jamiehewitt15 in the `feature/ethers` branch i've updated a lot of jsDocs , i think typedoc is not strictly direct connected to jsdocs but is using the method signature and the comments. 
The question would be is it worth having this PR targeting main or should we use the mentioned branch above since it refactors pretty heavily the oceanjs methods and has updated jsdoc comments. 
If we just need to run the script that generates the docs once more after we merge the ethers PR then definitely worth keeping it targeting main.

jamiehewitt15:
> jsDocs

Good point, it's worth testing out TypeDoc with your latest changes. I'll open a new PR for that

jamiehewitt15:
Superseded by: https://github.com/oceanprotocol/ocean.js/pull/1724

PR #1724, opened 2023-04-26
Adding docs with Typedoc

jamiehewitt15:
Fixes https://github.com/oceanprotocol/docs/issues/1202

Changes proposed in this PR:

- Adding docs with Typedoc
- Updating CI to automatically update the docs folder.


jamiehewitt15:
> also are we going to keep this triggered only at release, or we should update the docs to be generated more often like merge in main or smth?

Yeah, good point. I can add it as part of the CI, like we do with the autogenerated guides

bogdanfazakas (approved):
lgtm

PR #1730, opened 2023-05-11
Add consumer parameter types

moritzkirstein:
Reference: deltaDAO/ocean.js#1

While working on bringing `ConsumerParameter`s to Ocean Market, we learned that the required types are currently missing from ocean.js.

## Proposed Changes
Add required types as documented in [Ocean Protocol DDO Docs](https://docs.oceanprotocol.com/core-concepts/did-ddo#consumer-parameters)
- Add new `ConsumerParameter` interface
- Add `consumerParameters` to the `MetadataAlgorithm` interface
- Add `consumerParameters` to the `Service` interface

## Additional Information
While publishing we noticed that current [validation schemas in aquarius](https://github.com/oceanprotocol/aquarius/blob/18fd04da1b53d149997080851dcadbd55e7420e4/aquarius/ddo_checker/shacl_schemas/v4/remote_4.5.0.ttl#L178-L231) only support `string` types for all properties other than `required`.
The proposed types of the PR are therefore considering this behaviour.

It could make sense to either
1. update aqua shapes (if behaviour from docs is preferred)
2. or note how to handle other types (e.g., we currently JSON.stringify the `default` value and optional `options` object in our implementation that's coming for market during publish / edit and parse for display in UI)


alexcos20:
lgtm, waiting for github to recover

bogdanfazakas (approved):
lgtn

PR #1747, opened 2023-07-06
Update provider nonce handling 

### LINKED ISSUE #1735: Provider nonce handling
We have mixed flow of how nonce is used when sending requests to provider.

Some functions are calling getNonce, others are just sending Date.now().

Proposal:
  - for all requests, call getNonce()
  - increase it by 1
  - use the new nonce to sign

bogdanfazakas:
Fixes #1735  .

Changes proposed in this PR:

- Updates the usage of nonce, use nonce returned by provider incremented by 1 for each provider call requiring the nonce instead of current timestamp

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,7 @@ jobs:
         run: |
           bash -x start_ocean.sh  --no-aquarius --no-elasticsearch --no-provider --no-dashboard 2>&1 > start_ocean.log &
         env:
-          CONTRACTS_VERSION: v1.1.14
+          CONTRACTS_VERSION: v1.1.8
       - run: npm ci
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -120,6 +120,7 @@ jobs:
           bash -x start_ocean.sh --with-provider2 --no-dashboard --with-c2d 2>&1 > start_ocean.log &
         env:
           CONTRACTS_VERSION: v1.1.3
+          PROVIDER_VERSION: invalid-signature
```
alexcos20, line 123: we can remove this because of https://github.com/oceanprotocol/barge/pull/372

alexcos20 (approved):
lgtm

PR #1758, opened 2023-08-01
Fix/ Download file browser

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- remove head call for download file browser helper method
- 
-


alexcos20 (approved):
lgtm

PR #1777, opened 2023-10-06
fix fs use & error handling 

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- removes fs use from config helper to prevent errors in browser context
- update provider error handling helper method


alexcos20 (approved):
lgtm

PR #1781, opened 2023-10-23
Fix/ Incrementing string nonce

### LINKED ISSUE #1780: Nonce Increment Issue
**Describe the bug**
We’re encountering a nonce problem in the code, specifically within the Ocean.js library. The issue pertains to how nonces are incremented, causing an unexpected behavior.

**To Reproduce**
Running all Backend Components on latest versions v1.1.0 (Engine, Service, Pod-Config/Publishing) v2.1.3 (Provider, Aquarius v5.1.5)

Publish a Dataset and Algorithm. Buy it twice to run it two times. In the second time you will encounter a nonce issue.

Or buy this asset (https://portal.minimal-gaia-x.eu/asset/did:op:2e148b5c1ff12590c13bc018100743f0253f22f49cbaf7876612b2417f961789) as example

**Expected behavior**
1. **Nonce Behavior:** Nonces are meant to increase by one for each new transaction, but in the code, they are gaining an extra digit.
2. **Expected Nonce:** The expected behavior is for the nonce to increment by one. For instance, from 1697790595967 to 1697790595968 for sequential transactions.
3. **Actual Nonce:** Instead of this expected behavior, the nonce is growing by an additional digit. For example, from 1697790595967 to 16977905959671.

**Screenshots**

Bildschirmfoto 2023-10-20 um 10 56 16

**Potential Cause:**
The root cause of the issue seems to be related to how the nonce is being handled in the code within the `Provider.ts` file. It appears that the nonce is treated as a string and incremented incorrectly, leading to the extra digit.

**Suggested Fix:**
To address this issue, here’s what you can do:
1. Review the code in `Provider.ts` carefully, paying close attention to any sections where nonces are manipulated.
2. Ensure that the nonce is consistently treated as an integer rather than a string.
3. Implement the necessary logic to increment the nonce by one for each transaction as expected.
4. Return from getNonce as integer. 

**Additional Information**
The nonce in Pod-Configuration is created by "const nonce = Date.now()" which was also in the an older ocean.js version. 

**Update**
Found issue in the Provider (ocean_provider/routes/consume.py).
With changing from  int(nonce) to Decimal(nonce) this now returns a string instead of an integer in the response.
New response: {"nonce":"1697785106578"} old response: {"nonce": 1697785106578}
-       assert int(nonce) == new_nonce, "New nonce could not be stored correctly."
+        assert Decimal(nonce) == Decimal(
+           new_nonce
+        ), "New nonce could not be stored correctly."

-    response = jsonify(nonce=int(nonce)), 200
+    response = jsonify(nonce=Decimal(nonce)), 200

bogdanfazakas:
Fixes #1780  .

Changes proposed in this PR:

- propper nonce incrementation

### DIFF
--- a/src/services/Provider.ts
+++ b/src/services/Provider.ts
@@ -83,7 +83,7 @@ export class Provider {
     signal?: AbortSignal,
     providerEndpoints?: any,
     serviceEndpoints?: ServiceEndpoint[]
-  ): Promise<string> {
+  ): Promise<number> {
     if (!providerEndpoints) {
       providerEndpoints = await this.getEndpoints(providerUri)
     }
@@ -102,7 +102,7 @@ export class Provider {
       })
       const { nonce } = await response.json()
       console.log(`[getNonce] Consumer: ${consumerAddress} nonce: ${nonce}`)
-      const sanitizedNonce = !nonce || nonce === null ? '0' : nonce
+      const sanitizedNonce = !nonce || nonce === null ? 0 : Number(nonce)
       return sanitizedNonce
     } catch (e) {
       LoggerInstance.error(e)


alexcos20 (approved):
lgtm

PR #1785, opened 2023-11-07
Multiple fixes

bogdanfazakas:
Fixes # .

Changes proposed in this PR:

- set higher min gas fee for sepolia test network to avoid the underpriced tx issues (faced multiple times inside the insights app)
- fixed integration tests
- upgrade node version from 16 to node 20
- update ci files to use newer versions of node
- added more details for a tx send if that tx fails


jamiehewitt15:
Is there any particular reason the tests are failing?

bogdanfazakas:
> Is there any particular reason the tests are failing?

there is a problem with the repo tests that I'm trying to fix atm, not related to the PR but would be nice to fix them before we have a new lib release

src/utils/ContractUtils.ts
```diff
@@ -149,8 +158,10 @@ export async function sendTx(
   overrides.gasLimit = estGas.add(20000)
   try {
     const trxReceipt = await functionToSend(...args, overrides)
+    const receipt = await trxReceipt.wait()
```
jamiehewitt15, line 161: `'receipt' is assigned a value but never used`

```suggestion
   await trxReceipt.wait()
```

mariacarmina (approved):

PR #1822, opened 2024-06-21
wip: update compute status/stop to support `agreementId`, not `documentId`, sig fix

### LINKED ISSUE #1820: compute Status/Stop  is using did instead of agreementId, for the C2D endpoint call
This issue is present on all components of this flow.
Starting backwards, the operator service expects ONE of:

- owner (consumer address)
- jobId (the job identifier)
- agreementId (the tx id proving the transaction)

The Provider, when calling the Operator Service, uses the `documentId` , instead of the `agreementId`.

The SDK is doing the following call to the Provider Instance
```
public async computeStatus(
    providerUri: string,
    consumerAddress: string,
    jobId?: string,
    did?: string,
    signal?: AbortSignal
  ): Promise<ComputeJob | ComputeJob[]> {
...
 let url = `?consumerAddress=${consumerAddress}`
    url += (did && `&documentId=${this.noZeroX(did)}`) || ''
    url += (jobId && `&jobId=${jobId}`) || ''


```
The fact that is doing the call to `this.noZeroX(did)` (to remove leading `0x`) clearly indicates that its not actually expecting a DID, but a **transaction hash**, since a DID would also start with **did:op:**

At the end we are sending the did as documentId parameter

Provider will then call Operator Service, which is not expecting a DID at all. On Provider side, we transform the `consumerAddress` param into `owner`, we grab the `jobId` and this `documentId` (no logic will be done with this one)

Then we call Operator Service compute status URL:
`@services.route("/compute", methods=["GET"]) def get_compute_job_status(): .... msg = f"You have to specify one of agreementId, jobId or owner"`
Here the agreementId will always be null, as we're not providing it anywhere, and the documentId is also useless to identify a job status

paulo-ocean:
Fixes #1820 .

Changes proposed in this PR:

- support for `agreementId` on get compute job status (documentId is not used, only `owner`, `jobId` or `agreementId`)
- 
-


test/integration/ComputeExamples.test.ts
```diff
@@ -661,6 +662,8 @@ describe('Compute-to-data example tests', async () => {
     /// Let's save the compute job it, we re going to use later
     /// ```Typescript
     computeJobId = computeJobs[0].jobId
+    // eslint-disable-next-line prefer-destructuring
```
jamiehewitt15, line 665: This should be written with three slashes in this file `///` - the ComputeExamples.md file is auto-generated from this file and one of the slashes gets removed in the process. 

```suggestion
    /// eslint-disable-next-line prefer-destructuring
```

jamiehewitt15:
Looks like there is an issue with the integration test

paulo-ocean:
> Looks like there is an issue with the integration test

probably ipfs related, just timeouts... nothing really changed here anyway

test/integration/PublishEditConsume.test.ts
```diff
@@ -78,7 +78,8 @@ const ifpsFile: Files = {
   files: [
     {
       type: 'ipfs',
-      hash: 'QmdMBw956S3i2H2ioS9cERrtxoLJuSsfjzCvkqoDgUa2xm'
+      hash: 'QmRhsp7eghZtW4PktPC2wAHdKoy2LiF1n6UXMKmAhqQJUA'
+      // not exists anymore? 'QmdMBw956S3i2H2ioS9cERrtxoLJuSsfjzCvkqoDgUa2xm'
```
jamiehewitt15, line 82: Is this the only test where we were using `QmdMBw956S3i2H2ioS9cERrtxoLJuSsfjzCvkqoDgUa2xm`?

jamiehewitt15 (approved):
Looks good to me

test/integration/PublishEditConsume.test.ts
paulo-ocean: i believe so. btw, i picked the new hash from the ips tests on ocean node (we had similar issues there with ipfs tests some time ago).. i might be wrong but i have the impression these ipfs files get deleted  after a while

PR #1832, opened 2024-07-01
make service endpoint / service name check case insensitive

### LINKED ISSUE #1831: fileInfo endpoint should be updated to use camelCase, and be aligned with other endpoints
Right now, this is the only endpoint service name not respecting the `camelCase`
Should be `fileInfo` instead of `fileinfo`
Corresponding piece:
```
public async getFileInfo(
    file: UrlFile | Arweave | Ipfs | GraphqlQuery | Smartcontract,
    providerUri: string,
    withChecksum: boolean = false,
    signal?: AbortSignal
  ): Promise<FileInfo[]> {
...
const path = this.getEndpointURL(serviceEndpoints, 'fileinfo')
      ? this.getEndpointURL(serviceEndpoints, 'fileinfo').urlPath
      : null
```
References:
[https://github.com/oceanprotocol/ocean-node/issues/517](https://github.com/oceanprotocol/ocean-node/issues/517)

paulo-ocean:
Fixes #1831 

Changes proposed in this PR:

- makes service endpoint check case insensitive
- we have/had issues for instance with `fileinfo` service because that endpoint was not following camelCase like all others. Just by not doing a case sensitive comparison we avoid this type of issues


jamiehewitt15 (approved):
Looks good to me

PR #1843, opened 2024-09-03
Issue 1841 pub flow - createAsset Fn

### LINKED ISSUE #1841: Add publish flow
Add `createAsset` helper function ( get inspiration [from tests](https://github.com/oceanprotocol/ocean.js/blob/c622b213ff19f091e6ce194c7dc512ee2cc40626/test/integration/helpers.ts#L21)) 
- include oasis integration, check if dt4, if so add FileObject in dt constructor
- add possibility to use template 1,2 or 4

paulo-ocean:
Fixes #1841 .

Changes proposed in this PR:

- generic function (exported) to create an asset
- update config, set confidential evm key
- calculate template index (from id or address)
- if confidential evm have filesObject on constructor (no need to provider encrypt)


src/utils/Assets.ts
```diff
@@ -0,0 +1,191 @@
...
+  const nftAddress = nftCreatedEvent.args.newTokenAddress
+  const datatokenAddressAsset = tokenCreatedEvent.args.newTokenAddress
+  // create the files encrypted string
+  assetUrl.datatokenAddress = datatokenAddressAsset
+  assetUrl.nftAddress = nftAddress
+  // TODO if template 4 no need to encrypt it??
+  if (templateIndex !== 4) {
+    ddo.services[0].files = await ProviderInstance.encrypt(assetUrl, chainID, providerUrl)
+  } else {
+    ddo.services[0].files = assetUrl
```
alexcos20, line 155: if filesObject is stored in the datatoken contract, there is no point of keeping it in ddo.  Especially unencrypted

alexcos20 (changes requested):

src/utils/Assets.ts
```diff
@@ -0,0 +1,191 @@
...
+    cap: '100000',
+    feeAmount: '0',
+    paymentCollector: account,
+    feeToken: providerFeeToken || config.oceanTokenAddress,
+    minter: account,
+    mpFeeAddress: ZERO_ADDRESS
+  }
+
+  // include fileObject in the DT constructor
+  if (templateIndex === 4) {
```
alexcos20, line 99: if network is confidential EVM.
We need to extend config with this extra key.

confidential EVM = Sapphire, Sapphire testnet.   In the future we will have more, like Ocean's L1 or OceanEnterprise chains

src/utils/Assets.ts
```diff
@@ -0,0 +1,191 @@
...
+  const nftCreatedEvent = getEventFromTx(trxReceipt, 'NFTCreated')
+  const tokenCreatedEvent = getEventFromTx(trxReceipt, 'TokenCreated')
+
+  const nftAddress = nftCreatedEvent.args.newTokenAddress
+  const datatokenAddressAsset = tokenCreatedEvent.args.newTokenAddress
+  // create the files encrypted string
+  assetUrl.datatokenAddress = datatokenAddressAsset
+  assetUrl.nftAddress = nftAddress
+  // TODO if template 4 no need to encrypt it??
+  if (templateIndex !== 4) {
```
alexcos20, line 152: if network is confidential EVM

src/utils/Assets.ts
```diff
@@ -0,0 +1,191 @@
...
+  const nftParamsAsset: NftCreateData = {
+    name,
+    symbol,
+    templateIndex,
+    tokenURI: 'aaa',
+    transferable: true,
+    owner: account
+  }
+  const datatokenParams: DatatokenCreateParams = {
+    templateIndex,
```
alexcos20, line 89: templateIndex should be calculated.

Don't go to "templateIndex == templateId", this will fail.

For instance, on sapphire we have the following config now:  (https://github.com/oceanprotocol/contracts/blob/main/addresses/address.json#L261C3-L282C5)

```
oasis_saphire_testnet": {
    "chainId": 23295,
    "Ocean": "0x973e69303259B0c2543a38665122b773D28405fB",
    "OPFCommunityFeeCollector": "0xe8c6Dc39602031A152440311e364818ba25C2Bc1",
    "startBlock": 2565487,
    "Router": "0x5FBC2A29f7C8e4533dECD14Dc1c561A8eF0d4e31",
    "FixedPrice": "0x0846aec7aADD547131484Bf7fE548589b7F057fe",
    "ERC20Template": {
      "1": "0x519d462294958A935D34351b3ad9F98cCAB0b7f8",
      "2": "0x2bCa501a6f9FB8a96f1b7B09D4f74A32ACAa7584",
      "3": "0xf00c16665b79F23c04e5c8652Fe9853046b987c8",
      "4": "0x89b752C3e272feBA17B363EaaBFe0d653cF75bE7",
      "5": "0xb4914dfeC7f70C708673620eDF6d8839F81E7612"
    },
    "ERC721Template": {
      "1": "0x9Ab15cc945b2B58A51384539211dE8a88a1355E8"
    },
    "Dispenser": "0xbde8224d15885c2E46Be665b9392E619cD142de8",
    "ERC721Factory": "0x315ef329D214cde5cE7C206104aC35B1fe66bb8c",
    "PredictoorHelper": "0x2e4cb19faD077768fDcB04Bc0549AB76BE8eA63F",
    "AccessListFactory": "0xdA412123a3Cc0D921bcf7246f8742e3b7c3E7b94"
  },
  ```
  
  Guess what?  templateIndex 4 is disabled..
  
  ```bash
  Using the following as start point:
{
  chainId: 23295,
  Ocean: '0x973e69303259B0c2543a38665122b773D28405fB',
  OPFCommunityFeeCollector: '0xe8c6Dc39602031A152440311e364818ba25C2Bc1',
  startBlock: 2565487,
  Router: '0x5FBC2A29f7C8e4533dECD14Dc1c561A8eF0d4e31',
  FixedPrice: '0x0846aec7aADD547131484Bf7fE548589b7F057fe',
  ERC20Template: {
    '1': '0x519d462294958A935D34351b3ad9F98cCAB0b7f8',
    '2': '0x2bCa501a6f9FB8a96f1b7B09D4f74A32ACAa7584',
    '3': '0xf00c16665b79F23c04e5c8652Fe9853046b987c8',
    '4': '0x89b752C3e272feBA17B363EaaBFe0d653cF75bE7',
    '5': '0xb4914dfeC7f70C708673620eDF6d8839F81E7612'
  },
  ERC721Template: { '1': '0x9Ab15cc945b2B58A51384539211dE8a88a1355E8' },
  Dispenser: '0xbde8224d15885c2E46Be665b9392E619cD142de8',
  ERC721Factory: '0x315ef329D214cde5cE7C206104aC35B1fe66bb8c',
  PredictoorHelper: '0x2e4cb19faD077768fDcB04Bc0549AB76BE8eA63F',
  AccessListFactory: '0xdA412123a3Cc0D921bcf7246f8742e3b7c3E7b94'
}
nftCount
BigNumber { value: "1" }
Router owner: 0xC7EC1970B09224B317c52d92f37F5e1E4fF6B687
currentTokenCount
BigNumber { value: "5" }


------------
templateIndex:1
[
  '0x519d462294958A935D34351b3ad9F98cCAB0b7f8',
  true,
  templateAddress: '0x519d462294958A935D34351b3ad9F98cCAB0b7f8',
  isActive: true
]
templateId: 1


------------
templateIndex:2
[
  '0x2bCa501a6f9FB8a96f1b7B09D4f74A32ACAa7584',
  true,
  templateAddress: '0x2bCa501a6f9FB8a96f1b7B09D4f74A32ACAa7584',
  isActive: true
]
templateId: 2


------------
templateIndex:3
[
  '0xf00c16665b79F23c04e5c8652Fe9853046b987c8',
  true,
  templateAddress: '0xf00c16665b79F23c04e5c8652Fe9853046b987c8',
  isActive: true
]
templateId: 3


------------
templateIndex:4
[
  '0x89b752C3e272feBA17B363EaaBFe0d653cF75bE7',
  false,
  templateAddress: '0x89b752C3e272feBA17B363EaaBFe0d653cF75bE7',
  isActive: false
]
templateId: 4


------------
templateIndex:5
[
  '0xb4914dfeC7f70C708673620eDF6d8839F81E7612',
  true,
  templateAddress: '0xb4914dfeC7f70C708673620eDF6d8839F81E7612',
  isActive: true
]
templateId: 4
```


so, the proper way of determining templateIndex is :

```js

    const currentTokenCount = await factoryERC721.connect(owner).getCurrentTemplateCount(options);
    for(i=1;i<=currentTokenCount;i++){
        const tokenTemplate = await factoryERC721.connect(owner).getTokenTemplate(i,options);
        console.log("\n\n------------\ntemplateIndex:"+i)
        console.log(tokenTemplate)
        const erc20Template = await ethers.getContractAt("ERC20Template", tokenTemplate["templateAddress"]);
        const id=await erc20Template.connect(owner).getId()
        console.log("templateId: "+id)
        if(tokenTemplate.isActive && id==SEARCHED_TEMPLATE_ID || tokenTemplate["templateAddress"]==SEARCHED_TEMPLATE_ADDRESS){
                               return i
        }
    }
    return null     
    ```

File: src/config/Config.ts [hunk 5]
```diff
@@ -44,6 +44,11 @@ export class Config {
    */
   public nftFactoryAddress?: string
 
+  /**
+   * ERC20 templates addresses
+   */
+  public eRC20Template?: any
```
alexcos20, line 50: why do we need this?

src/config/ConfigHelper.ts
```diff
@@ -168,6 +180,17 @@ export const configHelperNetworks: Config[] = [
   }
 ]
 
+export const KNOWN_CONFIDENTIAL_EVMS = {
```
alexcos20, line 183: why do we need this?
We should always compare or act based on chainId and not on network name

src/config/ConfigHelper.ts
```diff
@@ -179,6 +202,7 @@ export class ConfigHelper {
         FixedPrice,
         Dispenser,
         ERC721Factory,
+        ERC20Template, // added to Config
```
alexcos20, line 205: why do we need this? We can fetch the list from ERC721Factory

src/config/ConfigHelper.ts
```diff
@@ -194,6 +218,7 @@ export class ConfigHelper {
       } = customAddresses[network]
       configAddresses = {
         nftFactoryAddress: ERC721Factory,
+        eRC20Template: ERC20Template, // added to Config
```
alexcos20, line 221: why do we need this? We can fetch the list from ERC721Factory

src/utils/Assets.ts
```diff
@@ -0,0 +1,332 @@
...
+}
+
+/**
+ * Use this function if don't need to check if the template if active
+ * (not 100% reliable if we need to check at smart contract level)
+ * @param chainID the chain identifier
+ * @param template the id or the template address
+ * @returns the index of the template from the 'ERC20Template' object
+ */
+export async function getTemplateIndexOnList(
```
alexcos20, line 81: why do we need this function?

src/utils/Assets.ts
```diff
@@ -0,0 +1,332 @@
...
+  name: string,
+  symbol: string,
+  owner: Signer,
+  assetUrl: any,
+  template: string | number, // If string, it's template address , otherwise, it's templateId
+  ddo: any,
+  encryptDDO: boolean = true, // default is true
+  providerUrl: string,
+  providerFeeToken: string,
+  nftContractAddress: string, // addresses.ERC721Factory,
```
alexcos20, line 181: change to nftFactory.
but why? We can get nftFactory from addresses(env or contracts npm package), so we don't need it as param

src/utils/Assets.ts
```diff
@@ -0,0 +1,332 @@
...
+  const nftAddress = nftCreatedEvent.args.newTokenAddress
+  const datatokenAddressAsset = tokenCreatedEvent.args.newTokenAddress
+  // create the files encrypted string
+  assetUrl.datatokenAddress = datatokenAddressAsset
+  assetUrl.nftAddress = nftAddress
+  // if template 4 no need to encrypt it??
+  if (!config.confidentialEVM) {
+    ddo.services[0].files = await ProviderInstance.encrypt(assetUrl, chainID, providerUrl)
+  }
+
```
alexcos20, line 298:
```suggestion
else{
    ddo.services[0].files=null
}
```

alexcos20: if it's confidential, that we don't store the unencrypted URL in the ddo, cause anyone can grab it

test/integration/PublishFlows.test.ts
```diff
@@ -198,6 +199,112 @@ describe('Publish tests', async () => {
...
 
+  it('should publish a dataset with fixed price (create NFT + Datoken + fixed price) using createAsset() fn', async () => {
+    const fixedPriceDdo: DDO = { ...genericAsset }
+    const ownerAddress = publisherAccount
+    const asset = await createAsset(
+      'test asset',
+      'TEST',
+      ownerAddress,
+      assetUrl,
+      '0x13a9813e7CE3fdc867C23Ef9c084570B7F2abdeC', // template 1 on dev network
```
alexcos20, line 210: I would a templateID here.
If you want to test with addresses, then read addresses.json and use one of the values.
But never hardcode it

src/utils/Assets.ts
```diff
@@ -0,0 +1,332 @@
...
+ * @param fixedRateAddress fixed rate exchange address
+ * @param baseTokenAddress base token address (ocean)
+ * @returns ddo id as string
+ */
+export async function createAsset(
+  name: string,
+  symbol: string,
+  owner: Signer,
+  assetUrl: any,
+  template: string | number, // If string, it's template address , otherwise, it's templateId
```
alexcos20, line 176: rename to  templateIDorAddress pls, so everybody knows that they need to pass either an templateID (and not template Index), or templateContractAddress

test/unit/AssetUtils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+
+    // optimism sepolia
+    // 11155420
+    assert(
+      isConfidentialEVM(11155420) === false,
+      `Chain Id: "11155420" is wrongly considered a confidental EVM`
+    )
+  })
+
+  it('should get correct template index from contract artifacts (using SC address as template)', async () => {
```
alexcos20, line 44: Why do we need this?
Never assume any hardcoded address.
Maybe next week we will change the templates, cause we have found a bug.  Or deploy a new product..
Then all ocean.js tests will fail, although they shoudn't

test/unit/AssetUtils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    const wrongOne = await getTemplateIndexOnList(
+      KNOWN_CONFIDENTIAL_EVMS.networks[1].chainId, // testnet chain
+      '12'
+    )
+    assert(wrongOne === -1, 'wrong template index, should be inexistent!')
+
+    const okIndex = await getTemplateIndexOnList(
+      KNOWN_CONFIDENTIAL_EVMS.networks[0].chainId, // mainnet chain
+      '0x4dD281EB67DED07E76E413Df16176D66ae69e240'
```
alexcos20, line 53: see above

test/unit/AssetUtils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    )
+    assert(okIndex >= 1, 'wrong template index, should exist!')
+
+    const okIndexNonConfidential = await getTemplateIndexOnList(
+      11155111, // sepolia
+      '0xDEfD0018969cd2d4E648209F876ADe184815f038'
```
alexcos20, line 59: see above

test/unit/AssetUtils.test.ts
```diff
@@ -0,0 +1,112 @@
...
+    assert(okIndexNonConfidential === 2, 'Should be template 2 for sepolia!')
+
+    const notOkIndexNonConfidential = await getTemplateIndexOnList(
+      11155111, // sepolia
+      '0xDEfD0018969cd2d4E648209F876ADe184815f022' // wrong template
+    )
+    assert(notOkIndexNonConfidential === -1, 'Template should not exist on sepolia!')
+  })
+
+  it('should get correct template index from contract artifacts (using template ID as template)', async () => {
```
alexcos20, line 70: see above

alexcos20 (changes requested):

src/config/ConfigHelper.ts
paulo-ocean: `getConfig`
`public getConfig(network: string | number, infuraProjectId?: string): Config {`
...
 uses either network name or chain id... If we supply and search for a "name"  we might not able to check if it is a confidential EVM... specially because we have wrong names on the contracts address.json  ('saphire' misspelled with only 1 'p'). Was trying to overcome this obstacle and make it more flexible, but of course i can put only chain ids on the constant

src/config/ConfigHelper.ts
paulo-ocean: not needed, probably was thinking about something else at the time

paulo-ocean:
btw, the service "files" cannot be null for confidential EVM, but empty string instead... 
if its null the DDO does not passes against schema validation
schema:ServiceShape
```
....
sh:targetClass schema:Service ;
    sh:property [
        sh:path schema:id ;
        sh:datatype xsd:string ;
        sh:pattern "^(.*)$" ;
        sh:minCount 1;
        sh:maxCount 1;
    ] ;
...
```

alexcos20:
> btw, the service "files" cannot be null for confidential EVM, but empty string instead... if its null the DDO does not passes against schema validation schema:ServiceShape
> 
> ```
> ....
> sh:targetClass schema:Service ;
>     sh:property [
>         sh:path schema:id ;
>         sh:datatype xsd:string ;
>         sh:pattern "^(.*)$" ;
>         sh:minCount 1;
>         sh:maxCount 1;
>     ] ;
> ...
> ```

then let's go with empty string

test/integration/PublishFlows.test.ts
```diff
@@ -198,6 +199,111 @@ describe('Publish tests', async () => {
...
+      fixedPriceDdo,
+      true, // encrypted ddo
+      providerUrl,
+      ZERO_ADDRESS, // provider fee token
+      aquarius
+    )
+
+    console.log('Published asset, ddo id:', asset)
+
+    // const nftParams: NftCreateData = {
```
mariacarmina, line 220: shall we use this code after merging this branch with main?

mariacarmina (commented):
Hello @paulo-ocean can you also merge the main branch into this PR please?

test/integration/PublishFlows.test.ts
paulo-ocean: not the console.log :-) I've cleaned the test. If you mean the new function createAsset, yes, its an option, or we can do it all in many steps as we we're doing before (like some of the other tests)

src/utils/Assets.ts
```diff
@@ -0,0 +1,350 @@
...
+    cap: '100000',
+    feeAmount: '0',
+    paymentCollector: account,
+    feeToken: providerFeeToken || config.oceanTokenAddress,
+    minter: account,
+    mpFeeAddress: ZERO_ADDRESS
+  }
+
+  // include fileObject in the DT constructor
+  if (config.confidentialEVM) {
```
mariacarmina, line 199: **Observation**: If datatoken 4 has the file object and we want to create a bundled tx (meaning deploy NFT, datatoken 4 and pricing schema at once) and files object needs to be encrypted on chain, then we should also have a test when we check, if confidentialEVM flag is set on true we need to `sapphire.wrap()` the owner wallet address, otherwise the tx won't be encrypted on sapphire chain.
**Question**: When creating a datatoken from a standalone NFT, NFT class should also have this confidentialEVM flag when creating a datatoken template 4, right?

mariacarmina (commented):
I have an observation and a question regarding datatoken 4 creation.

src/utils/Assets.ts
mariacarmina: for question, I have created https://github.com/oceanprotocol/ocean.js/pull/1859 in case.

paulo-ocean: `sapphire.wrap()` is a specific API for Oasis. At this point its true that its the only confidential EVM we have, but that could be different tomorrow, we **cannot**  do shapire.wrap() on a TX just because its "confidentialEVM". It only works for Oasis shappire. In any case i think that applies only for `calldata` when interacting with the SC, as the smart contract should run on their own secure enclave... not really sure if that's on the scope of this function here. Maybe @alexcos20  can have some input here.. its easy, but in any case we need to check the chain, its not enough see if its confidential or not.

Regarding your question, i'm not sure if i understand your point.. When we have a confidential EVM, the requirement would be pass the filesObject on the DatatokenCreateParams

paulo-ocean: > for question, I have created #1859 in case.

i see, but you're again doing a fixed match:  confidentialEVM === Shappire 
afaik, that might not always be the case

mariacarmina: I agree @paulo-ocean to have a stronger condition and include chain id as well. I wanted to raise an awareness that when creating a datatoken (either bundled or standalone) to encrypt the tx since we pass the files object, even though files object is already encoded in utf8 bytes, that was related to my question. I agree to also merge your PR and do this `sapphire.wrap()` thing separate in my PR if needed when creating a datatoken.

mariacarmina: updated my PR with the stronger condition and tests as well.

mariacarmina (approved):
I can approve, the only pending point that we discovered on this PR is tackled on https://github.com/oceanprotocol/ocean.js/pull/1859

jamiehewitt15 (approved):
Looks good to me

paulo-ocean:
merging, any open point we can address on a separate PR, as we need to update cli and the flow was already tested anyway

PR #1845, opened 2024-09-04
Integrate contracts v2.2.0

mariacarmina:
Fixes # .

Changes proposed in this PR:

- integrated factory contract
- updated contract version
- created test script for sapphire


src/@types/NFTFactory.ts
```diff
@@ -21,3 +21,12 @@ export interface NftCreateData {
   transferable: boolean
   owner: string
 }
+
+export interface AccessListData {
+  name: string
+  symbol: string
+  tokenURI: string[]
```
alexcos20, line 28: why do we need this?

src/contracts/AccessList.ts
```diff
@@ -41,6 +39,33 @@ export class AccessListContract extends SmartContract {
...
+   * Get Symbol of Access list
+   * @return {Promise<string>} Symbol
+   */
+  public async getSymbol(accessListAddress: string): Promise<string> {
+    const accessListContract = this.getContract(accessListAddress)
+    return await accessListContract.symbol()
+  }
+
   /**
    * Mint ERC721 contract
```
alexcos20, line 70: mint new token, or add address to accesslist

it's not minting another erc721 contract

src/contracts/AccessList.ts
```diff
@@ -99,6 +124,13 @@ export class AccessListContract extends SmartContract {
     return <ReceiptOrEstimate<G>>trxReceipt
   }
 
+  /**
+   * Burn Access List
```
alexcos20, line 128: delete address from accessList    or  burn tokenId

src/contracts/AccessListFactory.ts
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  /**
+   * Create new Access List Contract
+   * @param {AccessListData} listData The data needed to create an NFT.
+   * @param {Boolean} [estimateGas] if True, return gas estimate
+   * @return {Promise<string|BigNumber>} The transaction hash or the gas estimate.
+   */
+  public async deployAccessListContract<G extends boolean = false>(
+    listData: AccessListData,
```
alexcos20, line 42: I would advice you to delete that typing (AccessListData) and use function params

src/contracts/NFT.ts
```diff
@@ -70,22 +74,32 @@ export class Nft extends SmartContract {
     )
     if (estimateGas) return <G extends false ? string : BigNumber>estGas
 
+    const addresses = [minter, paymentCollector, mpFeeAddress, feeToken]
+    if (accessListContract) {
```
alexcos20, line 78: only templateID == 4 supports accessLists , so pls make sure that we don't pass it to other template, because those templates might use those fields for something else

test/integration/PublishEditConsume.test.ts
```diff
@@ -349,49 +349,49 @@ describe('Publish consume test', async () => {
...
-      resolvedGraphqlAssetDdo.id,
-      resolvedGraphqlAssetDdo.services[0].datatokenAddress,
-      await consumerAccount.getAddress(),
-      resolvedGraphqlAssetDdo.services[0].id,
-      0,
-      datatoken,
-      providerUrl
-    )
-    assert(grapqlOrderTx, 'Ordering graphql dataset failed.')
+    // arwaveOrderTx = await orderAsset(
```
alexcos20, line 352: Out of curiosity, why did we removed this tests?

test/config.ts
```diff
@@ -48,3 +48,16 @@ export const getAddresses = () => {
   )
   return data.development
 }
+
+export const getAddressesForSapphire = (testnet: boolean) => {
```
alexcos20, line 52: Maybe it's a good ideea to have two functions:
 - getAddressesForChain(chainID)
 - getAddressesForSapphire which calls getAddressesForChain

test/integration/PublishEditConsume.test.ts
mariacarmina: I have commented them for debug because these were failing.

src/contracts/AccessListFactory.ts
mariacarmina: ok sure.

src/contracts/AccessList.ts
mariacarmina: I'll modify

src/@types/NFTFactory.ts
mariacarmina: it's a typing as `NftCreateData` is used when creating a NFT.

test/config.ts
mariacarmina: I have that locally, not pushed yet.

test/integration/PublishEditConsume.test.ts
alexcos20: well, you commented the order , but left the downloads on next tests :) and those are failing without orders..

mariacarmina: I can focus on this problem, now since I am blocked on testing the integration side.

alexcos20:
why for every call to accesslist functions do i need to provide accessListAddress ?

just have it in the constructor and do something like:
```
accessList = new AccessListContract('0x123')
owner = await accessList.getOwner()
```

mariacarmina:
> why for every call to accesslist functions do i need to provide accessListAddress ?
> 
> just have it in the constructor and do something like:
> 
> ```
> accessList = new AccessListContract('0x123')
> owner = await accessList.getOwner()
> ```

From my point of view to pass the address for every call since we have methods related to allow list and deny list, because right now they are all together. If you want, we can have 2 children classes, one for allow list and one for deny list which can store the address in the constructor.

mariacarmina:
> why for every call to accesslist functions do i need to provide accessListAddress ?
> 
> just have it in the constructor and do something like:
> 
> ```
> accessList = new AccessListContract('0x123')
> owner = await accessList.getOwner()
> ```

It wasn't a specific preference, I can change that AccessList class to extend SmartContractWithAddress class and store the address in constructor. The only reason for using this approach is that Nft class has in each method nft address passed as parameter.

jamiehewitt15:
All the integration tests passed on the most recent run. The error message reports a naming conflict in the upload of `coverage` - perhaps try just changing the name to `integration-coverage`?

jamiehewitt15:
The issue with the tests, is likely because `upload-artifact@v2` is deprecated. Can you update it to `upload-artifact@v4` in `.github/workflows/ci.yml`

.github/workflows/ci.yml
```diff
@@ -143,7 +143,7 @@ jobs:
         run: docker logs ocean_aquarius_1 && docker logs ocean_provider_1 && docker logs ocean_provider2_1 && docker logs ocean_computetodata_1
         if: ${{ failure() }}
       - name: Upload coverage
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         with:
           name: coverage
```
jamiehewitt15, line 148: We're uploading three artifacts all with the same name and you're getting a name conflict error. 

I suggest changing this one:

```suggestion
          name: coverage-integration
```



And change the one for the unit test to `coverage-unit`

jamiehewitt15:
You're getting `Error: Timeout of 40000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.` in the integration tests. Have you already tried rerunning the tests to see if it's a one time issue?

Also, you can try increasing the timeout to see if this fixes the issue.

jamiehewitt15:
Everything is passing now, I just restarted the tests and it works.

mariacarmina:
> You're getting `Error: Timeout of 40000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.` in the integration tests. Have you already tried rerunning the tests to see if it's a one time issue?
> 
> Also, you can try increasing the timeout to see if this fixes the issue.

tried that previously, increased to 60000 ms, it was related with coverage reports.

mariacarmina:
Screenshot 2024-09-15 at 17 32 08

mariacarmina:
set file object with encrypted format (the last test): https://explorer.oasis.io/testnet/sapphire/tx/0x2b8ce371f987a3e0309555973b677a75a3307fbe13855c4e43ae23bc666311ce

mariacarmina:
The integration tests for ordering different types of assets is an older issue which pops up due to exceeded timeout. I commented the assets which rely on external services such as ipfs, arweave and create an issue for fixing those tests in the future since they are not impacted by this PR changes.

jamiehewitt15 (commented):
Why are the sapphire integration tests in a folder called `scripts`? Seems like it would be better placed in `/test/integration/`

.gitignore
```diff
@@ -3,7 +3,8 @@ dist/
 .nyc_output/
 coverage/
 doc/
-test/**/*.js
+test/unit/*.js
+test/integration/*.js
```
jamiehewitt15, line 7: Why have you done this? Before we ignored all javascript files in the test directory, now you want only to ignore js files if they are in either `/test/unit` or `test/integration`? I don't see any benefit to changing this

README.md
```diff
@@ -128,6 +128,21 @@ npm run test:integration
 npm run test:integration:cover
 ```
 
+### Sapphire Integration Tests
+
+Currently, there is used Oasis Sapphire Test network, please export the `PRIVATE_KEY` before testing.
```
jamiehewitt15, line 133: This is worded poorly, I think what you are trying to say is: 

```suggestion
We are currently using the live Oasis Sapphire Test network for the integration tests. Please export the `PRIVATE_KEY` before running the tests.
```

test/scripts/SapphireIntegration.test.ts
```diff
@@ -0,0 +1,260 @@
...
+import { Datatoken4 } from '../../src/contracts/Datatoken4'
+import { AbiItem, Config, Nft, NftCreateData } from '../../src'
+import ERC20Template4 from '@oceanprotocol/contracts/artifacts/contracts/templates/ERC20Template4.sol/ERC20Template4.json'
+import { getEventFromTx } from '../../src/utils'
+
+describe('Sapphire tests', async () => {
+  const provider = sapphire.wrap(
+    ethers.getDefaultProvider(sapphire.NETWORKS.testnet.defaultGateway)
+  )
+  const wallet = new ethers.Wallet(process.env.PRIVATE_KEY, provider)
```
jamiehewitt15, line 18: Given that these tests are using a private key on a live network, at some point, the wallet address is going to run out of tokens and the tests will fail. Is there some sort fo longer-term plan here, like getting the tests to work with barge? If so, do we have an issue for this?

package.json
```diff
@@ -30,6 +30,7 @@
     "mocha": "TS_NODE_PROJECT='./test/tsconfig.json' mocha --config=test/.mocharc.json --node-env=test --exit",
     "test": "npm run lint && npm run test:unit:cover && npm run test:integration:cover",
     "test:unit": "npm run mocha -- 'test/unit/**/*.test.ts'",
+    "test:sapphire": "npm run mocha -- 'test/scripts/**/*.test.ts'",
```
bogdanfazakas, line 33: lets move the test file to its proper location /test/integration/
also renaming it to Sapphire.test.ts would be nice
then update the test:saphire to  "npm run mocha -- 'test/integration/Sapphire.test.ts'" .. and then just ignore the newloy moved file in the test:integration comand

src/contracts/AccessList.ts
```diff
@@ -15,106 +14,189 @@ export class AccessListContract extends SmartContract {
...
+    const balance = await this.contract.balanceOf(address)
+    return await this.unitsToAmount(null, balance, 18)
+  }
+
+  /**
+   * Add address to access list
    * @param {String} user Minter address
    * @param {String} tokenUri tokenURI
    * @param {Boolean} estimateGas if True, return gas estimate
    * @return {Promise<ReceiptOrEstimate>} transactionId
```
bogdanfazakas, line 90: returns the transaction receipt

src/contracts/AccessList.ts
```diff
@@ -15,106 +14,189 @@ export class AccessListContract extends SmartContract {
...
       tokenUris
     )
     return <ReceiptOrEstimate<G>>trxReceipt
   }
 
+  /**
+   * Delete address from access list
+   * @param {Number} tokenId token ID
+   * @param {Boolean} estimateGas if True, return gas estimate
+   * @return {Promise<ReceiptOrEstimate>} transactionId
```
bogdanfazakas, line 141:
```suggestion
   * @return {Promise<ReceiptOrEstimate>} returns the transaction receipt or the estimateGas value
```

src/contracts/AccessList.ts
```diff
@@ -15,106 +14,189 @@ export class AccessListContract extends SmartContract {
...
   }
 
   /**
-   * Batch Mint ERC721 contract
-   * @param {String} accessListAddress AccessList contract address
+   * Batch add addresses to the access list
    * @param {String} users Minter addresses
    * @param {String} tokenUris tokenURI
    * @param {Boolean} estimateGas if True, return gas estimate
    * @return {Promise<ReceiptOrEstimate>} transactionId
```
bogdanfazakas, line 116:
```suggestion
   * @return {Promise<ReceiptOrEstimate>} returns the transaction receipt or the estimateGas value
```

src/contracts/AccessList.ts
```diff
@@ -15,106 +14,189 @@ export class AccessListContract extends SmartContract {
...
       tokenId
     )
     return <ReceiptOrEstimate<G>>trxReceipt
   }
+
+  /**
+   * Transfer Ownership of an access list, called by owner of access list
+   * @param {Number} newOwner new owner of the access list
+   * @param {Boolean} estimateGas if True, return gas estimate
+   * @return {Promise<ReceiptOrEstimate>} transactionId
```
bogdanfazakas, line 164:
```suggestion
   * @return {Promise<ReceiptOrEstimate>}  returns the transaction receipt or the estimateGas value
```

src/contracts/AccessListFactory.ts
```diff
@@ -0,0 +1,144 @@
...
+      this.contract.deployAccessListContract,
+      nameAccessList,
+      symbolAccessList,
+      transferable,
+      owner,
+      user,
+      tokenURI
+    )
+    const trxReceipt = await tx.wait()
+    const events = getEventFromTx(trxReceipt, 'NewAccessList')
```
bogdanfazakas, line 78: what if the tx fails, need to provide a proper response for the client otherwise the line bellow  events.args[0] will fail with uncaught error

src/utils/Asset.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Get the artifacts address from the address.json file
+ * either from the env or from the ocean-contracts dir
+ * @returns data or null
+ */
+export function getOceanArtifactsAdresses(): any {
+  try {
+    if (process.env.ADDRESS_FILE) {
+      // eslint-disable-next-line security/detect-non-literal-fs-filename
+      const data = fs.readFileSync(process.env.ADDRESS_FILE, 'utf8')
```
bogdanfazakas, line 19: i would also add a default here if the address file is not present and remove the condition above 

```suggestion
    const data = JSON.parse(
    // eslint-disable-next-line security/detect-non-literal-fs-filename
    fs.readFileSync(
      process.env.ADDRESS_FILE ||
        `${homedir}/.ocean/ocean-contracts/artifacts/address.json`,
      'utf8'
    )
  )
```

src/config/ConfigHelper.ts
```diff
@@ -155,6 +155,16 @@ export const configHelperNetworks: Config[] = [
     subgraphUri: 'https://v4.subgraph.sapphire-mainnet.oceanprotocol.com/',
     explorerUri: 'https://explorer.oasis.io/mainnet/sapphire/',
     gasFeeMultiplier: 1
+  },
+  {
+    ...configHelperNetworksBase,
+    chainId: 23295,
+    network: 'oasis_saphire_testnet',
```
paulo-ocean, line 162: this is misspelled again, chain name is: `oasis_sapphire_testnet`

src/utils/Asset.ts
paulo-ocean: can't see anything wrong here, what are you exactly suggesting @bogdanfazakas ?
this is perfectly fine
```
export function getOceanArtifactsAdresses(): any {
  try {
    if (process.env.ADDRESS_FILE) {
      // eslint-disable-next-line security/detect-non-literal-fs-filename
      const data = fs.readFileSync(process.env.ADDRESS_FILE, 'utf8')
      return JSON.parse(data)
    }
    return addrs
  } catch (error) {
    return addrs
  }
}
```
The `process.env.ADDRESS_FILE` should be the file path
if not defined or wrongly defined it returns the defaults from address.json, you're like repeating the same thing twice

.gitignore
mariacarmina: It was when I created a .js file, I'll correct it now since I have the .ts script, not .js

test/scripts/SapphireIntegration.test.ts
mariacarmina: That is the reason that I did not include the tests in the CI, because everytime the account will be drained, we'll have to top up the funds for those. For the moment it's not tested with CI, only locally. We can create an issue for it. @mihaisc suggested to have a separate job which not runs at every push but he said to leave this problem for now, I'll create an issue for it.

package.json
mariacarmina: Ok, sure I can do that.

src/contracts/AccessListFactory.ts
mariacarmina: Yes, the same should be applied in NFTFactory when creating a NFT, try-catch block is missing from there too.

src/contracts/Datatoken4.ts
```diff
@@ -131,30 +137,62 @@ export class Datatoken4 extends Datatoken {
...
+   * @param {String} providerAddress
+   * @param {Bytes} providerSignature
+   * @param {Bytes} consumerData
+   * @param {Bytes} consumerSignature
+   * @param {String} consumerAddress
+   * @return {Promise<Bytes>} returns file object
+   */
+  public async getFileObject(
+    dtAddress: string,
+    serviceId: number,
```
paulo-ocean, line 180: We need to refactor all these occurrences of `serviceID` (including comments).. It definitely is NOT the service id, but the service index.. so `serviceIndex` instead, otherwise we get easily confused

mariacarmina: right thing, sure, do that now.

src/contracts/AccessListFactory.ts
```diff
@@ -0,0 +1,148 @@
...
+    config?: Config,
+    abi?: AbiItem[]
+  ) {
+    super(address, signer, network, config, abi)
+    this.abi = abi || this.getDefaultAbi()
+  }
+
+  /**
+   * Create new Access List Contract
+   * @param {AccessListData} listData The data needed to create an NFT.
```
paulo-ocean, line 37: lets kill this comment about AccessListData :-)

test/scripts/SapphireIntegration.test.ts
jamiehewitt15: ok cool, fair enough

src/utils/Asset.ts
bogdanfazakas: yep makes sense then, we can close this then

src/contracts/Datatoken4.ts
```diff
@@ -30,19 +30,23 @@ export class Datatoken4 extends Datatoken {
     abi?: AbiItem[]
   ) {
     super(signer, network, config, abi)
+    this.abi = this.getDefaultAbi()
     // Wrap signer's address for encrypted data tx
     this.signer = sapphire.wrap(signer)
```
paulo-ocean, line 35: is the template only for confidential EVM sapphire?
Cause we're wrapping the signer with their sdk call `sapphire.wrap(...)` ..
and we have a few functions doing it again like:
`confidentialEVM === true ? sapphire.wrap(this.signer) : this.signer, ...`
So, we get a "double" wrap() for confidential EVM and a wrap() for any other EVM using this template 
For now, i would just remove the wrap() call on the constructor, and later, once we merge the other PR we can refactor this a bit more (from the config we would know if is confidential evm and if we should wrap the calls

mariacarmina: I can remove the wrapper from constructor to avoid confusions.

mariacarmina: Template 4 is only for confidential EVM, I mean, the scope of its creation is destinated for sapphire and files encryption on sapphire chain.

src/contracts/AccessListFactory.ts
mariacarmina: created https://github.com/oceanprotocol/ocean.js/pull/1853

bogdanfazakas: yes but the logic here still need few more changes, what if the tx fails, there would be no events matching the name we pass, so for at line 89 you will end up with an "cannot read property x of undefined"

mariacarmina: I put a check only for `tx` returned by `sendTx`, because if that is not present the code below will fail.

bogdanfazakas: i dont't really get it maybe i'm missing something, so if sendTx fails it will only log an error cause here we don't have a try catch or something handling that, so it go on wi calling `tx.wait()`, then trying to access events for that tx etc

jamiehewitt15 (approved):
Looks good to me

paulo-ocean (approved):
lgtm

mariacarmina:
@alexcos20 can you check this PR please?

src/contracts/AccessListFactory.ts
mariacarmina: if `sendTx` will fail, either will throw an error that will be caught in the try-catch block within `sendTx` function or the try-catch block from `deployAccessListContract` function which is surrounded by, either the tx receipt is null. If tx returned from `sendTx` does not exist, this will throw an error with message: `Tx for deploying new access list was not processed on chain.`. It will not go to `tx.wait()` please check my latest code where I have added the condition, because it is above `tx.wait()`.

alexcos20 (approved):

PR #1851, opened 2024-09-19
Fix ipfs CID.

### LINKED ISSUE #1849: Fix ipfs test
**Describe the bug**
Fix ipfs test by replacing the hash for ipfs asset on PublishEditConsume.test.ts file
**Expected behavior**
Pass the test on CI.

mariacarmina:
Fixes #1849  .

Changes proposed in this PR:

- 
- 
-


test/integration/PublishEditConsume.test.ts
```diff
@@ -146,6 +136,31 @@ function delay(interval: number) {
   }).timeout(interval + 100)
 }
 
+function uploadToIpfs(data: any): Promise<string> {
+  return new Promise((resolve, reject) => {
+    axios
+      .post(
+        'http://172.15.0.16:5001/api/v0/add',
```
jamiehewitt15, line 143: Looks like your still using a local IPFS node? I thought you were going to use our cluster?

mariacarmina: I used in the previous commit

mariacarmina: https://github.com/oceanprotocol/ocean.js/commit/676c328af44e1f617556aef0355bd2cf92c43854

jamiehewitt15: I was suggesting that we use the IPFS cluser that we're hosting

Same as what we're doing in the ocean-node tests:
https://github.com/oceanprotocol/ocean-node/blob/c5799f9d187851922a9e80a090e0e19325a06aa9/src/test/unit/storage.test.ts#L518 

There is also an IPFS CID there we're using for the get storage info test that is working fine:  `QmRhsp7eghZtW4PktPC2wAHdKoy2LiF1n6UXMKmAhqQJUA`

jamiehewitt15: The `QmRhsp7eghZtW4PktPC2wAHdKoy2LiF1n6UXMKmAhqQJUA` is working fine in the ocean-node tests. So I'd suggest we just try and use that here as well. So we skip the uploading step

jamiehewitt15: I also just checked and that asset is downloadable https://ipfs.oceanprotocol.com/ipfs/QmRhsp7eghZtW4PktPC2wAHdKoy2LiF1n6UXMKmAhqQJUA

jamiehewitt15: It is worth a try but it may not work because the provider in barge is using the local IPFS node: https://github.com/oceanprotocol/barge/blob/3abd5effa33e39d739dd51f0b3f1249475a4429c/start_ocean.sh#L115

It can take a few minutes for a new IPFS node to connect to enough peers to discover the asset, leading the test to timeout.

mariacarmina: yes, it did throw a timeout error

.github/workflows/ci.yml
```diff
@@ -140,7 +140,7 @@ jobs:
         env:
           PROVIDER_URL: 'http://172.15.0.4:8030'
       - name: docker logs
-        run: docker logs ocean_aquarius_1 && docker logs ocean_provider_1 && docker logs ocean_provider2_1 && docker logs ocean_computetodata_1
+        run: docker logs ocean-ipfs-1
```
bogdanfazakas, line 143: i think we need to revert this or maybe just add ipfs logs if we need them

src/services/Provider.ts
```diff
@@ -372,7 +372,7 @@ export class Provider {
     let initializeUrl = this.getEndpointURL(serviceEndpoints, 'initialize')
       ? this.getEndpointURL(serviceEndpoints, 'initialize').urlPath
       : null
-
+    console.log(initializeUrl)
```
bogdanfazakas, line 375: let remove this console logs

src/services/Provider.ts
```diff
@@ -389,12 +389,13 @@ export class Provider {
         headers: { 'Content-Type': 'application/json' },
         signal
       })
+      console.log(response)
```
bogdanfazakas, line 392: same here

package.json
```diff
@@ -54,6 +54,7 @@
     "web3": "^1.8.0"
   },
   "dependencies": {
+    "axios": "^1.7.7",
```
bogdanfazakas, line 57: do we really need axios, we already fetch from  'cross-fetch' you can use that instead of axios so we don't end up having 2 deps with same usage

test/integration/PublishEditConsume.test.ts
```diff
@@ -146,6 +136,31 @@ function delay(interval: number) {
   }).timeout(interval + 100)
 }
 
+function uploadToIpfs(data: any): Promise<string> {
+  return new Promise((resolve, reject) => {
+    axios
```
bogdanfazakas, line 141: let's use fetch() instead of axios

test/integration/helpers.ts
```diff
@@ -174,6 +174,7 @@ export async function orderAsset(
   datatoken: Datatoken,
   providerUrl: string
 ) {
+  console.log(did)
```
bogdanfazakas, line 177: delete this pls

test/integration/PublishEditConsume.test.ts
```diff
@@ -146,6 +136,31 @@ function delay(interval: number) {
...
+          '\r\n--------------------------a28d68b1c872c96f--\r\n',
+        {
+          headers: {
+            'Content-Type':
+              'multipart/form-data; boundary=------------------------a28d68b1c872c96f'
+          }
+        }
+      )
+      .then(function (response: any) {
+        console.log('hash: ', response.data.Hash)
```
bogdanfazakas, line 155: same let's delete this

.github/workflows/ci.yml
mariacarmina: I'll revert how it was before

bogdanfazakas (approved):
lgtm

jamiehewitt15 (approved):
Looks good to me

PR #1859, opened 2024-10-03
Add confidential EVM flag when creating a datatoken 4 in NFT class

mariacarmina:
Fixes # .

Changes proposed in this PR:

- Wrap signer for every contract function


src/contracts/NFTFactory.ts
```diff
@@ -70,7 +74,10 @@ export class NftFactory extends SmartContractWithAddress {
     try {
       const tx = await sendTx(
         estGas,
-        this.signer,
+        this.config.confidentialEVM === true &&
```
paulo-ocean, line 77: Are we sure we have the proper` this.config`  here? (Same for all other similar classes, where this change was made) . I think the `Config` object is only available if we passed it on the constructor, or if we provided a network on the constructor.
This is what we have on the `SmartContract` constructor:
```
constructor(
    signer: Signer,
    network?: string | number,
    config?: Config,
    abi?: AbiItem[]
  ) {
    this.signer = signer
    this.config = config || new ConfigHelper().getConfig(network || 'unknown')
    this.abi = abi || this.getDefaultAbi()
  }
```
If we don't supply a `Config` object or a valid `network` it executes the the default:
`config = new ConfigHelper().getConfig('unknown')`
This will return a `null` config

mariacarmina: I agree, the easiest way is to make network mandatory and to have a check for supported networks.
What do you think @paulo-ocean?

mariacarmina: Actually, signer has a method called "signer.getChainId()", we can use that and have a check for supported networks.

mariacarmina: It will generate breaking change within SDK contracts classes and it implies a lor of modifications around all the classes making network a mandatory parameter. I already modified some tests to provide the network as well, but I think here comes the user responsability to set a config object with the necessary parameters. What I can improve to the check is to validate if config object has confidential EVM key or not, otherwise, use the regular signer.

src/contracts/AccessList.ts
```diff
@@ -122,10 +124,11 @@ export class AccessListContract extends SmartContractWithAddress {
   ): Promise<ReceiptOrEstimate<G>> {
     const estGas = await this.contract.estimateGas.batchMint(users, tokenUris)
     if (estimateGas) return <ReceiptOrEstimate<G>>estGas
-
     const trxReceipt = await sendTx(
       estGas,
-      this.signer,
+      this.config && 'sdk' in this.config && this.config.sdk === 'oasis'
```
paulo-ocean, line 129: I would create a function for this check... we're repeating the same lines of code multiple times. Also, since we don't have a constants or a enum type... its better to not write the same strings in multiple places, its very error prone

src/config/ConfigHelper.ts
```diff
@@ -315,13 +330,8 @@ export class ConfigHelper {
         addresses
       )
     }
-    config.confidentialEVM =
```
paulo-ocean, line 318: since you removed the check here... what about  all the addresses/chains coming from this line?:
`let contractAddressesConfig = this.getAddressesFromEnv(config.network, addresses)`
which value will these have for `sdk` field?  I think **None** :-)
only the hardcoded ones on `configHelperNetworks` will have it

src/contracts/AccessList.ts
mariacarmina: I agree that we need to remove the code duplication from the contract level and put it into sendTx. What is unclear for me is that how we can manage to retrieve the config if it was already defined by the user.
If the user already defined it's own config object, we have to retrieve it without creating a new one. Should we have a check in the 'args' or set an optional parameter for config within `sendTx`? I am saying even I will create a separate function, we still need to verify if there is an existing config object already set.

paulo-ocean: > I agree that we need to remove the code duplication from the contract level and put it into sendTx. What is unclear for me is that how we can manage to retrieve the config if it was already defined by the user. If the user already defined it's own config object, we have to retrieve it without creating a new one. Should we have a check in the 'args' or set an optional parameter for config within `sendTx`? I am saying even I will create a separate function, we still need to verify if there is an existing config object already set.

Not sure what's the problem, if the config was passed as argument to the smart contract class?? .
The Config is a collection of 'network' configurations... and for each network there is only one entry, so if you do a ` new ConfigHelper().getConfig(network)` i don't see how that could give a different entry from the "original one" (if any)..
Afaik, the "args" are the params for the smart function to be called... putting the config there not only would force the user to pass it, but it would be even more messy, since that is not a SC function argument... Thats really confusing...

paulo-ocean: there are possibilities as well. You could have a method on abstract `SmartContract` class, like
getWrappedSigner() or whatever
where you could put the repeated code:
```
this.config && 'sdk' in this.config && this.config.sdk === 'oasis'
        ? sapphire.wrap(this.signer)
        : this.signer,
```
that returns the Signer (wrapped or not)

Then inside each class where `sendTx() `is beeing called you could just put `this.getWrappedSigner()` instead of those lines.. In this case we would not touch `sendTx` ... it would still be 93 changes but much cleaner

mariacarmina: I prefer the second method it will be much cleaner since we'll call a function, not having duplicated lines of code.

src/config/ConfigHelper.ts
mariacarmina: I updated the config object within `getConfig`, contractAddressesConfig contains only contract addresses, so I set the sdk instead inside config object, not the addresses. Let me know if I missed anything from your suggestion.

> the main issue is not addressed... this.config will be null everywhere, as we are no passing the Config to any of the classes, so i don't see how this line can ever evaluate to 'true'

For the config, it shouldn't be null, now since SmartContract class handles the creation in the constructor.

paulo-ocean (approved):
lgtm,
but i think we have to check why the tests are not running

jamiehewitt15:
> we have to check why the tests are not running

Yeah, what's the deal with the tests?

PR #1863, opened 2024-10-21
Bump version of eslint to 9.9.1.

mariacarmina:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30,7 +30,7 @@
         "chai": "^4.3.6",
         "chai-spies": "^1.0.0",
         "cross-env": "^7.0.3",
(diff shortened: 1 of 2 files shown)


paulo-ocean:
seems the tests are failing exactly because of this dependency.
btw, shouldn't it be 9.9.1 ?
Screenshot from 2024-11-07 09-18-19

jamiehewitt15:
> shouldn't it be 9.9.1 ?

The latest version is `9.14.0`.

> some stuff can simply be incompatible

Yeah true. This is a dev dependency after all so we shouldn't spend too much time on this.

PR #1868, opened 2024-10-29
Add checks for ddo chain id and did. 

mariacarmina:
Fixes # .

Changes proposed in this PR:

- Add checks for ddo chain id and did. 
- Added check for confidential evm.

### DIFF
--- a/CodeExamples.md
+++ b/CodeExamples.md
@@ -167,7 +167,7 @@ Next, we define the metadata that will describe our data asset. This is what we 
     '@context': ['https://w3id.org/did/v1'],
     id: '',
     version: '4.1.0',
-    chainId: 4,
+    chainId: 8996,
     nftAddress: '0x0',
     metadata: {
       created: '2021-12-20T14:35:20Z',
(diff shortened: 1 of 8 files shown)


jamiehewitt15:
Can you check what thee issue with the integration test is?

mariacarmina:
> Can you check what the issue with the integration test is?

It seems that the modifications work fine :)) so the tests failed as expected, I'll modify the chain id from the samples DDOs to match barge test chain id.

src/config/ConfigHelper.ts
```diff
@@ -339,7 +339,9 @@ export class ConfigHelper {
       }
     }
 
-    config.accessListFactory = contractAddressesConfig.accessListFactory
```
paulo-ocean, line 342: Why this change? It makes no difference, except that now we have more lines of code :-)

mariacarmina: I saw some errors when testing vscode extension and importing the latest version of the lib, that accessListFactory key is undefined from contractAddressesConfig, because it was mandatory and did not have the check before.

paulo-ocean: but `accessListFactory` is not mandatory, this is how it is defined on Config class:
`accessListFactory?: string`
And the changes does nothing

paulo-ocean (approved):
lgtm

PR #1877, opened 2024-11-20
chore: signDDO

AdriGeorge:
Changes proposed in this PR:

- Sign credential with waltId
- Sign credential with private key


paulo-ocean (commented):
how do we know if the functions like `signCredential` and `signCredentialWithWaltId` actually do what's expected? I don't see any usage of these, no unit tests, etc? And the` 'privateKey'` is supposed to be sent clear text? (i don't think we have any similar case)

package.json
```diff
@@ -56,10 +56,12 @@
   "dependencies": {
     "@oasisprotocol/sapphire-paratime": "^1.3.2",
     "@oceanprotocol/contracts": "^2.2.0",
+    "axios": "^1.7.7",
```
jamiehewitt15, line 59: Why have you added axios when we already have cross-fetch?

PR #1889, opened 2024-12-02
Add back check for access list factory.

mariacarmina:
Fixes # .

Changes proposed in this PR:

From this discussion https://github.com/oceanprotocol/ocean.js/pull/1868#discussion_r1824334821, I reopened the topic, that assignement does not check if  the accessListFactory exists and throws this error for not EVM chains (for example in barge):
Error publishing asset: Cannot read properties of undefined (reading 'accessListFactory')

### DIFF
--- a/src/config/ConfigHelper.ts
+++ b/src/config/ConfigHelper.ts
@@ -338,7 +338,7 @@ export class ConfigHelper {
         config.sdk = 'evm'
       }
     }
-    if ('accessListFactory' in contractAddressesConfig && config.sdk === 'oasis') {
+    if ('accessListFactory' in contractAddressesConfig) {
       config.accessListFactory = contractAddressesConfig.accessListFactory
     }
 
--- a/src/config/ConfigHelper.ts
+++ b/src/config/ConfigHelper.ts
@@ -338,7 +338,7 @@ export class ConfigHelper {
         config.sdk = 'evm'
       }
     }
-    if ('accessListFactory' in contractAddressesConfig) {
+    if (contractAddressesConfig && 'accessListFactory' in contractAddressesConfig) {
       config.accessListFactory = contractAddressesConfig.accessListFactory
     }
 
--- a/src/config/ConfigHelper.ts
+++ b/src/config/ConfigHelper.ts
@@ -338,8 +338,9 @@ export class ConfigHelper {
         config.sdk = 'evm'
       }
     }
-
-    config.accessListFactory = contractAddressesConfig.accessListFactory
+    if ('accessListFactory' in contractAddressesConfig && config.sdk === 'oasis') {
+      config.accessListFactory = contractAddressesConfig.accessListFactory
+    }
 
     config = { ...config, ...contractAddressesConfig }
 


src/config/ConfigHelper.ts
```diff
@@ -338,8 +338,9 @@ export class ConfigHelper {
         config.sdk = 'evm'
       }
     }
-
-    config.accessListFactory = contractAddressesConfig.accessListFactory
+    if ('accessListFactory' in contractAddressesConfig && config.sdk === 'oasis') {
```
paulo-ocean, line 341: is the `accessListFactory`  ONLY related with the oasis network?
I don't think so

mariacarmina: > is the accessListFactory ONLY related with the oasis network?
I don't think so. Its related with template 4, not a specific chain (even if for the moment is only oasis using it)
What are we exactly trying to achieve here?

It's related with same networks, for example on polygon you don't have accesslist factory. We are trying to achieve here is not to enforce accessListFactory and to check if it exists. Maybe the check with config.sdk shouldn't be there, because you can have accessListFactory on other networks

paulo-ocean: Not sure if i made myself clear, but the current change does not solve the issue, its wrong
we should check if `contractAddressesConfig` exists..
and only then do this:
`config.accessListFactory = contractAddressesConfig.accessListFactory` (this is OK)
This error:
`Error publishing asset: Cannot read properties of undefined (reading 'accessListFactory')`
means `contractAddressesConfig` is `undefined` and we're trying to access the property `accessListFactory` from it 

the change should be something like:
```
if(contractAddressesConfig) {
  config.accessListFactory = contractAddressesConfig.accessListFactory
}
```

jamiehewitt15 (dismissed):
Looks good to me

paulo-ocean:
this is definitely NOT OK

This:
```
if ('accessListFactory' in contractAddressesConfig) {
      config.accessListFactory = contractAddressesConfig.accessListFactory
}
 
```
Will break if `contractAddressesConfig` is `null` or `undefined`
we cannot search for a property in something that does not exist, fyi:
[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/in_operator_no_object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/in_operator_no_object)

You can simply, and safely, do as i suggested above
```
if(contractAddressesConfig) {
  config.accessListFactory = contractAddressesConfig.accessListFactory
}
```

mariacarmina:
> you really like to complicate @mariacarmina , not sure why :-)
> 
> This:
> 
> ```
> if (contractAddressesConfig && 'accessListFactory' in contractAddressesConfig) {
>       config.accessListFactory = contractAddressesConfig.accessListFactory
>  }
> ```
> 
> Has exactly the same effect as my suggested change (bellow), except that you made you more complicate (2 checks instead of 1). Can't really understand why but anyway
> 
> ```
> if(contractAddressesConfig) {
>   config.accessListFactory = contractAddressesConfig.accessListFactory
> }
> ```

I do not like to complicate @paulo-ocean :) but as I mentioned previously, the assignment is related to 'accessListFactory' -> so it is good practice to check as well if 'accessListFactory' is present in contractAddressesConfig, because not all the chains have accessListFactory key, because they are not confidential EVM. For me your check is incomplete, we surely need to check if contractAddressesConfig exists, but also if accessListFactory is present because that's what we want to assign. Please check https://github.com/oceanprotocol/contracts/blob/main/addresses/address.json and search 'AccessListFactory' and you'll see that only for oasis that key exists.
I need to modify the key name from contractAddressesConfig, it is not 'accessListFactory', it's 'AccessListFactory'.

paulo-ocean:
> > you really like to complicate @mariacarmina , not sure why :-)
> > This:
> > ```
> > if (contractAddressesConfig && 'accessListFactory' in contractAddressesConfig) {
> >       config.accessListFactory = contractAddressesConfig.accessListFactory
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Has exactly the same effect as my suggested change (bellow), except that you made you more complicate (2 checks instead of 1). Can't really understand why but anyway
> > ```
> > if(contractAddressesConfig) {
> >   config.accessListFactory = contractAddressesConfig.accessListFactory
> > }
> > ```
> 
> I do not like to complicate @paulo-ocean :) but as I mentioned previously, the assignment is related to 'accessListFactory' -> so it is good practice to check as well if 'accessListFactory' is present in contractAddressesConfig, because not all the chains have accessListFactory key, because they are not confidential EVM. For me your check is incomplete, we surely need to check if contractAddressesConfig exists, but also if accessListFactory is present because that's what we want to assign. Please check https://github.com/oceanprotocol/contracts/blob/main/addresses/address.json and search 'AccessListFactory' and you'll see that only for oasis that key exists.

its exactly the same @mariacarmina ... but keep your ideas, i know nothing, and should not be trying to help

PR #1895, opened 2024-12-13
make retries and interval bigger & configurable on waitForAqua

### LINKED ISSUE #1894: Allow bigger/configurable interval on waitForAqua() (when checking if a DDO is already indexed)
Right now, the function `waitForAqua` which is used whenever we want to check the status of  a DDO (for instance on the CLI `getDDO` command). It does not have a configurable interval to do the checks..
It keeps retrying itself every 1.5 seconds, for a MAX of up to 100 retries... (all hardcoded)
This is kind of pointless if the DDO is taking some time to get indexed..
We should at least double the default interval (so, retry every 3 seconds) and make it a configurable argument for the function. Additionally, the MAX retries could also be configurable
This is also helpful to lower the number of inconsequent requests possible made against the node (which also can post rate/connection limits)

paulo-ocean:
Fixes #1894 

Changes proposed in this PR:

- Make number of retries and interval configurable. increase interval default
- 
-


paulo-ocean:
> rename to waitForIndexer or something else, we don't have Aqua anymore :D

yeah, good point, thanks :+1:

bogdanfazakas (approved):
LGTM

PR #1905, opened 2025-01-29
Use serviceIndex instead of hardcoded indexes.

### LINKED ISSUE #1803: Ocean.js should support assets with multiple services & files
Looks like it's not complete yet: https://github.com/oceanprotocol/ocean.js/blob/7c4da83ee3f8c0ff8fa417b7c41e3ae72924e960/src/utils/OrderUtils.ts#L105

mariacarmina:
Fixes #1803  .

Changes proposed in this PR:

- Use serviceIndex instead of hardcoded indexes.
- fix workflow

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,12 +119,12 @@ jobs:
       - run: npm run build:metadata
       - name: Delete default runner images
         run: |
-          docker image rm node:18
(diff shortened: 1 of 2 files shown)


mariacarmina:
For OrderUtils, by default the serviceIndex is set to 0, but it can be modified by the interface which uses this lib.

PR #1914, opened 2025-02-19
add validate function signature changes, missing update docs and others

### LINKED ISSUE #1913: Update Aquarius validate DDO call (missing some optional parameters)
This is related with [https://github.com/oceanprotocol/ocean-node/pull/835](https://github.com/oceanprotocol/ocean-node/pull/835)
validateDDO command
expects `publisherAddress`  `nonce`  and `signature`
Without these fields (and some more checks) the node does not add the signature on the response anymore

paulo-ocean:
Fixes #1913 

Changes proposed in this PR:

- updates Aquarius "`validate`" function (for validateDDO).. Ocean node needs '`nonce`', '`signature`' and '`publisherAddress`' to validate ... signer and provider uri as extra params
- still use legacy validate (no signatures), as backup/retry mechanism just in case provider does not support the newest changes (ex: old provider)
- refactor legacy helpers/`createAsset` function name to helpers/`createAssetHelper` (to avoid confusion with  newer fn `utils/createAsset`)


jamiehewitt15 (approved):
Looks good to me

alexcos20:
I've changed the base of this PR to be release_4_0, so watch out for changes.
Especially everything DDO related, which must use ddo.js

PR #1932, opened 2025-03-25
Make integration tests use Ocean Node not Provider or Aquarius

### LINKED ISSUE #1930: Update ocean js docs for free compute release
check docs and complete if needed.

bogdanfazakas:
Fixes #1930  .

Changes proposed in this PR:

- make integration tests use ocean node instead of old components
- update config object to contain ocean node uri and remove provider, metadatacache and subgraph references from config
- remove not supported storage types
- keep only free c2d tests we'll add paid compute jobs later on
- update docs
- update ci to add node logs if integration tests fail
- fix IPFS upload and addded back the ipfs publish, edit and consume tests


test/integration/CodeExamples.test.ts
```diff
@@ -487,9 +489,13 @@ describe('Marketplace flow tests', async () => {
       consumerAccount,
       await consumerAccount.getChainId()
     )
+
+    await new Promise((resolve) => setTimeout(resolve, 5000))
```
giurgiur99, line 493: Do we need this timeout here?

test/integration/ComputeFlow.test.ts
```diff
@@ -240,10 +240,12 @@ async function waitTillJobEnds(): Promise<number> {
       const jobStatus = (await ProviderInstance.computeStatus(
         providerUrl,
         await consumerAccount.getAddress(),
-        freeComputeJobId,
-        resolvedDdoWith5mTimeout.id
+        freeComputeJobId
       )) as ComputeJob
-      if (jobStatus?.status === 70) {
+      console.log('Job status: ', jobStatus)
```
giurgiur99, line 245: Maybe we don't need this logs here

test/integration/ComputeFlow.test.ts
```diff
@@ -415,9 +417,11 @@ describe('Compute flow tests', async () => {
     computeValidUntil = Math.floor(mytime.getTime() / 1000)
 
     // we choose the free env
-    const computeEnv = computeEnvs[resolvedDdoWith5mTimeout.chainId].find(
-      (ce) => ce.priceMin === 0 || isDefined(ce.free)
+    const computeEnv = computeEnvs.find(
+      (ce) =>
+        !ce?.fees || ce.fees.find((fee) => fee.symbol === 'OCEAN' && fee.amount === '0')
     )
+    console.log('Free compute environment = ', computeEnv)
```
giurgiur99, line 424: Here the logs can be removed

giurgiur99 (approved):
LGTM

CodeExamples.md
```diff
@@ -578,10 +597,12 @@ Lets check that the download URL was successfully received
       console.log(fileData)
     } catch (e) {
       LoggerInstance.error('Download failed', e)
-      assert.fail('Download failed')
+      assert.fail('Download failed') ///
```
mariacarmina, line 600: Is this required as a marker for code examples?

CodeExamples.md
```diff
@@ -578,10 +597,12 @@ Lets check that the download URL was successfully received
...
     }
-  
 ```
+<!--
```
mariacarmina, line 603: Can we remove those timeout comments?

ComputeExamples.md
```diff
@@ -585,115 +594,137 @@ Now, let's check that we successfully published a algorithm (create NFT + Datato
...
     )
+-->
 
 let's check the free compute environment
 ```Typescript
-    const computeEnv = computeEnvs[resolvedDatasetDdo.chainId].find(
-      (ce) => ce.priceMin === 0 || isDefined(ce.free)
+    const computeEnv = computeEnvs.find(
+      (ce) =>
+        !ce?.fees || ce.fees.find((fee) => fee.symbol === 'OCEAN' && fee.amount === '0')
```
mariacarmina, line 614: Should we still keep the OCEAN symbol for compute env fee? Besides OCEAN token, do we have other compute env with fee amount 0?

test/integration/PublishEditConsume.test.ts
```diff
@@ -402,31 +405,32 @@ describe('Publish consume test', async () => {
   // )
   // assert(ipfsOrderTx, 'Ordering ipfs dataset failed.')
```
mariacarmina, line 406: I remember we had a discussion regarding uncommenting the IPFS tests which were failing before, did it occur again the behaviour? What error did we encounter on this test?
I think we can uncomment since the test above publishes the IPFS asset

mariacarmina (commented):
I left some questions, let's align on those, otherwise this PR is in a good shape right now.

ComputeExamples.md
bogdanfazakas: i would say keep it , just in case someone is using old components to run tests on, and those return fees in ocean token

test/integration/CodeExamples.test.ts
bogdanfazakas: Nope will remove those

test/integration/ComputeFlow.test.ts
bogdanfazakas: good catch, was used for debug purposes 👍

test/integration/ComputeFlow.test.ts
bogdanfazakas: good catch, was used for debug purposes 👍

CodeExamples.md
bogdanfazakas: no i've tried to remove that from the md file, but we need a different approach, will update it

CodeExamples.md
bogdanfazakas: nope we need to keep them

mariacarmina:
The problem with ipfs test might be caused by the fact that `ddo.json` file does not exist in order to be uploaded, resulting to `null` CID from ipfs is returned which trigger in the node this error for missing CID for IPFS file

bogdanfazakas:
> The problem with ipfs test might be caused by the fact that `ddo.json` file does not exist in order to be uploaded, resulting to `null` CID from ipfs is returned which trigger in the node this error for missing CID for IPFS file. Let's try storing locally this file in a separate folder https://raw.githubusercontent.com/oceanprotocol/provider/4cd244bc18565c0d7bf631f97b584cbd199f0d03/tests/resources/ddo_sample_file.txt and rename the file from POST request to `ddo.txt` for example.

fixed this, was an issue with the actual upload to ipfs, now consume works fine with ipfs files as well

mariacarmina (approved):
Nice job! Let's merge!

PR #1938, opened 2025-04-28
Dependencies update

giurgiur99:
Fixes # .

Changes proposed in this PR:

- Latest ddo-js


package.json
```diff
@@ -98,10 +88,8 @@
     "release-it": "^18.1.2",
     "source-map-support": "^0.5.21",
     "ts-node": "^10.9.2",
-    "ts-node-register": "^1.0.0",
     "tsx": "^4.19.2",
     "typedoc": "^0.25.1",
-    "typedoc-plugin-markdown": "^3.15.2",
```
mariacarmina, line 104: Let's keep this, since it is used in the docs generation script https://github.com/oceanprotocol/ocean.js/blob/d2d54a5d5077c986a2051016ea94bfd96e81660c/package.json#L27

package.json
```diff
@@ -70,25 +67,18 @@
...
     "auto-changelog": "^2.4.0",
-    "c8": "^10.1.3",
     "chai": "^4.3.6",
-    "chai-spies": "^1.0.0",
-    "cross-env": "^7.0.3",
     "eslint": "^8.23.1",
     "eslint-config-oceanprotocol": "^2.0.3",
     "eslint-config-prettier": "^9.0.0",
     "eslint-plugin-prettier": "^4.2.1",
-    "fs": "0.0.1-security",
```
mariacarmina, line 91: Please let's keep this dependency for the moment, it appears in multiple places. It should use the node's default fs package, but this is not in the scope now

mariacarmina (commented):
Some observations regarding the dependencies from my side

package.json
giurgiur99: For future references https://www.npmjs.com/package/fs

mariacarmina (approved):
LGTM!

bogdanfazakas (approved):

PR #1949, opened 2025-05-27
Fix bugs and log fixes

giurgiur99:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/services/Aquarius.ts
+++ b/src/services/Aquarius.ts
@@ -140,8 +140,8 @@ export class Aquarius {
     if (signer) {
       try {
         // make it optional and get from env if not present
-        if (process.env.OCEAN_NODE_URL) {
-          providerUrl = process.env.OCEAN_NODE_URL
+        if (process.env.NODE_URL) {
+          providerUrl = process.env.NODE_URL
         } else {
           providerUrl = process.env.PROVIDER_URL
         }
@@ -167,7 +167,7 @@ export class Aquarius {
         // let signatureMessage = publisherAddress
         const signatureMessage = ddo.id + nextNonce
         const signature = await signRequest(signer, signatureMessage)
-        const data = { ddo, publisherAddress, nonce: nextNonce, signature }
+        const data = { ...ddo, publisherAddress, nonce: nextNonce, signature }
         response = await fetch(path, {
           method: 'POST',
           body: JSON.stringify(data),
--- a/src/config/ConfigHelper.ts
+++ b/src/config/ConfigHelper.ts
@@ -175,9 +175,9 @@ export class ConfigHelper {
     let configAddresses: Partial<Config>
 
     const getUris = () => {
-      if (process.env.OCEAN_NODE_URL) {
+      if (process.env.NODE_URL) {
         return {
-          oceanNodeUri: process.env.OCEAN_NODE_URL
+          oceanNodeUri: process.env.NODE_URL
         }
       }
       return {}
@@ -287,9 +287,9 @@ export class ConfigHelper {
     try {
       addresses = process.env.ADDRESS_FILE
         ? JSON.parse(
-            // eslint-disable-next-line security/detect-non-literal-fs-filename
-            fs.readFileSync(process.env.ADDRESS_FILE, 'utf8')
-          )
+          // eslint-disable-next-line security/detect-non-literal-fs-filename
+          fs.readFileSync(process.env.ADDRESS_FILE, 'utf8')
+        )
         : null
     } catch (e) {
       console.log(e)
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -114,94 +123,41 @@ jobs:
...
         working-directory: ${{ github.workspace }}/barge
         run: |
-          export CONTRACTS_VERSION=v2.3.0
-          bash -x start_ocean.sh --no-aquarius --no-provider --no-dashboard --with-typesense 2>&1 > start_ocean.log &
-      - run: npm ci
-      - run: npm run build
-      - run: docker image ls
-      - name: Delete default runner images
-        run: |
-          docker image rm -f node:20
```
mariacarmina, line 124: Can we keep those docker image removals in the workflow if they exist? The runner pull those images by default and we do not need them.

.github/workflows/ci.yml
```diff
@@ -114,94 +123,41 @@ jobs:
...
+              docker inspect ocean-node-1
+              exit 1
+            fi
             sleep 10
-            [ -f "$HOME/.ocean/ocean-contracts/artifacts/ready" ] && break
+            attempt=$((attempt + 1))
           done
-          cat $ADDRESS_FILE
-
+      - run: npm ci
```
mariacarmina, line 147: Can we havethose commands with npm under one `run` command? Thanks!

.github/workflows/ci.yml
```diff
@@ -114,94 +123,41 @@ jobs:
...
+      - run: npm run build
+      
       - name: docker logs
         run: docker logs ocean_ocean-contracts_1 && docker logs ocean_kindcluster_1 && docker logs ocean_computetodata_1 && docker logs ocean_typesense_1
```
mariacarmina, line 151: These should be updated and monitor the actual containers, meaning we can safely remove `docker logs ocean_kindcluster_1, docker logs ocean_computetodata_1` and add ocean node container instead.

.github/workflows/ci.yml
```diff
@@ -114,94 +123,41 @@ jobs:
...
-          exit 1
-
       - name: integration
         run: npm run test:integration:cover
         env:
           INDEXING_RETRY_INTERVAL: 4000
           INDEXING_MAX_RETRIES: 120
-          OCEAN_NODE_URL: 'http://127.0.0.1:8001'
+          NODE_URL: 'http://127.0.0.1:8001'
+          PROVIDER_URL: 'http://172.15.0.3:8545'
```
mariacarmina, line 160: As far as I know, the only place where PROVIDER_URL was used is in Aquarius as fallback if OCEAN_NODE_URL was not present, I do not fully underdstand the purpose of PROVIDER_URL, we are using for RPC storage or backwards compatible with legacy Provider? If it is used for RPC storage, you can rename it to `RPC` and use it in the `development` config https://github.com/oceanprotocol/ocean.js/pull/1949/files#diff-13d78f1779ca97ed6a5284a28b60c1ea808ebd2f5792d0a306c16c8b3c463a35R10. Else, if it is for backwards compatible with legacy Provider, then the CI workflow needs to start a new barge with legacy components and the host for it will be `http://172.15.0.4:8030`.
If we do not have tests on legacy components, I suggest to remove from flow and document it in the README as all the envs should be documented :)

src/services/Aquarius.ts
```diff
@@ -108,111 +108,68 @@ export class Aquarius {
...
+          validatorAddress: jsonResponse.publicKey,
+          r: jsonResponse.r[0],
+          s: jsonResponse.s[0],
+          v: jsonResponse.v
+        }
+      }
+    } catch (e) {
+      // retry with legacy path validation
+      LoggerInstance.error(
+        'Metadata validation failed using publisher signature validation (perhaps not supported or legacy Aquarius), retrying with legacy path...',
```
mariacarmina, line 169: There is no retry for legacy Aquarius, if we want backwards compatibility, then we should retry the legacy path, otherwise only the log should be updated.

mariacarmina (commented):
I left some comments about PROVIDER_URL and retry legacy paths for Aquarius. In rest it looks aligned with what has been discussed.

Thanks!

mariacarmina (approved):
LGTM! Let's merge this

.github/workflows/ci.yml
```diff
@@ -55,20 +49,33 @@ jobs:
...
+            if [ $attempt -eq $max_attempts ]; then
+              echo "Error: Node container failed to start within 300 seconds"
+              docker ps -a
+              docker inspect ocean-node-1
+              exit 1
+            fi
+            sleep 10
+            attempt=$((attempt + 1))
+          done
+      - run: npm ci
```
mariacarmina, line 78: Can we have those `run` actions under one statement please? Thank you!

mariacarmina (commented):
One small suggestion from my side

PR #1957, opened 2025-06-16
optional checks

giurgiur99:
Changes proposed in this PR:

- Fix: Start compute fails if no dataset provided
-  Fix: Start compute fails if no rpc initialized (pass chainId instead of getting it from rpc)


alexcos20 (approved):
lgtm

PR #1982, opened 2025-08-05
fix hexlify numbers

giurgiur99:
Changes proposed in this PR:

- Replace oceanNodeUri with the local node, do no rely on external node
- Hexlify is no longer used to convert numbers to hexstrings, should be replace by 'toBeHex' in ethers.js

```
// v5; converting numbers to hexstrings
hex = hexlify(35)

// v6; converting numbers to hexstrings
hex = toBeHex(35)
```
Reference: https://docs.ethers.org/v6/migrating/


mariacarmina (approved):
LGTM!

PR #2055, opened 2026-03-23
p2p impl

giurgiur99:
Fixes #2053 

Changes proposed in this PR:

- Split Provider.ts into BaseProvider (routing), HttpProvider, P2pProvider — auto-routes by URI format
- Run CI tests with HTTP & P2P
- Full P2P transport over libp2p: KadDHT, bootstrap, mDNS, TLS, retry on dropped connections
- enableTcp flag in P2PConfig for Node.js (off by default for browsers)
- getDownloadUrl() streams binary via lpStream → DownloadResponse
- getComputeResult() new streaming method → AsyncGenerator<Uint8Array> with resumable offset
- computeStatus() takes signerOrAuthToken consistently across all layers
- computeStart payment body includes resources in both transports
- Aquarius.resolve/waitForIndexer/validate route through P2P when URL is a multiaddr
- NFT.setMetadata stores bare peer ID on-chain (strips multiaddr prefix)


alexcos20 (commented):

Overall risk: medium

Summary:
This pull request introduces P2P communication capabilities to the `ocean.js` library by refactoring the `Provider` service. The original `Provider` logic is now encapsulated within an `HttpProvider`, and a new `P2pProvider` handles `libp2p` based communication. A `BaseProvider` acts as a facade, routing requests to the appropriate underlying provider based on the `nodeUri`. The CI pipeline has been updated to test both HTTP and P2P integration paths. Key changes include adding `libp2p` dependencies, updating Node.js versions in CI, and modifying `NFT.ts` and `Aquarius.ts` to leverage the new P2P routing for metadata decryption and DDO operations. While the architectural change is positive, there are some areas for careful review, particularly around P2P stream handling and return type consistency.

Comments:
• [INFO][other] Node.js version updated from `20` to `22` across multiple jobs. While Node.js 22 is generally backward-compatible, it's good practice to monitor for any unexpected regressions in the CI or runtime with this new version.
• [WARNING][bug] The regex `/p2p/([^/]+)/` is used to extract the peer ID from `metadataDecryptorUrl`. This regex assumes a very specific multiaddr format. If multiaddrs can have different structures (e.g., `/p2p/Qm.../tcp/4001`), this regex might not be robust enough. Consider using a dedicated `multiaddr` parsing library function for more reliable extraction, or refine the regex to account for more variations.
• [INFO][style] The refactoring of the `Provider` class into `BaseProvider`, `HttpProvider`, and `P2pProvider` using a facade pattern is a good architectural improvement. This will make the codebase more modular and extensible for future transport mechanisms.
• [WARNING][other] The `isP2pUri` function relies on `startsWith('/')`, `startsWith('12D3')`, `startsWith('16Uiu')`, or `startsWith('Qm')`. While pragmatic for common peer IDs/multiaddrs, this could be brittle if new peer ID prefixes emerge or if multiaddr formats vary in unexpected ways. A more explicit check for `/p2p/` within the string might be more robust, potentially leveraging `multiaddr` library's parsing capabilities.
• [INFO][other] The `bufToHex` utility function handles various buffer representations (JSON string, object, indexed object). Given its use for fields like `hash`, `publicKey`, `r`, `s` in `validateDdo`, which are often hex strings, ensure the `Buffer.from(data).toString()` conversion correctly yields hex (e.g., if `data` bytes are actual ASCII hex characters, not raw bytes of the hex value). If the data is truly raw bytes that *represent* a hex string, `toString('hex')` might be more appropriate. Clarification on the exact input/output intent would be beneficial.
• [WARNING][other] The `libp2pNode` and `lastBootstrapKey` are global variables. This means there will only be one `libp2p` instance for the entire application, which could lead to issues in scenarios requiring multiple distinct P2P connections or different configurations, or in a testing environment where state needs to be reset between tests. Consider managing the `libp2pNode` lifecycle per `P2pProvider` instance or within a more controlled singleton pattern if global state is strictly intended.
• [ERROR][bug] In `dialAndStream`, the stream is closed immediately after writing the payload (`await stream.close()`). For a typical request/response pattern over `lpStream`, the client would write the request, then *read the response* from the same stream, and only then close it. Closing the stream immediately might prevent the client from receiving the full response or cause errors on the server side if it's still trying to write to a closed stream. This pattern needs to be carefully verified with the expected `libp2p` stream usage and the `lpStream` library's behavior. If `lpStream` detaches read/write, it might be fine, but it seems suspicious.
• [ERROR][bug] The `getComputeResultUrl` method in `P2pProvider` returns `Promise<any>` which resolves to an `AsyncGenerator<Uint8Array>`, while the `HttpProvider` counterpart returns `Promise<string>` (a URL). The `BaseProvider` facade's signature for this method is `Promise<string>`, creating a type mismatch at compile time for `P2pProvider` and a potential runtime error for callers expecting a `string` (URL) when interacting with a P2P `nodeUri`. The return type of `BaseProvider.getComputeResultUrl` needs to be updated to reflect both possibilities (e.g., `Promise<string | AsyncGenerator<Uint8Array>>`), and all call sites in the library need to be updated to handle the `AsyncGenerator` type when a P2P URI is used.
• [INFO][style] Removed `console.log` statements for nonce, which is good cleanup for production code.

alexcos20 (commented):

Overall risk: medium

Summary:
This pull request introduces P2P communication support to `ocean.js`, refactoring the core `Provider` logic into a `BaseProvider`, `HttpProvider`, and `P2pProvider`. It updates the CI to Node.js 22, adds necessary `libp2p` dependencies, and modifies `Aquarius` and `NFT` services to leverage the new P2P capabilities. Integration tests are updated to cover both HTTP and P2P transports. The architectural changes are a significant improvement for supporting multiple communication protocols.

Comments:
• [INFO][other] The CI workflow is updated to Node.js 22 across multiple jobs. This is a good move to stay current, but ensure all project dependencies have been thoroughly tested for full compatibility with Node.js 22.
• [WARNING][other] The `test_integration` job for the `p2p` transport matrix entry explicitly runs `npm run test:integration` instead of `npm run test:integration:cover`. Additionally, the 'Upload coverage' step is conditional (`if: matrix.transport == 'http'`). This means P2P integration tests are not contributing to code coverage. Please clarify if this is a temporary measure or a known limitation, and if there are plans to include P2P test coverage in the future to ensure comprehensive testing.
• [INFO][style] The change from `import ... assert { type: 'json' }` to `import ... with { type: 'json' }` is correct for Node.js 22's adoption of 'import attributes' over 'import assertions'. This aligns with the Node.js version update in CI.
• [INFO][other] The logic to extract the peer ID from `metadataDecryptorUrl` when it's a P2P multiaddr is a necessary and well-commented change to ensure correct routing for P2P decrypt operations within the indexer. Good job anticipating the indexer's requirements.
• [INFO][architecture] The refactoring of the large `Provider` class into `BaseProvider`, `HttpProvider`, and `P2pProvider` is an excellent architectural improvement. It significantly enhances modularity, maintainability, and extensibility, making it easier to manage different communication protocols. This is a strong positive for the codebase.
• [INFO][other] The `Aquarius` service now correctly routes DDO resolution and validation requests to the appropriate provider (HTTP or P2P) using `isP2pUri` and `ProviderInstance`. This is a critical integration point for the new P2P functionality.
• [WARNING][other] The `isP2pUri` function's check for multiaddrs starting with `Qm` might be overly broad. While `Qm` often indicates a CID (which can be part of a P2P multiaddr), it's not exclusively for multiaddrs and could potentially lead to false positives. Consider if a more precise check (e.g., validating the full multiaddr structure or looking for `/p2p/` explicitly) is feasible and desirable.
• [WARNING][other] The `getComputeResultUrl` method in `P2pProvider` returns `any` (an async generator of `Uint8Array` chunks), which differs significantly from `HttpProvider`'s `Promise<string>` (a URL). While the `BaseProvider`'s `getDownloadUrl` properly handles `string | DownloadResponse`, this specific method in `BaseProvider` still returns `Promise<string>`. This inconsistency in return types for `getComputeResultUrl` between HTTP and P2P paths could be a source of confusion or bugs for consumers. This method in `BaseProvider` should probably return `Promise<string | AsyncGenerator<Uint8Array>>` and be handled accordingly in calling code.
• [INFO][performance] The `getOrCreateLibp2pNode` function correctly reuses the `libp2pNode` instance based on the bootstrap key. This is good for performance and resource management, preventing unnecessary node creation and destruction.
• [INFO][other] The `sendP2pCommand` includes retry logic with `MAX_RETRIES` and `RETRY_DELAY_MS` for connection closed/reset errors. This is a good practice for improving the resilience of P2P communication against transient network issues.
• [INFO][other] In `P2pProvider.getEndpoints`, the `STATUS` command is used. It's important to note that this returns a different structure (`address`, `providerAddress`, `providerAddresses`) than the HTTP provider's `getEndpoints` which returns `serviceEndpoints`. This is handled by `BaseProvider` but could still be a nuance for callers inspecting the raw response.
• [INFO][other] The `downloadFile` function is updated to gracefully handle both HTTP URLs (strings) and pre-collected `DownloadResponse` objects (from P2P). This demonstrates good adaptability for the new P2P download flow, providing a unified interface for consumers.
• [INFO][style] Changing `delay(delayTimeout)` to an `it('should delay', ...)` block with `setTimeout` is a good practice for asynchronous testing in Mocha. It ensures that the delay is properly awaited and test timeouts are respected, making the test suite more reliable.

alexcos20 (commented):

Overall risk: medium

Summary:
This pull request introduces a significant architectural change by integrating P2P transport capabilities into the `ocean.js` library. The `Provider` class has been refactored into a `BaseProvider` facade that dynamically selects between `HttpProvider` and a new `P2pProvider` based on the provided node URI. This enables DDO resolution, file info retrieval, and compute operations over P2P. 

Key changes include: 
- **P2P Integration**: Full `libp2p` implementation for a new transport layer.
- **Refactoring**: `Provider.ts` is now a lightweight dispatcher to `HttpProvider.ts` (existing logic) and `P2pProvider.ts` (new logic).
- **CI/CD Updates**: Node.js version bumped to 22, and integration tests are matrixed to cover both HTTP and P2P scenarios.
- **Dependency Management**: Extensive additions of `libp2p` related packages.
- **Syntax Modernization**: `import assertions` updated from `assert` to `with`.

Overall, this is a substantial enhancement adding a critical new feature, but it also increases complexity and introduces a new set of dependencies and potential failure modes.

Comments:
• [WARNING][other] Updating Node.js to version 22 is a significant jump. While keeping up-to-date is good, Node.js 20 is the current LTS. Moving to an even newer non-LTS (or recently-LTS'd) version might introduce compatibility issues with existing dependencies or build tools that haven't fully adapted yet. Consider if Node.js 20 LTS would be a more stable choice for CI, or if Node.js 22 is strictly necessary for specific features.
• [WARNING][bug] The coverage upload is conditional on `matrix.transport == 'http'`. This means code paths specific to the new P2P implementation will not have their coverage reported. It's crucial to ensure P2P code paths are also adequately covered by integration tests and included in coverage reports to prevent regressions and ensure code quality for this new, complex feature.
• [INFO][other] Good catch to extract the bare peer ID for P2P decrypt routing. This demonstrates understanding of the different formats expected by HTTP vs. P2P endpoints.
• [WARNING][other] The `Aquarius` class now acts as a dispatcher for DDO resolution, delegating to `ProviderInstance` if `aquariusURL` is a P2P URI. This means `aquariusURL` can now effectively be a 'nodeUrl' capable of P2P. This is a significant conceptual shift for Aquarius's role. It would be beneficial to add comments explaining this, and ensure `ProviderInstance`'s P2P setup (`setupP2P`) is always called before Aquarius is used with a P2P URI.
• [INFO][style] This is a massive refactor of `Provider.ts`, moving all the implementation details into `HttpProvider.ts` and `P2pProvider.ts`. This is a very clean and maintainable approach for supporting multiple transport protocols. Great job on the abstraction!
• [INFO][performance] `connectionManager: { maxConnections: 100 }` - This is a reasonable default. Depending on usage patterns and network characteristics, this value might need tuning in production for optimal performance and resource usage.
• [WARNING][other] Silently catching and ignoring errors during `node.dial` for newly discovered peers (`.catch(() => {})`) could mask important connection issues during P2P discovery. While it's in the event listener for 'peer:discovery', it might be better to log these errors at a debug or info level at least, to aid in troubleshooting P2P connectivity problems.
• [WARNING][bug] The `bufToHex` function's logic `if (typeof val === 'string') { try { val = JSON.parse(val) } catch { return val } }` followed by `const bytes = Array.isArray(val?.data) ? val.data : typeof val?.[0] === 'number' ? Object.values(val) : null` seems overly complex and potentially brittle. If `val` is expected to be a JSON string representation of a Buffer, it should probably just `Buffer.from(JSON.parse(val).data).toString('hex')`. If it's expected to be a `Buffer` or `Uint8Array`, then `Buffer.from(val).toString('hex')` is sufficient. This current implementation tries to handle multiple inconsistent input types, which might lead to unexpected behavior. It also uses `Object.values(val)` which is not guaranteed to maintain order for integer keys in older JS engines, although modern ones do. Let's simplify or clarify the expected input `val`.
• [INFO][performance] The retry mechanism for P2P commands is a good addition for network resilience. `DEFAULT_MAX_RETRIES` and `DEFAULT_RETRY_DELAY_MS` are set, but it would be good to make these configurable via `P2PConfig` to allow consumers to tune their P2P interaction strategy.
• [INFO][other] The `getDownloadUrl` in `P2pProvider` now returns a `DownloadResponse` (containing an `ArrayBuffer`), which is directly consumable by `FetchHelper.downloadFile`. This is a good design, ensuring consistency in the download interface across HTTP and P2P, despite the underlying mechanism differences.
• [INFO][other] Removing `computeRoutePath` and `hasFreeComputeSupport` from examples is good, as the `ProviderInstance` now abstracts away the transport choice. This simplifies client-side code, which is a major benefit of the refactoring.
• [INFO][other] The removal of the `freeComputeRouteSupport` flag and the conditional logic simplifies the `ComputeFlow` test significantly, showcasing the benefit of the new `BaseProvider` abstraction. This implies the `ProviderInstance` is now expected to handle the route determination internally.
• [WARNING][bug] The hardcoded IP address `172.15.0.5` for local P2P bootstrap in the test setup (`_P2PWarmup.test.ts`) is specific to a Docker Compose network configuration. This makes the test less portable and prone to failure if the local environment deviates from this setup. It should ideally be retrieved dynamically (e.g., from an environment variable or Docker network inspection) or made configurable in `P2PConfig` for testing purposes.

alexcos20 (commented):

Overall risk: medium

Summary:
This pull request introduces a significant architectural change by refactoring the `Provider` service to support both HTTP and P2P communication protocols. The existing `Provider` class is now a facade (`BaseProvider`) that dispatches requests to either `HttpProvider` or `P2pProvider` based on the provided node URI. This enables `ocean.js` to interact with Ocean Protocol services over P2P, enhancing decentralization and resilience. Key changes include adding new `libp2p` dependencies, updating CI workflows to Node.js 22 and incorporating P2P integration tests, and adjusting existing API calls (e.g., `getDownloadUrl`) to handle different return types for P2P streams vs. HTTP URLs. The documentation examples for compute-to-data have been simplified to reflect the new, more abstract provider usage.

Comments:
• [INFO][other] The introduction of a matrix strategy for `test_integration` to include both `http` and `p2p` transports is excellent for ensuring compatibility and testing the new P2P functionality. This is a critical step for validating the core changes.
• [INFO][refactor] The removal of `computeRoutePath` and `hasFreeComputeSupport` variables and associated conditional logic from `ComputeExamples.md` is a good simplification. It reflects the abstraction where `ProviderInstance` now internally handles the transport logic without requiring explicit route checks, making the API cleaner for consumers.
• [INFO][feature] The addition of numerous `libp2p` related dependencies (e.g., `@chainsafe/libp2p-noise`, `@libp2p/bootstrap`) clearly indicates the integration of P2P capabilities. This is a core new feature. Consider adding a note in the README or upgrade guide about potential bundle size impact for browser environments.
• [STYLE][other] Changing `assert { type: 'json' }` to `with { type: 'json' }` is a syntax update conforming to a newer TC39 proposal. This is a good maintenance update, but ensure runtime compatibility if targeting older Node.js versions (though CI suggests Node.js 22 is targeted).
• [WARNING][bug] The logic to strip the P2P multiaddr part from `metadataDecryptorUrl` before calling `setMetaData` is a good fix for P2P compatibility, ensuring only the bare peer ID is used by the contract. However, this implies the contract or downstream services expect a bare peer ID. Please verify if the contract *only* supports bare peer IDs or if multiaddrs (e.g., `/ip4/.../p2p/...`) should be fully supported if the P2P service needs connection details.
• [INFO][feature] Integrating `isP2pUri` checks within `Aquarius` to delegate DDO resolution to `ProviderInstance` for P2P URIs is a clean way to extend Aquarius functionality to P2P. This makes Aquarius transport-agnostic.
• [WARNING][refactor] This file undergoes a massive refactoring. The original `Provider` class is completely replaced by a facade `BaseProvider` (from `BaseProvider.js`). This is a significant architectural change. While good for modularity, it represents a potentially breaking change for any consumer directly instantiating `Provider` or relying on its internal structure, even if `ProviderInstance` now points to the new `BaseProvider`. This should be clearly communicated in release notes.
• [INFO][refactor] The `isP2pUri` function is central to dispatching between HTTP and P2P providers. The current logic checks for leading `/` or common Peer ID prefixes. This approach is practical but might need review if new multiaddr formats or Peer ID versions emerge that are not covered. Ensure it's robust against different valid multiaddr forms representing P2P endpoints.
• [WARNING][security] Hardcoding `DEFAULT_BOOTSTRAP_PEERS` means these are centralized entry points for DHT discovery. While necessary, it could be a single point of failure if these bootstrap nodes are down or malicious. It's recommended to make these configurable via environment variables or a configuration object, especially in production environments, to allow for more flexible and secure bootstrapping.
• [ERROR][bug] The `getComputeResultUrl` method in `P2pProvider` returns an async generator (`result`) for streaming data directly, while its `HttpProvider` counterpart and the API definition (`Promise<string>`) expect a URL string. This is a significant discrepancy and a breaking change in behavior. Callers that expect a string URL will break when using P2P. This requires either: 
1.  Making the return type consistent (e.g., both return a stream or both return a URL that resolves to a stream).
2.  Clearly documenting this breaking change and requiring explicit type checks in consuming code.
3.  Potentially wrapping the P2P stream in a pseudo-URL or a more unified `DownloadResponse` object as `FetchHelper.ts` does for `downloadFile`.
• [WARNING][performance] The `bufToHex` function seems to perform JSON parsing on `val` if it's a string, then checks for `Buffer` type. This might introduce unnecessary overhead if `val` is frequently a non-JSON string. Also, directly checking `val?.type === 'Buffer'` is less robust than `Buffer.isBuffer(val)` or `val instanceof Uint8Array`. It would be more efficient to handle `Uint8Array` or `Buffer` first, and only attempt `JSON.parse` if it's a string that explicitly requires it for its expected structure.
• [INFO][feature] Updating `downloadFile` to accept `string | DownloadResponse` and directly return the `DownloadResponse` if provided is an elegant solution to accommodate the P2P streaming behavior while maintaining a fallback for HTTP URLs. This helps abstract the underlying transport for direct file downloads.
• [INFO][refactor] The removal of `let providerInstance: Provider` and direct usage of `ProviderInstance` reflects the new singleton facade pattern. This simplifies the tests and aligns with the new architecture.
• [WARNING][test] The `bootstrapPeers` logic here (`nodeUrl.startsWith('/') ? [nodeUrl] : [...]`) might be brittle. The `isP2pUri` function in `BaseProvider` has more comprehensive checks. Consider centralizing the determination of multiaddr types and configuring `bootstrapPeers` based on a clearer parsing of `NODE_URL` to avoid potential misconfigurations, especially if the `NODE_URL` is a bare peer ID that should use the default bootstrap peers but is instead treated as its own bootstrap peer.

alexcos20 (commented):

Overall risk: medium

Summary:
This pull request introduces significant P2P capabilities to `ocean.js` by refactoring the `Provider` service into a polymorphic `BaseProvider` class. It now dispatches requests to either `HttpProvider` or a newly implemented `P2pProvider` based on the `nodeUri` format (HTTP URL vs. P2P Multiaddr/PeerId). This enables `ocean.js` to communicate with Ocean Nodes via both traditional HTTP and the P2P network. The CI pipeline has been updated to include integration tests for both transport types, and documentation examples have been simplified to reflect the new, unified `ProviderInstance` interface. Key `libp2p` dependencies have been added to support the new networking layer.

Comments:
• [INFO][style] Changed single quotes to double quotes. This is a consistent style update across the CI file, improving uniformity.
• [INFO][other] Updated Node.js version from 20 to 22. This ensures compatibility with the latest Node.js LTS and allows leveraging new language features and performance improvements.
• [INFO][other] Introduced a build matrix for `test_integration` to run tests with both `http` and `p2p` transports. This is crucial for validating the new P2P functionality and ensuring existing HTTP flows remain functional.
• [INFO][documentation] Removed `computeRoutePath` and `hasFreeComputeSupport` variables. This simplifies the compute examples, as the `ProviderInstance` now abstracts away the transport layer details, making the client-side code cleaner.
• [WARNING][performance] Added a significant number of `libp2p` related dependencies. While essential for P2P functionality, this will increase the bundle size of `ocean.js`. For browser environments, consider a strategy for conditional loading or tree-shaking if P2P is not always required to optimize bundle size and load times.
• [INFO][style] Updated JSON module import syntax from `assert { type: 'json' }` to `with { type: 'json' }`. This aligns with the standardized ECMAScript Modules (ESM) JSON import proposal.
• [INFO][other] Added logic to extract a bare peer ID from `metadataDecryptorUrl` when a P2P multiaddr is provided. This is critical for correctly routing decryption requests through the P2P network, as the indexer expects a direct peer ID.
• [INFO][refactor] The `Aquarius` service now dynamically checks if the `aquariusURL` is a P2P URI using `isP2pUri` and delegates requests to `ProviderInstance.resolveDdo` or `ProviderInstance.validateDdo` when appropriate. This correctly integrates the new P2P transport layer for DDO resolution and validation.
• [INFO][refactor] This file has been refactored into a thin wrapper (`BaseProvider`) that acts as a facade. The core logic for HTTP and P2P communication has been moved into `HttpProvider.ts` and `P2pProvider.ts` respectively. This improves modularity and separation of concerns.
• [INFO][refactor] New file `BaseProvider.ts` introduced. This class serves as the main entry point for provider interactions, dynamically selecting `HttpProvider` or `P2pProvider` based on the provided `nodeUri`. This is a robust design for supporting multiple transport protocols transparently.
• [INFO][refactor] New file `HttpProvider.ts` introduced, encapsulating all HTTP-specific provider interaction logic. This cleans up the original `Provider.ts` file and clearly defines the HTTP communication responsibilities.
• [INFO][refactor] New file `P2pProvider.ts` introduced, containing all the logic for P2P communication using `libp2p`. This is the core of the P2P feature, implementing all necessary commands and handling the underlying network stack.
• [INFO][documentation] The `getComputeResultUrl` method in `P2pProvider` is named `*Url` but returns the direct result from `sendP2pCommand`, which might be actual data or an object, not necessarily a URL string, unlike its HTTP counterpart. This could be confusing. Consider renaming it to `getComputeResultData` or similar if it's primarily returning data, or clarify its expected return type in documentation.
• [INFO][performance] The `dhtLookupTimeout` defaults to 60 seconds. While necessary for DHT peer discovery, this is a long timeout. Users should be aware of this potential delay, especially for initial P2P connections or in environments with high network latency/sparse DHT. Documentation should highlight this.
• [INFO][refactor] The `downloadFile` function has been enhanced to accept `string | DownloadResponse`. This allows it to directly handle pre-collected binary data (likely from P2P transport) or fetch files via a URL (for HTTP transport), providing a unified interface for file downloads across both protocols.
• [INFO][testing] Replaced direct `delay()` call with a `it('should delay', ...)` block using `setTimeout`. This is the idiomatic way to handle asynchronous delays within Mocha tests, providing better control over test timeouts.
• [INFO][refactor] Removed the explicit instantiation of `Provider` as `providerInstance`. The `ProviderInstance` is now a globally shared singleton (`BaseProvider`) which is initialized (warmed up for P2P) by the `_P2PWarmup.test.ts` suite. This reflects the new singleton facade pattern.
• [INFO][testing] New test file `_P2PWarmup.test.ts` added. This is a critical addition to ensure the P2P `libp2p` node is correctly initialized and connected before other integration tests run. The underscore prefix ensures it runs first.
• [WARNING][security] The P2P bootstrap peer construction `/ip4/172.15.0.5/tcp/9001/ws/p2p/${nodeUrl}` hardcodes a Docker internal IP. This is acceptable for CI environments but should not be used in production configurations, as it's not a publicly resolvable or stable address for generic P2P peer discovery. Ensure clear documentation for users about proper bootstrap peer configuration.

alexcos20 (commented):

Overall risk: high

Summary:
This pull request introduces a significant architectural change to `ocean.js` by refactoring the `Provider` service to support both HTTP and P2P communication. The original `Provider` class is split into `HttpProvider` and `P2pProvider`, with a new `BaseProvider` (exposed as `ProviderInstance`) acting as a facade to delegate requests based on the `nodeUri`. It adds `libp2p` dependencies, updates CI workflows to test both transports, and adjusts existing services and examples to use the new architecture. A new P2P warmup test is also included.

Comments:
• [WARNING][performance] The `Upload coverage` step is conditionally run only for the `http` transport. While P2P is a new experimental feature, it's generally good practice to collect code coverage for all significant test permutations to ensure quality and prevent regressions across different operational modes. Consider adding coverage collection for `p2p` tests as well, even if it's stored as a separate artifact.
• [INFO][style] Changing `assert { type: 'json' }` to `with { type: 'json' }` is a good update, aligning with the current ECMAScript proposal for import assertions.
• [INFO][other] The logic to extract a bare peer ID for P2P decrypt routing (`metadataDecryptorUrl?.includes('/p2p/')`) is crucial for correctly handling P2P multiaddrs in the `setMetaData` contract call. This is a well-placed adjustment for P2P compatibility.
• [INFO][style] Removal of `console.log` for nonce is a good cleanup, especially if the log is not essential for debugging in production.
• [INFO][other] This file is almost entirely replaced, indicating a successful and extensive refactoring to introduce the new `BaseProvider` and `HttpProvider`/`P2pProvider` split. This is the core of the architectural change.
• [INFO][design] The `BaseProvider` class acting as a facade is a solid design pattern for abstracting the underlying transport mechanism (HTTP vs. P2P). This improves maintainability and allows for future transport extensions without modifying consumer code.
• [WARNING][security] Hardcoding `DEFAULT_BOOTSTRAP_PEERS` could be a single point of failure. While common, consider making these more easily configurable or dynamically discoverable from a trusted source, especially if the network topology changes frequently. For testing purposes, it's acceptable, but for production use cases, a more resilient bootstrap mechanism might be needed.
• [INFO][style] The use of `as any` for `this.p2pConfig.libp2p` could hide potential type mismatches if a user provides an incorrectly typed partial `Libp2pOptions` object. While convenient for merging configs, it reduces type safety. Consider if specific overrides could be typed more strictly or if the `Libp2pOptions` type allows for partial deep merges.
• [WARNING][performance] The `node.dial` call for discovered peers uses `AbortSignal.timeout(10000)`. If many peers are discovered simultaneously, this could lead to a large number of concurrent dial attempts that might strain resources or hit rate limits on some networks. Consider adding a small delay or a concurrent queue for dialing to manage this more gracefully, or limit the number of concurrent dials.
• [INFO][logic] The `bufToHex` function currently uses `Buffer.from(val.data).toString()` which defaults to 'utf8' encoding, and `uint8ArrayToString` also defaults to 'utf8'. If the intent is truly hex representation for fields like `result.hash` or `result.publicKey` (as seen in `validateDdo`), then `toString('hex')` should be explicitly used for `Buffer` and `uint8ArrayToString(..., 'base16')` for `uint8arrays`. Please clarify the expected output format for these fields and adjust accordingly.
• [WARNING][performance] The `lp.read` calls in `sendP2pCommand` and `getDownloadUrl` use `AbortSignal.timeout(DEFAULT_DIAL_TIMEOUT_MS)`. While important for preventing hangs, this timeout should be carefully chosen. For large data streams, a short timeout might prematurely close connections. Consider if the timeout should be configurable per operation or adjusted based on expected data size/stream duration for operations like `computeStreamableLogs` or `getComputeResult`.
• [ERROR][bug] There appears to be a bug here. `initializePSVerification` in `P2pProvider` is calling `PROTOCOL_COMMANDS.POLICY_SERVER_PASSTHROUGH` instead of `PROTOCOL_COMMANDS.INITIALIZE_PS_VERIFICATION`. This should be corrected to ensure the correct command is sent to the P2P node.
• [INFO][design] Modifying `downloadFile` to accept `string | DownloadResponse` allows it to seamlessly handle both HTTP URLs and pre-collected P2P binary data, which is a good adaptation for the new P2P transport. This ensures a consistent API for consumers regardless of the underlying transport.
• [WARNING][configuration] The hardcoded IP `'/ip4/172.15.0.5/tcp/9001/ws/p2p/${nodeUrl}'` for `bootstrapPeers` is specific to a local Docker/CI environment setup. This should be made more flexible (e.g., configurable via environment variables or a dedicated test config) to avoid breaking if the test environment IP changes or if P2P nodes are deployed differently. It's fine for initial testing, but consider long-term maintainability.

src/services/providers/P2pProvider.ts
```diff
@@ -0,0 +1,1160 @@
...
+      datasets: assets,
+      algorithm,
+      environment: computeEnv,
+      payment: { chainId, token, resources },
+      maxJobDuration: validUntil,
+      consumerAddress,
+      nonce,
+      signature
+    }
+    if (dockerRegistryAuth) body.dockerRegistryAuth = dockerRegistryAuth
```
ndrpp: shouldn't this be ecies encrypted the same as in the http handler? this applies to the multiple occurrences in this file

giurgiur99: Good catch, I fixed it in last commit

ndrpp (approved):

alexcos20 (commented):

Overall risk: medium

Summary:
This pull request introduces a significant architectural change by integrating P2P capabilities into `ocean.js`. The `Provider` class has been refactored into a `BaseProvider` facade that delegates requests to either an `HttpProvider` or a `P2pProvider` based on the `nodeUri` scheme. This allows `ocean.js` to communicate with both traditional HTTP-based Ocean nodes and new P2P nodes transparently. Key P2P features include `libp2p` node setup, DHT-based peer discovery, and streaming support for compute logs and results. The CI/CD workflow (`ci.yml`) has been updated to use Node.js 22 and includes matrix testing for both HTTP and P2P transports. Dependencies (`package.json`, `package-lock.json`) reflect the addition of numerous `libp2p` and `multiformats` packages, and the removal of some RDF-related dependencies. Code examples and integration tests have been updated to align with the new `ProviderInstance` usage and the abstraction of HTTP/P2P details. Overall, this is a major improvement for network flexibility and future-proofing.

Comments:
• [INFO][style] Updated Node.js version to `22` in build and test jobs. This is a good step to keep dependencies current and benefit from performance improvements and newer language features. Please ensure Node.js 22 is stable and fully supported across all development and deployment environments.
• [INFO][other] The `--with-typesense` flag was removed from the `start_ocean.sh` script. Could you please clarify if Typesense is no longer a required component for the integration tests, or if its management has moved elsewhere? If it's no longer needed, it's a good simplification.
• [INFO][other] Added extensive logging and inspection commands for `ocean-node-1` startup failures. This significantly improves debuggability for the CI pipeline, which is great for troubleshooting integration test issues.
• [WARNING][performance] The `test_integration` job now uses a strategy matrix to run tests for both `http` and `p2p` transports, which is excellent for ensuring compatibility. However, coverage is only uploaded for `http` transport. Is there a plan to enable coverage for `p2p` integration tests as well, or is this an acceptable temporary limitation? Not having coverage for `p2p` could hide potential issues.
• [INFO][other] The changelog clearly documents a significant number of `v7.0.0-next` releases with detailed commit messages, reflecting the substantial P2P implementation work. This is good for release tracking and understanding the development progression.
• [INFO][style] Replacing `if (process.env.NODE_URL) { config.oceanNodeUri = process.env.NODE_URL }` with `Object.assign(config, getNodeEndpointConfig())` centralizes environment variable handling for `oceanNodeUri`. This improves code maintainability and consistency.
• [INFO][cleanup] Removed `computeRoutePath` and `hasFreeComputeSupport` variables. This simplifies the compute examples, implying that the new `ProviderInstance` abstraction handles these concerns internally or makes them unnecessary for basic usage.
• [INFO][other] Updated `ProviderInstance.computeStatus` and `ProviderInstance.initializeCompute` to directly accept `consumerAccount` instead of `await consumerAccount.getAddress()`. This is a welcome API simplification for consumers of the library, as the provider can now internally resolve the address from the signer.
• [INFO][other] Updated package version to `7.0.0-next.6`. This indicates a major version bump, likely due to the breaking changes introduced by the P2P refactoring. Ensure all consumers of the library are aware of this change.
• [INFO][style] Improved the `exports` field in `package.json` for better module resolution across different environments (ESM, CJS, UMD, types). This is a good practice for modern JavaScript libraries.
• [INFO][cleanup] Added a `files` field to `package.json` to explicitly specify published package contents. This helps keep package size small and prevents unintended files from being published.
• [INFO][cleanup] Removed `@rdfjs/dataset`, `@rdfjs/formats-common`, `@zazuko/env-node` from `dependencies`. This is a positive change if these are no longer required for core runtime functionality, potentially reducing bundle size. Please confirm that DDO processing or other related functionalities do not implicitly rely on these anymore.
• [INFO][cleanup] Moved `@oceanprotocol/contracts` from `dependencies` to `devDependencies`. This is a good practice, indicating that the core library does not have a runtime dependency on the contracts package, but it's still needed for development and testing. This also helps reduce production bundle size.
• [INFO][other] Added numerous `libp2p`-related packages to `devDependencies`. This reflects the substantial P2P integration. While these are dev dependencies, consider the overall impact on local development environments and CI build times.
• [INFO][architecture] The `Provider.ts` file has been completely refactored to act as a facade (`BaseProvider`), delegating calls to `HttpProvider` or `P2pProvider` based on the `nodeUri`. This is an excellent architectural decision, providing a clean abstraction layer and enabling seamless transition between HTTP and P2P communication without modifying consumer code.
• [INFO][architecture] The new `BaseProvider.ts` introduces common utility functions and the main dispatching logic for HTTP and P2P. This modular approach significantly improves code organization and reusability.
• [INFO][architecture] The creation of `HttpProvider.ts` cleanly separates the HTTP-specific implementation details from the core `Provider` interface. This is crucial for maintaining modularity and allowing future network protocol additions.
• [INFO][architecture] The `P2pProvider.ts` file encapsulates all `libp2p` related logic, including node setup, peer discovery, and command sending. This is a complex but well-structured implementation for P2P communication.
• [INFO][other] The `setupP2P` method allows configuring the `libp2p` node, including bootstrap peers and timeouts. This flexibility is important for adapting to different network environments and optimizing P2P performance.
• [WARNING][security] In `dialByPeerId`, the last resort `node.dial(peerId, { signal: AbortSignal.timeout(10_000) })` does not use the originally passed `signal`. This could lead to a less controlled abort behavior if the consumer expects the operation to respect their signal. Consider passing the `signal` if it's still active, or explaining why a fixed timeout is preferred here.
• [INFO][other] The `sendP2pCommand` method includes retry logic for stream resets/closed connections. This improves the robustness of P2P communication, which is prone to transient network issues. The default retry mechanism seems reasonable.
• [INFO][other] The `getComputeResult` method for P2P now supports resumable downloads via an `offset` parameter. This is a valuable enhancement for handling large compute results and improving user experience.
• [INFO][other] Added P2P handling for `resolveDdo` and `validateDdo` directly within `Aquarius`. This ensures that Aquarius can leverage the P2P transport if configured, enhancing network flexibility.
• [WARNING][bug] The regex `metadataDecryptorUrl.match(/\/p2p\/([^/]+)/)` extracts a bare peer ID. This is correctly handled if `metadataDecryptorUrl` is indeed a multiaddr. However, if `metadataDecryptorUrl` is *not* a multiaddr (e.g., a plain URL, or an already bare peer ID), this regex might not match or might incorrectly parse. The current fallback `?? metadataDecryptorUrl` handles cases where it doesn't match the multiaddr pattern, which is good. But consider adding a clearer check for whether it's an HTTP URL vs. a P2P multiaddr or peer ID at the entry point of `setMetaData` to ensure the correct `decryptorUrl` format is always passed to the `nftContract`.
• [INFO][other] Modified `downloadFile` to accept `urlOrData: string | DownloadResponse`. This allows P2P responses (which are already collected binary data) to be passed through directly, avoiding unnecessary re-fetching and improving efficiency for P2P downloads.
• [INFO][testing] A dedicated `_P2PWarmup.test.ts` suite is a good idea to ensure the `libp2p` node is initialized once and reused across integration tests. This helps prevent redundant setup and potential test flakiness. The check for `isP2pUri(nodeUrl)` ensures this is only run when relevant.
• [INFO][testing] Added a new test case `Get compute result as stream` to validate the streaming capabilities of the P2P provider. This is crucial for ensuring the new streaming functionality works as expected.

src/services/providers/HttpProvider.ts
```diff
@@ -0,0 +1,1573 @@
...
+    payload.dataset = datasets[0]
+    // new field for C2D v2
+    payload.datasets = datasets
+    payload.algorithm = algorithm
+    if (metadata) payload.metadata = metadata
+    if (additionalViewers) payload.additionalViewers = additionalViewers
+    if (dockerRegistryAuth) {
+      const nodeKey = await this.getNodePublicKey(nodeUri)
+      if (nodeKey) {
+        payload.dockerRegistryAuth = eciesencrypt(
```
dnsi0: Shouldn't the payload field be `encryptedDockerRegistryAuth`?

src/services/providers/P2pProvider.ts
```diff
@@ -0,0 +1,1425 @@
...
+      signature
+    }
+    if (metadata) body.metadata = metadata
+    if (additionalViewers) body.additionalViewers = additionalViewers
+    if (policyServer) body.policyServer = policyServer
+    if (queueMaxWaitTime) body.queueMaxWaitTime = queueMaxWaitTime
+    if (dockerRegistryAuth) {
+      const nodeKey = await this.getNodePublicKey(nodeUri)
+      if (nodeKey)
+        body.dockerRegistryAuth = eciesencrypt(
```
dnsi0: Shouldn't the body field be `encryptedDockerRegistryAuth`?

src/services/providers/P2pProvider.ts
```diff
@@ -0,0 +1,1425 @@
...
+
+  public async getNodeJobs(
+    nodeUri: string | Multiaddr[],
+    jobId?: string,
+    fromTimestamp?: number,
+    signal?: AbortSignal
+  ): Promise<NodeComputeJob[]> {
+    try {
+      const body: Record<string, any> = {}
+      if (jobId) body.jobId = jobId
```
dnsi0: Is the `JOBS` command accepting jobId?

giurgiur99:
@dnsi0 Good catch, please check latest commit

dnsi0 (approved):

src/services/providers/HttpProvider.ts
```diff
@@ -0,0 +1,1572 @@
...
+   * @param {string} computeEnv The compute environment.
+   * @param {string} token The payment token address.
+   * @param {number} validUntil  The job expiration date.
+   * @param {string} nodeUri The provider URI.
+   * @param {SignerOrAuthToken} signerOrAuthToken Signer or auth token
+   * @param {ComputeResourceRequest[]} resources The resources to start compute job with.
+   * @param {number} chainId The chain used to do payments
+   * @param {any} policyServer Policy server data.
+   * @param {AbortSignal} signal abort signal
+   * @param {dockerRegistryAuth} dockerRegistryAuth Docker registry authentication data.
```
ndrpp, line 476: do we no longer need the `dockerRegistryAuth` and `output` fields on initializeCompute?

PR #2066, opened 2026-04-09
expose client protocols

alexcos20:
Changes proposed in this PR:

- expose client protocols in p2p
- although there is no actual handling of those now, on the network side we could have a classification of usage or filtering on additional tools


alexcos20 (commented):

Overall risk: medium

Summary:
This pull request introduces an `additionalRoles` configuration to the P2PProvider, enabling the advertising of new `libp2p` client protocols (e.g., `/ocean/client/1.0.0` and role-specific variants). The handlers for these protocols are currently placeholders, signaling intent for future functionality without implementing it. Additionally, the CI workflow has been updated to trigger on all branches.

Comments:
• [WARNING][other] Changing `branches` from `main` to `**'` means the CI workflow will now run on *all* branches, including feature branches and potentially temporary ones. While this might be intended for broader testing, it can significantly increase CI usage and costs. Please confirm this change is intentional and consider if `pull_request` triggers are sufficient, or if a more specific branch pattern is desired (e.g., `['main', 'develop', 'release/**']`).
• [INFO][other] Adding `additionalRoles` to `P2PConfig` is a good way to extend the system for future client types. It would be beneficial to add a bit more context in the PR description about the anticipated use cases for these roles, beyond the examples, to help reviewers understand the long-term vision.
• [WARNING][other] The `handleProtocolCommands` method is currently a no-op handler, as indicated by the comment. While advertising the protocol (`/ocean/client/1.0.0`) signals intent for future functionality, it's important to consider if advertising a protocol without immediate functionality could confuse other peers attempting to connect or interact with it. It might be useful to explicitly state in the PR description that this is purely for protocol advertising/discovery at this stage.
• [WARNING][style] The `// eslint-disable-next-line no-unused-vars` comment for `remotePeer` and `remoteAddr` is a temporary solution. When these variables are eventually used, the disable can be removed. For now, if they are truly not needed even for logging, consider omitting their destructuring to avoid the lint warning entirely, or perhaps adding a placeholder log if useful for debugging future implementations.
• [ERROR][bug] There appears to be an extra single quote in the protocol string when handling `additionalRoles`. It currently reads `'/ocean/client/${role}/1.0.0'` which would result in a protocol like `'`/ocean/client/dashboard/1.0.0'`. It should likely be just ``/ocean/client/${role}/1.0.0` to match the format of the base client protocol.

alexcos20:
> [ERROR][bug]

fixed in https://github.com/oceanprotocol/ocean.js/pull/2066/commits/878b58bee135132f819e72f73c8bb97a7b470fb4

giurgiur99 (approved):

ndrpp (approved):

PR #2077, opened 2026-04-20
fix auth for persistent storage commands

alexcos20:
Fixes # .

Changes proposed in this PR:

- 
- 
-

### DIFF
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -546,14 +546,12 @@ export class BaseProvider {
     nodeUri: string | Multiaddr[],
     signerOrAuthToken: Signer | string,
     owner: string,
-    chainId: number,
     signal?: AbortSignal
   ): Promise<PersistentStorageBucket[]> {
     return this.getImpl(nodeUri).getPersistentStorageBuckets(
       nodeUri,
       signerOrAuthToken,
       owner,
-      chainId,
       signal
     )
   }
@@ -611,7 +609,6 @@ export class BaseProvider {
     signerOrAuthToken: Signer | string,
     bucketId: string,
     fileName: string,
-    chainId: number,
     signal?: AbortSignal
   ): Promise<PersistentStorageDeleteFileResponse> {
     return this.getImpl(nodeUri).deletePersistentStorageFile(
@@ -619,7 +616,6 @@ export class BaseProvider {
       signerOrAuthToken,
       bucketId,
       fileName,
-      chainId,
       signal
     )
   }
--- a/src/services/providers/HttpProvider.ts
+++ b/src/services/providers/HttpProvider.ts
@@ -54,11 +54,9 @@ export class HttpProvider {
     signerOrAuthToken: Signer | string,
     command: string,
     signal?: AbortSignal
-  ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
+  ): Promise<{} | { consumerAddress: string; nonce: string; signature: string }> {
     if (typeof signerOrAuthToken === 'string') {
-      throw new Error(
-        'Persistent storage operations require a Signer because these commands require nonce/signature.'
-      )
+      return {}
     }
     const providerEndpoints = await this.getEndpoints(nodeUri)
     const serviceEndpoints = await this.getServiceEndpoints(nodeUri, providerEndpoints)
@@ -1660,9 +1658,13 @@ export class HttpProvider {
       PROTOCOL_COMMANDS.PERSISTENT_STORAGE_CREATE_BUCKET,
       signal
     )
+    const headers: Record<string, string> = { 'Content-Type': 'application/json' }
+    if (typeof signerOrAuthToken === 'string') {
+      headers.Authorization = signerOrAuthToken
+    }
     const response = await fetch(route, {
       method: 'POST',
-      headers: { 'Content-Type': 'application/json' },
+      headers,
       body: JSON.stringify({
         ...authPayload,
         accessLists: payload.accessLists ?? []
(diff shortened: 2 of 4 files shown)


alexcos20 (commented):

Overall risk: high

Summary:
This pull request aims to enable persistent storage operations using an `authToken` (string) in the `Authorization` header, in addition to the existing `Signer` mechanism. It modifies the `getPersistentAuthPayload` function to return an empty payload when an auth token string is provided, and adds logic to include the auth token in the `Authorization` header for several persistent storage commands. However, the implementation introduces significant inconsistencies and potential issues related to authentication flow.

Specifically, the `Authorization` header handling varies across different persistent storage methods within the `HttpProvider`. Some methods now explicitly set the `Authorization` header *only* for string tokens, omitting it for `Signers`, while others continue to use the `this.getAuthorization` helper, which is expected to handle both cases. This inconsistency creates a high risk of authentication failures depending on the specific command and the type of authentication credential (`Signer` vs. `string` token) used. Additionally, the change implies a shift in the authentication model for persistent storage, which requires explicit confirmation regarding backend compatibility and security implications.

Comments:
• [WARNING][security] The change to `getPersistentAuthPayload` now allows a string `signerOrAuthToken` to bypass the signed payload generation, returning an empty object `{}`. This implies that persistent storage operations can now authenticate *without* a signed `consumerAddress/nonce/signature` payload, relying solely on the `Authorization` header when a string token is used.

Please confirm that this is the intended security model change and that the provider backend (Aquarius) supports this behavior (i.e., requests with an `Authorization` header but no signed payload in the body/query parameters for persistent storage operations).
• [ERROR][bug] For `PERSISTENT_STORAGE_CREATE_BUCKET`, the `Authorization` header is now explicitly added *only* if `signerOrAuthToken` is a string. If a `Signer` is used, the `Authorization` header will be omitted for this command. This is inconsistent with other persistent storage methods that use `this.getAuthorization` (which is expected to handle both string tokens and Signers). This could lead to authentication failures when a `Signer` is used with this specific command.
• [WARNING][bug] For `PERSISTENT_STORAGE_GET_BUCKETS`, the `URLSearchParams` are now populated using `...authPayload`. When a string `signerOrAuthToken` is provided, `authPayload` is an empty object, meaning `consumerAddress`, `signature`, and `nonce` will be absent from the query parameters. The provider backend must be capable of authenticating these requests solely via the `Authorization` header (if present) without these specific query parameters. Please confirm backend compatibility for this change.
• [WARNING][bug] Similar to `PERSISTENT_STORAGE_GET_BUCKETS`, for `PERSISTENT_STORAGE_LIST_FILES`, the `URLSearchParams` are now populated using `...authPayload`. When `signerOrAuthToken` is a string, `authPayload` will be empty, leading to `consumerAddress`, `signature`, and `nonce` being absent from the query. Please confirm that the backend can authenticate these requests solely via the `Authorization` header.
• [WARNING][bug] Similar to other GET methods, for `PERSISTENT_STORAGE_GET_FILE_INFO`, the `URLSearchParams` are now populated using `...authPayload`. When `signerOrAuthToken` is a string, `authPayload` will be empty, omitting `consumerAddress`, `signature`, and `nonce` from the query. Please confirm backend compatibility for authenticating these requests solely via the `Authorization` header.
• [WARNING][bug] For `PERSISTENT_STORAGE_DOWNLOAD_FILE`, the `URLSearchParams` are now populated using `...authPayload`. When `signerOrAuthToken` is a string, `authPayload` will be empty, omitting `consumerAddress`, `signature`, and `nonce` from the query. Please confirm backend compatibility for authenticating these requests solely via the `Authorization` header.

**Consistency Note:** Unlike `CREATE_BUCKET` and `DELETE_FILE`, this method (and `uploadFile`, `updateFile`, `readFile`, `checkStatus`, `deleteBucket`) continues to use `this.getAuthorization(signerOrAuthToken)` for the `Authorization` header. This inconsistency in header handling logic across persistent storage commands is a concern (see general comment).
• [ERROR][bug] Similar to `PERSISTENT_STORAGE_CREATE_BUCKET`, for `PERSISTENT_STORAGE_DELETE_FILE`, the `Authorization` header is now explicitly added *only* if `signerOrAuthToken` is a string. If a `Signer` is used, the `Authorization` header will be omitted. This inconsistent behavior (compared to methods using `this.getAuthorization`) could lead to authentication failures when a `Signer` is used with this command.
• [ERROR][style] There is a significant inconsistency in how the `Authorization` header is applied across different persistent storage methods within `HttpProvider`.

-   Methods like `createBucket` and `deleteFile` use an explicit `if (typeof signerOrAuthToken === 'string') { headers.Authorization = signerOrAuthToken }` block, meaning the header is only present for string tokens and absent for `Signers`.
-   Methods like `uploadFile`, `downloadFile`, `updateFile`, `readFile`, `checkStatus`, and `deleteBucket` still rely on `this.getAuthorization(signerOrAuthToken)` for setting the header, which is expected to handle both string tokens and `Signers` (by generating a signed message for the header).

This discrepancy leads to different authentication behaviors depending on the specific persistent storage command and the type of authentication credential provided. Please standardize the approach for setting the `Authorization` header for *all* persistent storage operations to ensure consistent and predictable authentication behavior. This might involve: 
1. Ensuring `this.getAuthorization` is robust for all persistent storage needs and using it consistently everywhere, or
2. Adopting the explicit `if (typeof signerOrAuthToken === 'string')` pattern across all relevant methods, carefully considering how `Signer`-based authentication (signed header vs. signed payload) should be handled in that pattern.
• [WARNING][security] Similar to `HttpProvider.ts`, the `getPersistentAuthPayload` in `P2pProvider` now allows a string `signerOrAuthToken` to return an empty payload. This change suggests that P2P persistent storage operations can also authenticate *without* a signed `consumerAddress/nonce/signature` payload, relying on other mechanisms. Please confirm this change in the authentication model and P2P backend compatibility.

ndrpp (approved):

PR #2078, opened 2026-04-29
rework p2p connection

alexcos20:
# PR Description

## Summary

This PR refactors provider endpoint typing and P2P connection handling to support richer node identifiers and more reliable dialing behavior.

The core change is introducing a unified `OceanNode` type (`string | NodeP2P | PeerId`) and propagating it through `BaseProvider` and `P2pProvider`, while also improving P2P address filtering and signature handling for string-based auth/signature inputs.

### 1) New shared endpoint types

In `src/@types/Provider.ts`:

- Added `NodeP2P`:
  - `nodeId: string`
  - `multiaddress?: Multiaddr[]`
- Added `OceanNode = string | NodeP2P | PeerId`
- Added required imports for `Multiaddr` and `PeerId`

This enables callers to pass:
- a plain URL/peer string,
- a structured P2P object, or
- a native libp2p `PeerId`.

### 2) Dialing p2p nodes filters out p2p-circuits, and only enables them in a retry, if needed
Fix DHT settings

### 3) Allow signatures to be passed directly (instead of just authToken/Signer flow)

 Use SignerOrAuthTokenOrSignature everywhere where auth is needed

### 4) Removed fetchNodeLogs in favor of expanding downloadNodeLogs

### DIFF
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -142,6 +142,10 @@ export class BaseProvider {
     return isP2pUri(node) ? this.p2pProvider : this.httpProvider
   }
 
+  public getP2PProvider() {
+    return this.p2pProvider
+  }
+
   public async getNonce(
     nodeUri: OceanNode,
     consumerAddress: string,
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -91,14 +91,15 @@ export function isP2pUri(node: OceanNode): boolean {
     const nodeP2p = node as NodeP2P
     if (Array.isArray(nodeP2p.multiaddress) && nodeP2p.multiaddress.length > 0)
       return true
-    try {
-      multiaddr(nodeP2p.nodeId)
-      return true
-    } catch {}
-    try {
-      peerIdFromString(nodeP2p.nodeId)
-      return true
-    } catch {
+    if (nodeP2p.nodeId) {
+      try {
+        multiaddr(nodeP2p.nodeId)
+        return true
+      } catch {}
+      try {
+        peerIdFromString(nodeP2p.nodeId)
+        return true
+      } catch {}
       return false
     }
   }
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -86,20 +86,22 @@ export function isAgentSignature(v: unknown): v is CompleteSignature {
   )
 }
 
+function isPeerIdOrMultiAddr(param: string) {
+  try {
+    multiaddr(param)
+    return true
+  } catch {}
+  try {
+    peerIdFromString(param)
+    return true
+  } catch {
+    return false
+  }
+}
 export function isP2pUri(node: OceanNode): boolean {
   if (!node) return false
   if (typeof node === 'string') {
-    // Accept either peerId or multiaddr string as P2P
-    try {
-      multiaddr(node)
-      return true
-    } catch {}
-    try {
-      peerIdFromString(node)
-      return true
-    } catch {
-      return false
-    }
+    return isPeerIdOrMultiAddr(node)
   }
 
   // NodeP2P -> p2p
@@ -108,15 +110,7 @@ export function isP2pUri(node: OceanNode): boolean {
     if (Array.isArray(nodeP2p.multiaddress) && nodeP2p.multiaddress.length > 0)
       return true
     if (nodeP2p.nodeId) {
-      try {
-        multiaddr(nodeP2p.nodeId)
-        return true
-      } catch {}
-      try {
-        peerIdFromString(nodeP2p.nodeId)
-        return true
-      } catch {}
-      return false
+      return isPeerIdOrMultiAddr(nodeP2p.nodeId)
     }
   }
 
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -45,8 +45,9 @@ export { OCEAN_P2P_PROTOCOL, type P2PConfig } from './P2pProvider.js'
 export async function getConsumerAddress(
   signerOrAuthToken: SignerOrAuthTokenOrSignature
 ): Promise<string> {
-  if (typeof signerOrAuthToken === 'string') return decodeJwt(signerOrAuthToken).address
   if (isAgentSignature(signerOrAuthToken)) return signerOrAuthToken.consumerAddress
+  if (typeof signerOrAuthToken === 'string') return decodeJwt(signerOrAuthToken).address
+
   return signerOrAuthToken.getAddress()
 }
 
@@ -56,10 +57,7 @@ export async function getSignature(
   command: string
 ): Promise<string | null> {
   if (typeof signerOrAuthToken === 'string') {
-    // it's either a signature already (0x..) or a jwt token
-    if (signerOrAuthToken.startsWith('0x')) {
-      return signerOrAuthToken
-    } else return null
+    return null
   }
   if (isAgentSignature(signerOrAuthToken)) {
     return signerOrAuthToken.signature
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -29,7 +29,9 @@ import {
   PersistentStorageCreateBucketRequest,
   PersistentStorageDeleteFileResponse,
   PersistentStorageFileEntry,
-  PersistentStorageObject
+  PersistentStorageObject,
+  OceanNode,
+  NodeP2P
 } from '../../@types/index.js'
 import { type DDO, type ValidateMetadata } from '@oceanprotocol/ddo-js'
 import { decodeJwt } from '../../utils/Jwt.js'
@@ -52,7 +54,12 @@ export async function getSignature(
   nonce: string,
   command: string
 ): Promise<string | null> {
-  if (typeof signerOrAuthToken === 'string') return null
+  if (typeof signerOrAuthToken === 'string') {
+    // it's either a signature already (0x..) or a jwt token
+    if (signerOrAuthToken.startsWith('0x')) {
+      return signerOrAuthToken
+    } else return null
+  }
   const message = String(
     String(await signerOrAuthToken.getAddress()) + String(nonce) + String(command)
   )
@@ -63,19 +70,50 @@ export function getAuthorization(signerOrAuthToken: Signer | string): string | u
   return typeof signerOrAuthToken === 'string' ? signerOrAuthToken : undefined
 }
 
-export function isP2pUri(nodeUri: string | Multiaddr[]): boolean {
-  if (Array.isArray(nodeUri)) return true
-  if (!nodeUri) return false
-  try {
-    multiaddr(nodeUri)
-    return true
-  } catch {}
-  try {
-    peerIdFromString(nodeUri)
-    return true
-  } catch {
-    return false
+export function isP2pUri(node: OceanNode): boolean {
+  if (!node) return false
+  if (typeof node === 'string') {
+    // Accept either peerId or multiaddr string as P2P
+    try {
+      multiaddr(node)
+      return true
+    } catch {}
+    try {
+      peerIdFromString(node)
+      return true
+    } catch {
+      return false
+    }
   }
+
+  // NodeP2P -> p2p
+  if ('nodeId' in node || `multiaddress` in node) {
+    const nodeP2p = node as NodeP2P
+    if (Array.isArray(nodeP2p.multiaddress) && nodeP2p.multiaddress.length > 0)
+      return true
+    try {
+      multiaddr(nodeP2p.nodeId)
+      return true
+    } catch {}
+    try {
+      peerIdFromString(nodeP2p.nodeId)
+      return true
+    } catch {
+      return false
+    }
+  }
+
+  // PeerId (libp2p) -> p2p
+  if (typeof node === 'object' && typeof (node as any).toString === 'function') {
+    const s = String((node as any).toString())
+    try {
+      peerIdFromString(s)
+      return true
+    } catch {
+      return false
+    }
+  }
+  return false
 }
 
 export class BaseProvider {
@@ -83,13 +121,12 @@ export class BaseProvider {
   private p2pProvider = new P2pProvider()
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  protected getImpl(nodeUri: string | Multiaddr[]): any {
-    if (Array.isArray(nodeUri)) return this.p2pProvider
-    return isP2pUri(nodeUri) ? this.p2pProvider : this.httpProvider
+  protected getImpl(node: OceanNode): any {
+    return isP2pUri(node) ? this.p2pProvider : this.httpProvider
   }
 
   public async getNonce(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     consumerAddress: string,
     signal?: AbortSignal,
     providerEndpoints?: any,
@@ -107,7 +144,7 @@ export class BaseProvider {
   public async encrypt(
     data: any,
     chainId: number,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     policyServer?: any,
     signal?: AbortSignal
@@ -125,7 +162,7 @@ export class BaseProvider {
   public async checkDidFiles(
     did: string,
     serviceId: string,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     withChecksum: boolean = false,
     signal?: AbortSignal
   ): Promise<FileInfo[]> {
@@ -140,7 +177,7 @@ export class BaseProvider {
 
   public async getFileInfo(
     file: StorageObject,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     withChecksum: boolean = false,
     signal?: AbortSignal
   ): Promise<FileInfo[]> {
@@ -148,7 +185,7 @@ export class BaseProvider {
   }
 
   public async getComputeEnvironments(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signal?: AbortSignal
   ): Promise<ComputeEnvironment[]> {
     return this.getImpl(nodeUri).getComputeEnvironments(nodeUri, signal)
@@ -159,7 +196,7 @@ export class BaseProvider {
     serviceId: string,
     fileIndex: number,
     consumerAddress: string,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signal?: AbortSignal,
     userCustomParameters?: UserCustomParameters,
     computeEnv?: string,
@@ -184,7 +221,7 @@ export class BaseProvider {
     computeEnv: string,
     token: string,
     validUntil: number,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     consumerAddress: string,
     resources: ComputeResourceRequest[],
     chainId: number,
@@ -217,7 +254,7 @@ export class BaseProvider {
     serviceId: string,
     fileIndex: number,
     transferTxId: string,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     policyServer?: any,
     userCustomParameters?: UserCustomParameters
@@ -235,7 +272,7 @@ export class BaseProvider {
   }
 
   public async computeStart(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     computeEnv: string,
     datasets: ComputeAsset[],
@@ -273,7 +310,7 @@ export class BaseProvider {
   }
 
   public async freeComputeStart(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     computeEnv: string,
     datasets: ComputeAsset[],
@@ -305,7 +342,7 @@ export class BaseProvider {
   }
 
   public async computeStreamableLogs(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     jobId: string,
     signal?: AbortSignal
@@ -320,7 +357,7 @@ export class BaseProvider {
 
   public async computeStop(
     jobId: string,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     agreementId?: string,
     signal?: AbortSignal
@@ -335,7 +372,7 @@ export class BaseProvider {
   }
 
   public async computeStatus(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     jobId?: string,
     agreementId?: string,
@@ -351,7 +388,7 @@ export class BaseProvider {
   }
 
   public async getComputeResultUrl(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     jobId: string,
     index: number
@@ -365,7 +402,7 @@ export class BaseProvider {
   }
 
   public async getComputeResult(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     jobId: string,
     index: number,
@@ -382,7 +419,7 @@ export class BaseProvider {
 
   public async generateAuthToken(
     consumer: Signer,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signal?: AbortSignal
   ): Promise<string> {
     return this.getImpl(nodeUri).generateAuthToken(consumer, nodeUri, signal)
@@ -392,7 +429,7 @@ export class BaseProvider {
     address: string,
     signature: string,
     nonce: string,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signal?: AbortSignal
   ): Promise<string> {
     return this.p2pProvider.generateSignedAuthToken(
@@ -407,14 +444,14 @@ export class BaseProvider {
   public async invalidateAuthToken(
     consumer: Signer,
     token: string,
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signal?: AbortSignal
   ): Promise<{ success: boolean }> {
     return this.getImpl(nodeUri).invalidateAuthToken(consumer, token, nodeUri, signal)
   }
 
   public async resolveDdo(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     did: string,
     signal?: AbortSignal
   ): Promise<any> {
@@ -422,7 +459,7 @@ export class BaseProvider {
   }
 
   public async validateDdo(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     ddo: DDO,
     signer: Signer,
     signal?: AbortSignal
@@ -430,15 +467,12 @@ export class BaseProvider {
     return this.getImpl(nodeUri).validateDdo(nodeUri, ddo, signer, signal)
   }
 
-  public async isValidProvider(
-    url: string | Multiaddr[],
-    signal?: AbortSignal
-  ): Promise<boolean> {
+  public async isValidProvider(url: OceanNode, signal?: AbortSignal): Promise<boolean> {
     return this.getImpl(url).isValidProvider(url, signal)
   }
 
   public async PolicyServerPassthrough(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     request: PolicyServerPassthroughCommand,
     signal?: AbortSignal
   ): Promise<any> {
@@ -446,7 +480,7 @@ export class BaseProvider {
   }
 
   public async initializePSVerification(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     request: PolicyServerInitializeCommand,
     signal?: AbortSignal
   ): Promise<any> {
@@ -454,7 +488,7 @@ export class BaseProvider {
   }
 
   public async downloadNodeLogs(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signer: Signer,
     startTime: string,
     endTime: string,
@@ -478,14 +512,14 @@ export class BaseProvider {
   }
 
   public async getNodeStatus(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signal?: AbortSignal
   ): Promise<NodeStatus> {
     return this.getImpl(nodeUri).getNodeStatus(nodeUri, signal)
   }
 
   public async getNodeJobs(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     fromTimestamp?: number,
     signal?: AbortSignal
   ): Promise<NodeComputeJob[]> {
@@ -515,7 +549,7 @@ export class BaseProvider {
    * For auto-signed log fetching (HTTP or P2P), use downloadNodeLogs().
    */
   public async fetchNodeLogs(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     address: string,
     signature: string,
     nonce: string,
@@ -525,7 +559,7 @@ export class BaseProvider {
   }
 
   public async createPersistentStorageBucket(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     payload: PersistentStorageCreateBucketRequest,
     signal?: AbortSignal
@@ -543,7 +577,7 @@ export class BaseProvider {
   }
 
   public async getPersistentStorageBuckets(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     owner: string,
     signal?: AbortSignal
@@ -557,7 +591,7 @@ export class BaseProvider {
   }
 
   public async listPersistentStorageFiles(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     bucketId: string,
     signal?: AbortSignal
@@ -571,7 +605,7 @@ export class BaseProvider {
   }
 
   public async getPersistentStorageFileObject(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     bucketId: string,
     fileName: string,
@@ -587,7 +621,7 @@ export class BaseProvider {
   }
 
   public async uploadPersistentStorageFile(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     bucketId: string,
     fileName: string,
@@ -605,7 +639,7 @@ export class BaseProvider {
   }
 
   public async deletePersistentStorageFile(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     signerOrAuthToken: Signer | string,
     bucketId: string,
     fileName: string,
@@ -621,14 +655,14 @@ export class BaseProvider {
   }
 
   public async fetchConfig(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     payload: Record<string, any>
   ): Promise<any> {
     return this.p2pProvider.fetchConfig(nodeUri, payload)
   }
 
   public async pushConfig(
-    nodeUri: string | Multiaddr[],
+    nodeUri: OceanNode,
     payload: Record<string, any>
   ): Promise<any> {
     return this.p2pProvider.pushConfig(nodeUri, payload)
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -103,7 +103,7 @@ export function isP2pUri(node: OceanNode): boolean {
   }
 
   // NodeP2P -> p2p
-  if (typeof node === 'object' && ('nodeId' in node || `multiaddress` in node)) {
+  if (typeof node === 'object' && ('nodeId' in node || 'multiaddress' in node)) {
     const nodeP2p = node as NodeP2P
     if (Array.isArray(nodeP2p.multiaddress) && nodeP2p.multiaddress.length > 0)
       return true
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -1,5 +1,5 @@
 import { peerIdFromString } from '@libp2p/peer-id'
-import { multiaddr, type Multiaddr } from '@multiformats/multiaddr'
+import { multiaddr } from '@multiformats/multiaddr'
 import { Signer } from 'ethers'
 import {
   StorageObject,
@@ -22,7 +22,6 @@ import {
   DownloadResponse,
   NodeStatus,
   NodeComputeJob,
-  NodeLogsParams,
   NodeLogEntry,
   PersistentStorageAccessList,
   PersistentStorageBucket,
@@ -31,7 +30,9 @@ import {
   PersistentStorageFileEntry,
   PersistentStorageObject,
   OceanNode,
-  NodeP2P
+  NodeP2P,
+  AgentSignature,
+  signerOrAuthTokenOrSignature
 } from '../../@types/index.js'
 import { type DDO, type ValidateMetadata } from '@oceanprotocol/ddo-js'
 import { decodeJwt } from '../../utils/Jwt.js'
@@ -42,15 +43,15 @@ import { P2pProvider, type P2PConfig, type P2PRequestBodyStream } from './P2pPro
 export { OCEAN_P2P_PROTOCOL, type P2PConfig } from './P2pProvider.js'
 
 export async function getConsumerAddress(
-  signerOrAuthToken: Signer | string
+  signerOrAuthToken: signerOrAuthTokenOrSignature
 ): Promise<string> {
-  return typeof signerOrAuthToken === 'string'
-    ? decodeJwt(signerOrAuthToken).address
-    : signerOrAuthToken.getAddress()
+  if (typeof signerOrAuthToken === 'string') return decodeJwt(signerOrAuthToken).address
+  if (isAgentSignature(signerOrAuthToken)) return signerOrAuthToken.consumerAddress
+  return signerOrAuthToken.getAddress()
 }
 
 export async function getSignature(
-  signerOrAuthToken: Signer | string,
+  signerOrAuthToken: signerOrAuthTokenOrSignature,
   nonce: string,
   command: string
 ): Promise<string | null> {
@@ -60,14 +61,29 @@ export async function getSignature(
       return signerOrAuthToken
     } else return null
   }
+  if (isAgentSignature(signerOrAuthToken)) {
+    return signerOrAuthToken.signature
+  }
   const message = String(
     String(await signerOrAuthToken.getAddress()) + String(nonce) + String(command)
   )
   return signRequest(signerOrAuthToken, message)
 }
 
-export function getAuthorization(signerOrAuthToken: Signer | string): string | undefined {
+export function getAuthorization(
+  signerOrAuthToken: signerOrAuthTokenOrSignature
+): string | undefined {
   return typeof signerOrAuthToken === 'string' ? signerOrAuthToken : undefined
+}
+
+export function isAgentSignature(v: unknown): v is AgentSignature {
+  return (
+    !!v &&
+    typeof v === 'object' &&
+    typeof (v as any).consumerAddress === 'string' &&
+    typeof (v as any).nonce === 'string' &&
+    typeof (v as any).signature === 'string'
+  )
 }
 
 export function isP2pUri(node: OceanNode): boolean {
@@ -146,7 +162,7 @@ export class BaseProvider {
     data: any,
     chainId: number,
     nodeUri: OceanNode,
-    signerOrAuthToken: Signer | string,
+    signerOrAuthToken: signerOrAuthTokenOrSignature,
     policyServer?: any,
     signal?: AbortSignal
   ): Promise<string> {
@@ -256,7 +272,7 @@ export class BaseProvider {
     fileIndex: number,
     transferTxId: string,
     nodeUri: OceanNode,
-    signerOrAuthToken: Signer | string,
+    signerOrAuthToken: signerOrAuthTokenOrSignature,
     policyServer?: any,
     userCustomParameters?: UserCustomParameters
   ): Promise<string | DownloadResponse> {
@@ -274,7 +290,7 @@ export class BaseProvider {
 
   public async computeStart(
     nodeUri: OceanNode,
-    signerOrAuthToken: Signer | string,
+    signerOrAuthToken: signerOrAuthTokenOrSignature,
     computeEnv: string,
     datasets: ComputeAsset[],
     algorithm: ComputeAlgorithm,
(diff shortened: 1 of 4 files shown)


alexcos20 (commented):

Overall risk: high

Summary:
The PR successfully transitions the P2P connection APIs to use a new `OceanNode` type, replacing the previous `string | Multiaddr[]` usage. The update is comprehensive and touches all provider methods consistently. The change to allow raw signature strings in `getSignature` is also a nice improvement. However, there are a few critical logic bugs introduced in `P2pProvider.getConnection` involving `PeerId` extraction from multiaddrs, unhandled exceptions when parsing `nodeId`s, incorrect filtering conditions for DHT/peerStore lookups, and an inverted length comparison for the `p2p-circuit` fallback. These need to be addressed to ensure robust P2P dialing.

Comments:
• [WARNING][bug] If `multiaddress` is provided in the `NodeP2P` object but `nodeId` is undefined, passing `nodeP2p.nodeId` to `multiaddr()` or `peerIdFromString()` could lead to unexpected exceptions depending on library versions. It's safer to guard against an undefined `nodeId`.

```diff
   if ('nodeId' in node || `multiaddress` in node) {
     const nodeP2p = node as NodeP2P
     if (Array.isArray(nodeP2p.multiaddress) && nodeP2p.multiaddress.length > 0)
       return true
-    try {
-      multiaddr(nodeP2p.nodeId)
-      return true
-    } catch {}
-    try {
-      peerIdFromString(nodeP2p.nodeId)
-      return true
-    } catch {
-      return false
-    }
+    if (nodeP2p.nodeId) {
+      try {
+        multiaddr(nodeP2p.nodeId)
+        return true
+      } catch {}
+      try {
+        peerIdFromString(nodeP2p.nodeId)
+        return true
+      } catch {}
+    }
+    return false
   }
```
• [WARNING][performance] When `nodeUri` is a valid multiaddr string that contains a peer ID, `peerIdFromString(nodeUri)` will fail and `peerId` will remain `null`. This bypasses the existing connection reuse logic. You should attempt to extract the peer ID from the multiaddr to reuse connections.

```diff
     if (nodeUri && typeof nodeUri === 'string') {
       try {
         const addr = multiaddr(nodeUri)
         addrs.push(addr)
+        const extracted = this.peerIdFromMultiaddr(addr)
+        if (extracted) peerId = peerIdFromString(extracted)
       } catch {}
       try {
-        peerId = peerIdFromString(nodeUri)
+        if (!peerId) peerId = peerIdFromString(nodeUri)
       } catch {}
     }
```
• [ERROR][bug] If `nodeP2p.nodeId` happens to be a multiaddr string (which `isP2pUri` actually considers valid), `peerIdFromString` will throw an unhandled exception here, crashing the function. It should be wrapped in a try-catch and support multiaddr extraction as a fallback.

```diff
-        if (nodeP2p.nodeId) peerId = peerIdFromString(nodeP2p.nodeId)
+        if (nodeP2p.nodeId) {
+          try {
+            peerId = peerIdFromString(nodeP2p.nodeId)
+          } catch {
+            try {
+              const addr = multiaddr(nodeP2p.nodeId)
+              addrs.push(addr)
+              const extracted = this.peerIdFromMultiaddr(addr)
+              if (extracted) peerId = peerIdFromString(extracted)
+            } catch (err) {
+              LoggerInstance.debug(`[P2P] Invalid nodeId format: ${nodeP2p.nodeId}`)
+            }
+          }
+        }
```
• [WARNING][logic] If `addrs` contains only *undialable* addresses (e.g. unsupported local protocols), `addrs.length < 1` will be false, and it will skip looking up the peer ID in the `peerStore` or DHT. This will lead to an immediate failure down the line. You should ensure there are *dialable* addresses before skipping the lookup.

```diff
+    const hasDialable = () => addrs.some((ma) => this.isDialable(ma))
-    // let's build the ma addr & fetch from dht if needed
-    if (addrs.length < 1 && peerId) {
+    // let's build the ma addr & fetch from peerStore if needed
+    if (!hasDialable() && peerId) {
       try {
         const peerData = await node.peerStore.get(peerId)
         if (peerData?.addresses) {
           for (const addr of peerData.addresses) {
             addrs.push(addr.multiaddr)
           }
           LoggerInstance.debug(
             `[P2P] ${peerId.toString()}: ${peerData.addresses.length} peerStore addrs`
           )
         }
       } catch {
         LoggerInstance.debug(`[P2P] ${peerId.toString()}: not in peerStore`)
       }
     }
-    // if there are no known ma, search dht
-    if (addrs.length < 1 && peerId) {
+    // if there are no known ma, search dht
+    if (!hasDialable() && peerId) {
```
• [ERROR][logic] `afterPFilter` is the array length *after* filtering out `p2p-circuit` addresses. It can never be strictly greater than `beforePFilter`. Therefore, `afterPFilter > beforePFilter` will always be false, and the intended fallback to dial via p2p-circuits will never be executed.

```diff
     if (addrs.length < 1) {
       // try with p2p-circuits if available
-      if (!includeP2PCircuit && afterPFilter > beforePFilter) {
+      if (!includeP2PCircuit && beforePFilter > afterPFilter) {
         return this.getConnection(nodeUri, signal, true)
       }
       throw new Error('No valid multiaddresses, cannot connect')
     }
```

alexcos20 (commented):

Overall risk: medium

Summary:
The PR introduces structured `OceanNode` objects and improves peer connection logic for Ocean Protocol's P2P network. The architectural changes correctly replace legacy multiaddr structures with typed object definitions. However, there are a few implementation details that can cause performance bottlenecks (recursive DHT lookups) and potential runtime errors.

Comments:
• [WARNING][bug] Using backticks for property names in the `in` operator is non-standard and likely a typo. Additionally, if `node` is a primitive type (e.g., a number passed from JavaScript), the `in` operator will throw a runtime `TypeError`. Consider adding a `typeof node === 'object'` check to safely validate it.

```diff
-  if ('nodeId' in node || `multiaddress` in node) {
+  if (typeof node === 'object' && ('nodeId' in node || 'multiaddress' in node)) {
```
• [INFO][performance] When a multiaddr string containing a peer ID (e.g., `/ip4/.../p2p/Qm...`) is passed, the peer ID is not immediately extracted, which skips the `getConnections` and `peerStore` optimization checks for that connection. You should utilize the existing `peerIdFromMultiaddr` method here to quickly identify already-connected peers.

```diff
     if (nodeUri && typeof nodeUri === 'string') {
       try {
         const addr = multiaddr(nodeUri)
         addrs.push(addr)
+        if (!peerId) {
+          const pidStr = this.peerIdFromMultiaddr(addr)
+          if (pidStr) peerId = peerIdFromString(pidStr)
+        }
       } catch {}
```
• [WARNING][bug] Missing parentheses on `peerId.toString()`. This will print the function signature (e.g., `[native code]`) instead of the actual peer ID string.

```diff
-        LoggerInstance.debug(
-          `[P2P] ${peerId.toString}: DHT findPeer failed: ${err.message}`
-        )
+        LoggerInstance.debug(
+          `[P2P] ${peerId.toString()}: DHT findPeer failed: ${err.message}`
+        )
```
• [ERROR][performance] If all dialable addresses are filtered out because they are `p2p-circuit` paths, the method recursively calls itself (`this.getConnection(..., true)`). This will erroneously clear out the initially parsed addresses and re-trigger the slow DHT lookups and `peerStore` queries from scratch. Furthermore, the last-resort fallback to dial by `peerId` was removed. You can optimize this by keeping the filtered arrays locally and dialing `peerId` directly as a fallback.

```diff
-    addrs = addrs.filter((ma) => this.isDialable(ma))
-    const beforePFilter = addrs.length
-    if (!includeP2PCircuit) addrs = addrs.filter((ma) => this.isNotP2PCircuit(ma))
-
-    const afterPFilter = addrs.length
-
-    if (addrs.length < 1) {
-      // try with p2p-circuits if available
-      if (!includeP2PCircuit && afterPFilter < beforePFilter) {
-        // we have some p2p-circuit addrs, let's try them
-        return this.getConnection(nodeUri, signal, true)
-      }
-      throw new Error('No valid multiaddresses, cannot connect')
-    }
+    const dialableAddrs = addrs.filter((ma) => this.isDialable(ma))
+    const directAddrs = dialableAddrs.filter((ma) => this.isNotP2PCircuit(ma))
+    
+    // Prefer direct paths if available, otherwise fallback to circuit paths
+    addrs = !includeP2PCircuit && directAddrs.length > 0 ? directAddrs : dialableAddrs
+
+    if (addrs.length < 1) {
+      if (peerId) {
+        try {
+          const conn = await node.dial(peerId, { signal })
+          LoggerInstance.debug(`[P2P] peerId dial SUCCESS via ${conn.remoteAddr}`)
+          return conn
+        } catch (err: any) {
+          // Fall through to error
+        }
+      }
+      throw new Error('No valid multiaddresses, cannot connect')
+    }
```

alexcos20 (commented):

Overall risk: high

Summary:
This PR refactors P2P address handling by introducing a unified `OceanNode` type, replacing ad-hoc `string | Multiaddr[]` logic. The connection management in `P2pProvider` has been significantly streamlined, consolidating PeerStore and DHT searches into a unified flow. However, there are a few critical issues to address: a potential runtime `TypeError` when falling back to `p2p-circuit` multiaddrs without a `peerId`, a logic flaw in `getSignature` that overrides nonce verification, and a loss of connection resilience (fallback to DHT) if explicitly provided multiaddresses fail to connect.

Comments:
• [ERROR][bug] If `nodeUri` contains a `p2p-circuit` multiaddr but no peer ID suffix, `peerId` will be `null`. The condition `afterPFilter < beforePFilter` evaluates to `true`, and calling `peerId.toString()` will throw a `TypeError: Cannot read properties of null`. Use optional chaining here and ensure the fallback object safely matches the `NodeP2P` interface.
• [ERROR][bug] Returning a pre-computed signature here completely ignores the `nonce` parameter. Functions that call `getSignature` (like `sendP2pCommand` or `getPersistentStorageSignaturePayload`) generate a fresh `nonce = String(Date.now())` and send it alongside the signature. Since the pre-computed signature was signed over a different nonce, verification on the provider node will consistently fail. Pre-computed signatures must be explicitly passed alongside their exact corresponding nonces (as is currently done in `fetchNodeLogs`) rather than intercepted here.
• [WARNING][architecture] In the previous implementation, if dialing the explicitly provided multiaddr failed, the code fell back to searching the PeerStore/DHT for alternative addresses via `dialByPeerId`. In this refactored version, if the provided `nodeUri` contains dialable addresses, `hasDialable()` is true, and PeerStore/DHT lookups are entirely skipped. If `node.dial(dialable)` subsequently fails, the error is immediately thrown. This removes connection resilience if an explicitly provided IP is offline but the peer is still reachable on another interface known to the DHT. Consider catching the dial error and allowing a fallback to DHT discovery if direct dialing fails.
• [INFO][style] Using template literal backticks for `` `multiaddress` `` is syntactically valid but inconsistent with the `'nodeId'` check on the same line. Consider using standard quotes for consistency.

alexcos20 (commented):

Overall risk: medium

Summary:
This PR introduces excellent structural improvements by adding the `OceanNode` type and `AgentSignature` interface. This centralizes pre-signed request validation and drastically improves the usability of agent-based workflows. The P2P connection logic is significantly upgraded, adding robust fallbacks (e.g., P2P-circuit relays). Overall the architecture is much cleaner. However, a few implementation issues need addressing: a header injection bug with the `fetch` API, a TypeScript interface violation during P2P recursive fallbacks, redundant logic checks, and a potential runtime crash if raw hex signature strings are evaluated as auth tokens.

Comments:
• [ERROR][bug] `this.getAuthorization(signerOrAuthToken)` returns `undefined` when the argument does not map to a valid JWT token. When `undefined` is directly passed as a dictionary value in `HeadersInit`, the `fetch` API casts it to the string `"undefined"`. You properly avoided this in `computeStop` using object spreading; the same safeguard should be applied here.
```diff
-        headers: {
-          'Content-Type': 'application/json',
-          Authorization: this.getAuthorization(signerOrAuthToken)
-        },
+        headers: {
+          'Content-Type': 'application/json',
+          ...(this.getAuthorization(signerOrAuthToken) ? { Authorization: this.getAuthorization(signerOrAuthToken) } : {})
+        },
```
• [WARNING][bug] `peerId?.toString()` evaluates to `string | undefined`. Because the `NodeP2P` interface strictly requires `nodeId` to be a `string`, passing `undefined` here violates the type contract. Provide a fallback empty string to satisfy TypeScript strict mode.
```diff
-        return this.getConnection(
-          { nodeId: peerId?.toString(), multiaddress: addrs },
-          signal,
-          true
-        )
+        return this.getConnection(
+          { nodeId: peerId ? peerId.toString() : '', multiaddress: addrs },
+          signal,
+          true
+        )
```
• [WARNING][bug] Assuming a raw hex string starting with `0x` is a valid standalone parameter might cause runtime crashes. Almost all provider methods immediately call `getConsumerAddress(signerOrAuthToken)`, which strictly attempts to decode a string via `decodeJwt(signerOrAuthToken)`. If a user actually passes a raw `0x...` string, `decodeJwt` will throw an exception. Consider removing this fallback check and documenting that callers must use the `AgentSignature` object for raw signatures instead.
```typescript
// Example: decodeJwt('0xabc...') throws an error.
```
• [INFO][style] The `isAgentSignature` method explicitly checks if the provided argument is an `object`. Therefore, if `typeof` evaluates to `'string'`, it is guaranteed not to be an `AgentSignature`. The `&& !isAgentSignature(...)` condition is logically redundant.
```diff
-    if (typeof signerOrAuthToken === 'string' && !isAgentSignature(signerOrAuthToken)) {
+    if (typeof signerOrAuthToken === 'string') {
```
• [INFO][style] Identical redundant check as above. It can be safely simplified.
```diff
-    if (typeof signerOrAuthToken === 'string' && !isAgentSignature(signerOrAuthToken)) {
+    if (typeof signerOrAuthToken === 'string') {
```
• [INFO][style] Minor syntax inconsistency: Using backticks for property lookups in an object is unnecessary and unconventional. Standard single quotes match the rest of your logic.
```diff
-  if (typeof node === 'object' && ('nodeId' in node || `multiaddress` in node)) {
+  if (typeof node === 'object' && ('nodeId' in node || 'multiaddress' in node)) {
```

alexcos20 (commented):

Overall risk: low

Summary:
This PR excellently introduces the `OceanNode` and `AgentSignature` abstractions, allowing for pre-signed requests that save network round-trips for nonce fetching. Unifying HTTP and P2P payload patterns is a great architectural win. There is one oversight in `BaseProvider`'s `validateDdo` signature that breaks the new contract, and a strict TypeScript type mismatch when attempting a circuit relay connection. Otherwise, the code is solid. LGTM!

Comments:
• [ERROR][bug] The parameter `signer: Signer` was left unchanged in the `BaseProvider` interface. It should be updated to `signerOrAuthToken: signerOrAuthTokenOrSignature` to match the underlying `P2pProvider` and `HttpProvider` implementations. Otherwise, consumers will encounter TypeScript errors when attempting to pass an `AgentSignature` here.

```diff
   public async validateDdo(
     nodeUri: OceanNode,
     ddo: DDO,
-    signer: Signer,
+    signerOrAuthToken: signerOrAuthTokenOrSignature,
     signal?: AbortSignal
   ): Promise<ValidateMetadata> {
-    return this.getImpl(nodeUri).validateDdo(nodeUri, ddo, signer, signal)
+    return this.getImpl(nodeUri).validateDdo(nodeUri, ddo, signerOrAuthToken, signal)
   }
```
• [ERROR][bug] Passing `null` for `nodeId` violates the `NodeP2P` interface (which defines `nodeId: string`). In projects with `strictNullChecks` enabled, this will cause a TypeScript compilation error. Consider passing an empty string or explicitly casting the object.

```diff
         return this.getConnection(
-          { nodeId: peerId ? peerId.toString() : null, multiaddress: addrs },
+          { nodeId: peerId ? peerId.toString() : '', multiaddress: addrs } as NodeP2P,
           signal,
           true
         )
```
• [INFO][style] The logic to conditionally resolve the `nonce` and `signature` (checking if it's an `AgentSignature` vs manually fetching via `getNonce` and signing) is duplicated heavily (~20 times) across `HttpProvider` and `P2pProvider`. Consider extracting this into a protected helper method (e.g., `protected async resolveAuthPayload(...)`) to keep the code DRY and easier to maintain.
• [INFO][other] Removing `fetchNodeLogs` and allowing `downloadNodeLogs` to seamlessly handle both auto-signed and pre-signed requests via `AgentSignature` is a fantastic API unification. Great design choice here!

alexcos20 (commented):

Overall risk: medium

Summary:
This PR introduces excellent structural improvements, cleanly abstracting P2P node types via `OceanNode` and unifying signature checks via `AgentSignature` and `getSignedCommandParams`. The DRY refactoring across endpoints is highly commendable. However, there is a critical logic bug: the strict string checks (designed originally for persistent storage methods) were copied into the shared `getSignedCommandParams`, which will currently crash all standard endpoints when a JWT AuthToken is used instead of a `Signer`.

Comments:
• [ERROR][bug] The `getSignedCommandParams` method throws an error when `signerOrAuthToken` is a string. Because this method is now shared across many standard HTTP endpoints (like `computeStart`, `downloadNodeLogs`, etc.), any API flow using an AuthToken will immediately crash. We need to allow strings here and push the strict validation back to `getPersistentStorageSignaturePayload`.

```diff
@@ -48,7 +48,7 @@
     signal?: AbortSignal,
     providerEndpoints?: any,
     serviceEndpoints?: any
-  ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
+  ): Promise<{ consumerAddress: string; nonce: string; signature: string | null }> {
     if (!providerEndpoints) providerEndpoints = await this.getEndpoints(nodeUri)
     if (!serviceEndpoints)
       serviceEndpoints = await this.getServiceEndpoints(nodeUri, providerEndpoints)
@@ -59,11 +59,6 @@
         signature: signerOrAuthToken.signature
       }
     }
-    if (typeof signerOrAuthToken === 'string') {
-      throw new Error(
-        'Persistent storage operations require a Signer or AgentSignature (nonce/signature).'
-      )
-    }
 
     const consumerAddress = await this.getConsumerAddress(signerOrAuthToken)
     const nonce = (
@@ -88,14 +83,14 @@
     signal?: AbortSignal,
     providerEndpoints?: any,
     serviceEndpoints?: any
   ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
     if (typeof signerOrAuthToken === 'string') {
       throw new Error(
         'Persistent storage operations require a Signer or AgentSignature (nonce/signature).'
       )
     }
-    return this.getSignedCommandParams(
+    return (await this.getSignedCommandParams(
       nodeUri,
       signerOrAuthToken,
       command,
       signal,
       providerEndpoints,
       serviceEndpoints
-    )
+    )) as { consumerAddress: string; nonce: string; signature: string }
   }
```
• [ERROR][bug] Similarly to the HTTP Provider, throwing an error when `signature` is falsy breaks P2P flows that legitimately rely on AuthTokens (where `getSignature` returns null by design). You should allow `signature: string | null` and only strictly enforce its presence in `getPersistentStorageSignaturePayload`.

```diff
@@ -377,21 +377,18 @@
     signerOrAuthToken: signerOrAuthTokenOrSignature,
     command: string,
     signal?: AbortSignal
-  ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
+  ): Promise<{ consumerAddress: string; nonce: string; signature: string | null }> {
     if (isAgentSignature(signerOrAuthToken)) {
       return {
         consumerAddress: signerOrAuthToken.consumerAddress,
         nonce: signerOrAuthToken.nonce,
         signature: signerOrAuthToken.signature
       }
     }
     const consumerAddress = await this.getConsumerAddress(signerOrAuthToken)
     const nonce = ((await this.getNonce(nodeUri, consumerAddress, signal)) + 1).toString()
     const signature = await this.getSignature(signerOrAuthToken, nonce, command)
-    if (!signature) throw new Error(`Could not sign command ${command}.`)
     return { consumerAddress, nonce, signature }
   }
 
@@ -1546,7 +1543,9 @@
       throw new Error(
         'Persistent storage operations require a Signer or AgentSignature (nonce/signature).'
       )
     }
-    return this.getSignedCommandParams(nodeUri, signerOrAuthToken, command, signal)
+    const params = await this.getSignedCommandParams(nodeUri, signerOrAuthToken, command, signal)
+    if (!params.signature) throw new Error('Could not sign persistent storage request.')
+    return params as { consumerAddress: string; nonce: string; signature: string }
   }
```
• [INFO][style] The refactoring of the P2P dialing fallback logic inside `getConnection` (where circuits and DHTs are systematically traversed and retried) is a massive code-quality and reliability improvement over the previous array iteration model. Great work!

alexcos20 (commented):

Overall risk: medium

Summary:
The PR successfully refactors the provider modules to support the Multi-Agent Computing Protocol (MCP). Replacing `Signer | string` with `signerOrAuthTokenOrSignature` seamlessly accommodates pre-signed operations via `AgentSignature`, which is excellent for automated multi-agent environments. The introduction of `getSignedCommandParams` elegantly DRYs up parameter extraction across API requests. However, a logic flaw in the P2P connection dialing completely drops relayed addresses (`/p2p-circuit`) if the initial direct connection fails, which will cause connectivity issues in NAT/firewall environments. There are also a couple of minor TypeScript return type mismatches introduced by the new helper functions.

Comments:
• [ERROR][bug] When filtering out `/p2p-circuit` relay addresses to prefer direct connections, the current logic correctly falls back to dialing relayed addresses **only if there were zero direct addresses available** (`dialable.length < 1`). However, if direct addresses *are* found but they **fail to connect** (e.g., due to strict NAT/firewalls), `node.dial` throws an error and the function exits, completely abandoning the relayed fallback. You should catch the error and execute the fallback recursive call before throwing.

```diff
     try {
       const conn = await node.dial(dialable, { signal })
       LoggerInstance.debug(
         `[P2P] Dial SUCCESS via ${conn.remoteAddr} (limited=${conn.limits != null})`
       )
       return conn
     } catch (err: any) {
+      if (!includeP2PCircuit && afterPFilter < beforePFilter) {
+        LoggerInstance.debug(`[P2P] Direct dial failed, falling back to relayed addresses...`)
+        return this.getConnection(
+          { nodeId: peerId ? peerId.toString() : '', multiaddress: addrs } as NodeP2P,
+          signal,
+          true
+        )
+      }
       throw new Error(
         `Cannot dial peer ${peerId?.toString()}. ` +
           (addrs.length > 0
             ? `Found addrs: ${addrs.map(String).join(', ')}. `
             : 'No addresses found. ') +
           `Active connections: ${node.getConnections().length}. ` +
           err.message
       )
     }
```
• [WARNING][style] The `getSignature` method returns `Promise<string | null>`, but the return type signature for `getSignedCommandParams` forces `signature` to be a `string`. If `strictNullChecks` is enabled, this will cause a TypeScript compiler error. Update the return type to reflect that `signature` can be `null` (e.g., when a JWT auth token string is provided instead of a Signer).

```diff
   private async getSignedCommandParams(
     nodeUri: string,
     signerOrAuthToken: signerOrAuthTokenOrSignature,
     command: string,
     signal?: AbortSignal,
     providerEndpoints?: any,
     serviceEndpoints?: any
-  ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
+  ): Promise<{ consumerAddress: string; nonce: string; signature: string | null }> {
```
• [WARNING][style] Similar to `HttpProvider`, the return type of `signature` should accept `null` to accommodate the behavior of `getSignature`.

```diff
   private async getSignedCommandParams(
     nodeUri: OceanNode,
     signerOrAuthToken: signerOrAuthTokenOrSignature,
     command: string,
     signal?: AbortSignal
-  ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
+  ): Promise<{ consumerAddress: string; nonce: string; signature: string | null }> {
```
• [INFO][other] Removing `fetchNodeLogs` in favor of expanding `downloadNodeLogs` to natively handle `AgentSignature` pre-signed payloads is an excellent refactor. It reduces code duplication and streamlines the API interface. Great design choice.

alexcos20 (commented):

Overall risk: medium

Summary:
This PR introduces crucial architectural improvements for MCP (Multi-Agent/Machine Communication Protocol) by standardizing authentication and supporting pre-signed payloads via `AgentSignature` and `signerOrAuthTokenOrSignature`. The refactoring to centralize signature parameters via `getSignedCommandParams` is excellent and drastically reduces code duplication. The new multiaddress fallback logic in `P2pProvider` enhances connection resilience. There are a few TypeScript compilation errors due to mismatched types in persistent storage payloads that need to be addressed, as well as minor stylistic improvements.

Comments:
• [ERROR][bug] `getSignedCommandParams` returns a Promise where `signature` can be `string | null`. However, `getPersistentStorageSignaturePayload` expects a return type where `signature` is strictly `string`. While you explicitly throw if `typeof signerOrAuthToken === 'string'` (which guarantees `signature` will be a string at runtime), the TypeScript compiler cannot infer this across function boundaries and will throw a type error (`Type 'string | null' is not assignable to type 'string'`). You should add a type assertion or explicit check to fix the compilation error.

```typescript
    const params = await this.getSignedCommandParams(
      nodeUri,
      signerOrAuthToken,
      command,
      signal,
      providerEndpoints,
      serviceEndpoints
    )
    return params as { consumerAddress: string; nonce: string; signature: string }
```
• [ERROR][bug] Similar to the issue in `HttpProvider.ts`, returning `getSignedCommandParams` here directly will cause a TypeScript compilation error because the expected return type requires `signature: string`, but the function returns `signature: string | null`.

```typescript
    const params = await this.getSignedCommandParams(nodeUri, signerOrAuthToken, command, signal)
    return params as { consumerAddress: string; nonce: string; signature: string }
```
• [INFO][style] The imported `getConsumerAddress` from `BaseProvider.ts` already handles the `isAgentSignature(s)` logic natively. Checking it again here is redundant and can be simplified.

```typescript
  protected getConsumerAddress(s: signerOrAuthTokenOrSignature) {
    return getConsumerAddress(s)
  }
```
• [INFO][style] As noted in `HttpProvider.ts`, the imported `getConsumerAddress` handles `isAgentSignature` natively. This redundant check can be removed.

```typescript
  protected getConsumerAddress(s: signerOrAuthTokenOrSignature) {
    return getConsumerAddress(s)
  }
```
• [WARNING][style] In TypeScript, standard convention dictates using PascalCase for exported types and interfaces. Consider renaming `signerOrAuthTokenOrSignature` to `SignerOrAuthTokenOrSignature` for consistency.

```diff
-export type signerOrAuthTokenOrSignature = string | Signer | AgentSignature
+export type SignerOrAuthTokenOrSignature = string | Signer | AgentSignature
```
• [INFO][style] Calling `this.getAuthorization(signerOrAuthToken)` twice inside the object literal is slightly redundant and can cause TypeScript strict mode to complain about potentially spreading an `undefined` value. It's cleaner to assign it to a variable first.

```typescript
        const authHeader = this.getAuthorization(signerOrAuthToken)
        const response = await fetch(computeStartUrl, {
          method: 'POST',
          body: JSON.stringify(payload),
          headers: {
            'Content-Type': 'application/json',
            ...(authHeader ? { Authorization: authHeader } : {})
          },
          signal
        })
```
*(Note: This pattern occurs in several places throughout `HttpProvider.ts`)*

alexcos20 (commented):

Overall risk: medium

Summary:
The PR successfully introduces the `AgentSignature` interface and refactors the HTTP and P2P providers to centralize signing via `getSignedCommandParams`. This enables decentralized agents (MCP) to proxy signed requests seamlessly. However, there are a few issues to address: a potential breaking change/regression regarding `Multiaddr[]` arrays, a TypeScript compilation error in strict mode due to unhandled `null` signatures, and missing deduplication of libp2p addresses.

Comments:
• [WARNING][bug] The `isP2pUri` function previously checked for `Array.isArray(nodeUri)` to support `Multiaddr[]`. Since this check was removed, passing `Multiaddr[]` will return `false`, causing `getImpl()` to return `httpProvider`. If dropping `Multiaddr[]` is intentional, ensure this is documented as a breaking change. Otherwise, consider restoring the array check to maintain backward compatibility.

```diff
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -74,6 +74,9 @@
       return false
     }
   }
+
+  // Backward compatibility for Multiaddr[]
+  if (Array.isArray(node)) return true
 
   // NodeP2P -> p2p
   if (typeof node === 'object' && ('nodeId' in node || 'multiaddress' in node)) {
```
• [ERROR][bug] `getSignedCommandParams` returns a `Promise` where `signature` is typed as `string | null`. However, this method's return type expects `signature: string`. With `strictNullChecks` enabled, this will cause a TypeScript compilation error. You should verify the signature exists and typecast the return object.

```diff
--- a/src/services/providers/HttpProvider.ts
+++ b/src/services/providers/HttpProvider.ts
@@ -110,14 +110,16 @@
         'Persistent storage operations require a Signer or AgentSignature (nonce/signature).'
       )
     }
-    return this.getSignedCommandParams(
+    const params = await this.getSignedCommandParams(
       nodeUri,
       signerOrAuthToken,
       command,
       signal,
       providerEndpoints,
       serviceEndpoints
     )
+    if (!params.signature) throw new Error('Failed to generate signature for persistent storage.')
+    return params as { consumerAddress: string; nonce: string; signature: string }
   }
```
• [ERROR][bug] Similar to the HTTP provider, returning `this.getSignedCommandParams(...)` directly throws a TypeScript error because the signature could theoretically be `null` according to the type definitions.

```diff
--- a/src/services/providers/P2pProvider.ts
+++ b/src/services/providers/P2pProvider.ts
@@ -1561,7 +1561,9 @@
         'Persistent storage operations require a Signer or AgentSignature (nonce/signature).'
       )
     }
-    return this.getSignedCommandParams(nodeUri, signerOrAuthToken, command, signal)
+    const params = await this.getSignedCommandParams(nodeUri, signerOrAuthToken, command, signal)
+    if (!params.signature) throw new Error('Failed to generate signature for persistent storage.')
+    return params as { consumerAddress: string; nonce: string; signature: string }
   }
```
• [INFO][performance] The previous logic included a `Set` to deduplicate addresses retrieved from both `peerStore` and `DHT`. Removing this means duplicate multiaddrs might be pushed to `addrs` and subsequently dialed, which can trigger redundant Libp2p dialing efforts. It is highly recommended to deduplicate them before mapping/filtering.

```diff
--- a/src/services/providers/P2pProvider.ts
+++ b/src/services/providers/P2pProvider.ts
@@ -403,7 +403,11 @@
         }
       }
     }
-    let dialable = addrs.filter((ma) => this.isDialable(ma))
+    
+    const uniqueAddrs = Array.from(
+      new Map(addrs.map((ma) => [ma.toString(), ma])).values()
+    )
+    let dialable = uniqueAddrs.filter((ma) => this.isDialable(ma))
     const beforePFilter = dialable.length
```
• [INFO][style] The creation of `getSignedCommandParams` significantly cleans up repetitive code across all provider command methods. Great use of refactoring to support new authentication schemas without bloat!

alexcos20 (commented):

Overall risk: medium

Summary:
This Pull Request introduces significant enhancements for Multi-Agent Communication Protocol (MCP) interactions by adding support for `AgentSignature`, standardizing P2P endpoints with `OceanNode`, and DRYing up command parameter generation through `getSignedCommandParams()`. However, there are TypeScript return type mismatches introduced in the persistent storage operations that will break strict builds. The transition generally exhibits excellent architectural design, but care must be taken regarding removed APIs and P2P DHT configurations.

Comments:
• [ERROR][bug] There is a TypeScript return type mismatch. `getSignedCommandParams` returns a Promise where `signature` is `string | null`, but `getPersistentStorageSignaturePayload` expects the Promise to return `signature: string`. In TS `strict` mode, this will cause a compilation error. Consider validating and explicitly casting the result.

```diff
-    return this.getSignedCommandParams(
+    const params = await this.getSignedCommandParams(
       nodeUri,
       signerOrAuthToken,
       command,
       signal,
       providerEndpoints,
       serviceEndpoints
     )
+    if (!params.signature) {
+      throw new Error('Could not sign persistent storage request.')
+    }
+    return params as { consumerAddress: string; nonce: string; signature: string }
```
• [ERROR][bug] Similar to HttpProvider, there is a TypeScript return type mismatch here since `getSignedCommandParams` can return `signature: string | null`.

```diff
-    return this.getSignedCommandParams(nodeUri, signerOrAuthToken, command, signal)
+    const params = await this.getSignedCommandParams(nodeUri, signerOrAuthToken, command, signal)
+    if (!params.signature) {
+      throw new Error('Could not sign persistent storage request.')
+    }
+    return params as { consumerAddress: string; nonce: string; signature: string }
```
• [WARNING][bug] If TypeScript `strict` or `useUnknownInCatchVariables` mode is enabled, the `err` variable inside the catch block defaults to `unknown`. Accessing `err.name` will result in a compile-time error. Add an explicit `any` type or check `instanceof Error`.

```diff
-    } catch (err) {
-      if (err.name !== 'AbortError') console.error(err)
+    } catch (err: any) {
+      if (err?.name !== 'AbortError') console.error(err)
```
• [WARNING][other] The `fetchNodeLogs` method has been removed entirely in favor of the newly refactored `downloadNodeLogs`. This constitutes a breaking change for existing API consumers utilizing `fetchNodeLogs`. Ensure this is documented in the changelog, or consider preserving the method as a deprecated wrapper.
• [INFO][performance] Setting `clientMode: false` forces the node to act as a DHT server. While this optimizes network queries (valuable for always-on MCP agents), it can consume additional bandwidth and pollute the DHT with unreachable nodes if deployed behind a NAT without a publicly dialable address. Ensure this aligns with the expected consumer environments for `ocean.js`.
• [INFO][style] The connection fallback logic inside `getConnection` is excellent. Automatically prioritizing direct dials while gracefully falling back to relayed addresses (`p2p-circuit`) if none are dialable significantly improves connection robustness.

src/services/providers/BaseProvider.ts
```diff
@@ -40,56 +43,111 @@ import { P2pProvider, type P2PConfig, type P2PRequestBodyStream } from './P2pPro
...
+    typeof (v as any).consumerAddress === 'string' &&
+    typeof (v as any).nonce === 'string' &&
+    typeof (v as any).signature === 'string'
+  )
+}
+
+export function isP2pUri(node: OceanNode): boolean {
+  if (!node) return false
+  if (typeof node === 'string') {
+    // Accept either peerId or multiaddr string as P2P
```
giurgiur99: This try catch is repeated three times, can we extract it in a helper?

src/services/providers/HttpProvider.ts
```diff
@@ -75,6 +93,29 @@ export class HttpProvider {
...
   }
 
+  private async getPersistentStorageSignaturePayload(
+    nodeUri: string,
+    signerOrAuthToken: SignerOrAuthTokenOrSignature,
+    command: string,
+    signal?: AbortSignal,
+    providerEndpoints?: any,
+    serviceEndpoints?: any
+  ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
```
giurgiur99: we have a type for this

src/services/providers/HttpProvider.ts
```diff
@@ -27,39 +27,57 @@ import {
...
-    if (typeof signerOrAuthToken === 'string') {
-      return {}
+    signal?: AbortSignal,
+    providerEndpoints?: any,
+    serviceEndpoints?: any
+  ): Promise<{ consumerAddress: string; nonce: string; signature: string | null }> {
+    if (!providerEndpoints) providerEndpoints = await this.getEndpoints(nodeUri)
+    if (!serviceEndpoints)
+      serviceEndpoints = await this.getServiceEndpoints(nodeUri, providerEndpoints)
+    if (isAgentSignature(signerOrAuthToken)) {
```
giurgiur99, line 59: The isAgentSignature return must come before the endpoint fetche otherwise every agent-signature call performs two avoidable network calls.

src/services/providers/HttpProvider.ts
```diff
@@ -75,6 +93,29 @@ export class HttpProvider {
...
+    nodeUri: string,
+    signerOrAuthToken: SignerOrAuthTokenOrSignature,
+    command: string,
+    signal?: AbortSignal,
+    providerEndpoints?: any,
+    serviceEndpoints?: any
+  ): Promise<{ consumerAddress: string; nonce: string; signature: string }> {
+    if (typeof signerOrAuthToken === 'string') {
+      throw new Error(
+        'Persistent storage operations require a Signer or AgentSignature (nonce/signature).'
```
giurgiur99: why throw error if jwt? we can do buckets operations with authtokens

src/services/providers/HttpProvider.ts
alexcos20: fixed

src/services/providers/HttpProvider.ts
alexcos20: fixed

src/services/providers/HttpProvider.ts
alexcos20: fixed, all bucket operations are using getSignedCommandParams (both P2P and HTTP)

alexcos20 (commented):

Overall risk: high

Summary:
This PR introduces support for `CompleteSignature` and updates provider interfaces to use `OceanNode` and `SignerOrAuthTokenOrSignature`, modernizing the API to handle multiple authentication mechanisms. The architectural direction and P2P connection fallback mechanisms are excellent. However, there are critical bugs introduced in `getSignedCommandParams` that result in `consumerAddress` evaluating to `undefined` when using Auth Tokens, which will break request payloads. Additionally, there are missing signature propagation fixes in `validateDdo` and a potentially missing update for `validateDdo` in `HttpProvider`.

Comments:
• [ERROR][bug] When `signerOrAuthToken` is an auth token string, this block returns `{ consumerAddress: undefined, ... }`. This causes `consumerAddress` to be `undefined` in request payloads that rely on it (e.g., `computeStart`). You should decode the address from the JWT to preserve existing functionality. Also, updating the return type to `Promise<Partial<CompleteSignature>>` satisfies strict typing.

```diff
-    if (typeof signerOrAuthToken === 'string') {
-      return {
-        consumerAddress: undefined,
-        nonce: undefined,
-        signature: undefined
-      }
-    }
+    if (typeof signerOrAuthToken === 'string') {
+      return {
+        consumerAddress: await getConsumerAddress(signerOrAuthToken),
+        nonce: undefined,
+        signature: undefined
+      } as Partial<CompleteSignature>
+    }
```
• [ERROR][bug] Similar to `HttpProvider.ts`, when `signerOrAuthToken` is an auth token, `consumerAddress` is incorrectly returned as `undefined`. This breaks requests relying on the consumer address.

```diff
-    if (typeof signerOrAuthToken === 'string') {
-      return {
-        consumerAddress: undefined,
-        nonce: undefined,
-        signature: undefined
-      } as any
-    }
+    if (typeof signerOrAuthToken === 'string') {
+      return {
+        consumerAddress: await getConsumerAddress(signerOrAuthToken),
+        nonce: undefined,
+        signature: undefined
+      } as Partial<CompleteSignature>
+    }
```
• [ERROR][bug] When validating a DDO with an auth token, passing `null` for `signerOrAuthToken` prevents the `Authorization` header from being appended to the P2P message. Pass `signerOrAuthToken` so `sendP2pCommand` can extract and send the auth token properly.

```diff
-      { ddo, publisherAddress, nonce, signature },
-      null,
+      { ddo, publisherAddress, nonce, signature },
+      signerOrAuthToken,
       signal
```
• [ERROR][bug] `validateDdo` was updated in `BaseProvider` to accept `SignerOrAuthTokenOrSignature`, but `HttpProvider`'s implementation of `validateDdo` is missing from this PR. If `HttpProvider` still strictly expects a `Signer` object, passing a string auth token will cause runtime errors when it attempts to call `.getAddress()`. Ensure `HttpProvider.validateDdo` is also updated to handle the new type.
• [WARNING][bug] If `signerOrAuthToken` is a string, `consumerAddress` is entirely omitted from the payload, which may cause the node to reject the request. Refactoring this block to always include `consumerAddress` restores the old behaviour.

```diff
-    const payload: Record<string, any> = {
-      command: PROTOCOL_COMMANDS.COMPUTE_GET_RESULT,
-      jobId,
-      index,
-      offset
-    }
-
-    if (typeof signerOrAuthToken === 'string') {
-      payload.authorization = signerOrAuthToken
-    } else {
-      const { consumerAddress, nonce, signature } = await this.getSignedCommandParams(
-        nodeUri,
-        signerOrAuthToken,
-        PROTOCOL_COMMANDS.COMPUTE_GET_RESULT
-      )
-      payload.nonce = nonce
-      payload.signature = signature
-      payload.consumerAddress = consumerAddress
-    }
+    const { consumerAddress, nonce, signature } = await this.getSignedCommandParams(
+      nodeUri,
+      signerOrAuthToken,
+      PROTOCOL_COMMANDS.COMPUTE_GET_RESULT
+    )
+    const payload: Record<string, any> = {
+      command: PROTOCOL_COMMANDS.COMPUTE_GET_RESULT,
+      jobId,
+      index,
+      offset,
+      consumerAddress
+    }
+    if (typeof signerOrAuthToken === 'string') {
+      payload.authorization = signerOrAuthToken
+    } else {
+      payload.nonce = nonce
+      payload.signature = signature
+    }
```
• [WARNING][bug] If `signerOrAuthToken` is passed as a raw hex signature (`0x...`) without being wrapped in `CompleteSignature`, `typeof signerOrAuthToken === 'string'` is true, and it will be passed to `decodeJwt(signerOrAuthToken)`. This will throw an exception since a hex signature is not a valid JWT. Consider explicitly throwing an error for hex strings here or forcing raw signatures to be passed via the `CompleteSignature` type.
• [INFO][performance] Using `typeof (node as any).toString === 'function'` is true for almost every JavaScript object since they inherit from `Object.prototype`. This means `peerIdFromString` will unnecessarily be called (and catch exceptions) for standard generic objects. Consider checking for specific properties (e.g. `node.type === 'Ed25519'`) or using libp2p's `isPeerId` utility.
• [INFO][style] The public method `fetchNodeLogs` has been removed. Make sure this API contract change is intentional and well documented, as it is a breaking change for any downstream dependencies relying on this pre-signed log fetching.
• [INFO][performance] Excellent implementation of connection fallback logic. Prioritizing direct transports and correctly filtering out and falling back to relay (`/p2p-circuit`) addresses avoids unnecessary latency overhead when a direct route is available.

src/services/providers/BaseProvider.ts
alexcos20: fixed

alexcos20 (commented):

Overall risk: medium

Summary:
This PR introduces an excellent refactoring to abstract authentication mechanisms via `SignerOrAuthTokenOrSignature` and centralizes signature logic via `getSignedCommandParams`. It also improves `P2pProvider` by adding better connection resolution and graceful fallback to P2P-circuit relays. However, there are a few bugs related to URL query parameter construction that will serialize literal `"undefined"` strings when AuthTokens are used, and a crash vulnerability in `getConsumerAddress` when raw hex signatures are provided.

Comments:
• [INFO][style] A great refactoring! Centralizing the nonce and signature generation logic significantly reduces code duplication and potential bugs across the provider endpoints.
• [ERROR][bug] You explicitly allow raw `0x...` signature strings here, but `getConsumerAddress` unconditionally passes any string to `decodeJwt()`. If a caller passes a raw signature string, `getConsumerAddress` will crash trying to decode it as a JWT. Since `CompleteSignature` provides a proper way to pass raw signatures along with their consumer address, you should drop the `0x` check here to avoid this crash.
```diff
   if (typeof signerOrAuthToken === 'string') {
-    // it's either a signature already (0x..) or a jwt token
-    if (signerOrAuthToken.startsWith('0x')) {
-      return signerOrAuthToken
-    } else return null
+    return null
   }
```
• [WARNING][bug] Because `getSignedCommandParams` explicitly returns `{ nonce: undefined, signature: undefined }` for AuthTokens, methods that spread this object into `URLSearchParams` (e.g. `createPersistentStorageBucket`) will inadvertently serialize literal `"undefined"` strings (`nonce=undefined&signature=undefined`). Omitting the undefined keys entirely prevents this issue.
```diff
     if (typeof signerOrAuthToken === 'string') {
-      return {
-        consumerAddress: await getConsumerAddress(signerOrAuthToken),
-        nonce: undefined,
-        signature: undefined
-      }
+      return { consumerAddress: await getConsumerAddress(signerOrAuthToken) } as CompleteSignature
     }
```
• [WARNING][bug] When using an AuthToken, `nonce` and `signature` will be undefined. Concatenating them directly results in the literal strings `"undefined"` being sent as query parameters, which will cause validation to fail on the provider backend. You should append them conditionally.
```diff
     let consumeUrl = downloadUrl
     consumeUrl += `?fileIndex=${fileIndex}`
     consumeUrl += `&documentId=${documentId}`
     consumeUrl += `&transferTxId=${transferTxId}`
     consumeUrl += `&consumerAddress=${consumerAddress}`
-    consumeUrl += `&nonce=${nonce}`
-    consumeUrl += `&signature=${signature}`
+    if (nonce) consumeUrl += `&nonce=${nonce}`
+    if (signature) consumeUrl += `&signature=${signature}`
```
• [WARNING][bug] Similar to the `download` method, `URLSearchParams.set` will convert `undefined` variables into literal `"undefined"` strings in the query parameters. Only set these values if they are defined.
```diff
     const queryParams = new URLSearchParams()
     queryParams.set('consumerAddress', consumerAddress)
     queryParams.set('jobId', jobId)
     if (agreementId) queryParams.set('agreementId', agreementId)
-    queryParams.set('nonce', nonce)
-    queryParams.set('signature', signature)
+    if (nonce) queryParams.set('nonce', nonce)
+    if (signature) queryParams.set('signature', signature)
```
• [WARNING][bug] Avoid appending `"undefined"` to the result URL when `nonce` and `signature` are omitted due to AuthToken usage.
```diff
     if (!computeResultUrl) return null
     let resultUrl = computeResultUrl
-    resultUrl += `?jobId=${jobId}&index=${index}&consumerAddress=${consumerAddress}&nonce=${nonce}&signature=${signature}`
+    resultUrl += `?jobId=${jobId}&index=${index}&consumerAddress=${consumerAddress}`
+    if (nonce) resultUrl += `&nonce=${nonce}`
+    if (signature) resultUrl += `&signature=${signature}`
     return resultUrl
```
• [INFO][other] The fallback logic handling `p2p-circuit` multiaddrs is very cleanly implemented. Filtering out relays for the initial direct connection attempt and recursively falling back to relayed paths ensures robust connectivity.

alexcos20 (commented):

Overall risk: medium

Summary:
This PR introduces a robust refactor to node interactions, unifying parameter gathering via `getSignedCommandParams` and adding standard connection interfaces (`OceanNode`). The P2P network dial fallback logic (direct to relayed connections) is particularly well-designed and will substantially increase connection reliability. Furthermore, conditionally omitting the `Authorization` header prevents unexpected proxy errors. However, there is a critical logic error in URL construction inside `HttpProvider.encrypt` that will cause malformed requests, alongside minor type-safety concerns regarding the `CompleteSignature` assertions.

Comments:
• [ERROR][bug] String concatenating the `?chainId=...` parameter directly to a potentially `null` endpoint URL results in the literal string `"null?chainId=..."`. Since this evaluates as truthy, the `if (!path)` check on the next line is bypassed, leading to failed network requests. Resolve the base path before appending query params.

```diff
-    let path =
-      (this.getEndpointURL(serviceEndpoints, 'encrypt')
-        ? this.getEndpointURL(serviceEndpoints, 'encrypt').urlPath
-        : null) + `?chainId=${chainId}`
+    let path = this.getEndpointURL(serviceEndpoints, 'encrypt')?.urlPath
     if (!path) return null
+    path += `?chainId=${chainId}`
```
• [WARNING][style] Casting this partial object to `CompleteSignature` forces the TypeScript compiler to assume `nonce` and `signature` are populated strings, which they are not in this block. This weakens type safety and can cause hidden runtime bugs if upstream functions expect string methods. Consider changing the return type of `getSignedCommandParams` across the providers to `Promise<Partial<CompleteSignature> & { consumerAddress: string }>` instead.

```diff
-      return {
-        consumerAddress: await getConsumerAddress(signerOrAuthToken)
-      } as CompleteSignature
+      return {
+        consumerAddress: await getConsumerAddress(signerOrAuthToken)
+      } as Partial<CompleteSignature> & { consumerAddress: string }
```
• [WARNING][style] The `isAuthToken` variable was cleaned up elsewhere (e.g. `computeStop`) but was left behind here. While the logic technically holds since a `CompleteSignature` object sets `isAuthToken` to `false`, it's safer and more consistent with your refactoring to check for the parameters directly:

```diff
-    // Is signer, add signature and nonce
-    if (!isAuthToken) {
+    if (signature) {
       url += `&signature=${signature}`
+    }
+    if (nonce) {
       url += `&nonce=${nonce}`
     }
```
• [INFO][other] An empty catch block will silently swallow all errors (including `AbortError` if the signal is aborted, or unexpected network failures). It is highly recommended to at least log the error to debug outputs so network resolution issues aren't completely hidden.

```diff
-    } catch (err) {}
+    } catch (err: any) {
+      LoggerInstance.debug(`[P2P] findProviders failed: ${err.message}`)
+    }
```

giurgiur99 (approved):

PR #2102, opened 2026-06-25
Call started endpoint

dnsi0:
Fixes # .

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when starting compute jobs by sending “job started” notifications without blocking the job start.
  * Added safeguards so notification failures (or missing status info) don’t prevent successful job start.
  * Improved notification accuracy by using the correct started job when multiple results are returned, with safer error handling and clearer logging.

### DIFF
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -378,12 +378,10 @@ export class BaseProvider {
   /**
    * Resolves the node's peerId for the given nodeUri via the node STATUS command,
    * whose `id` field is the peerId.
-   * @param {string | Multiaddr[]} nodeUri The provider URI.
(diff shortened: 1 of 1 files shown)


src/services/providers/BaseProvider.ts
```diff
@@ -323,6 +325,13 @@ export class BaseProvider {
       dockerRegistryAuth,
       outputBucketId
     )
+    const job = Array.isArray(jobs) ? jobs[0] : jobs
+    try {
+      await this.notifyIncentiveBackendJobStarted(nodeUri, computeEnv, job)
```
giurgiur99: Let's do here a non-blocking fire and forget for faster responses. Just call without await and catch potential errors .catch()...

giurgiur99 (approved):

PR #2103, opened 2026-06-25
Migrate HTTP client from cross-fetch to native fetch

alexcos20:
# Migrate HTTP client from `cross-fetch` to native `fetch`

## What

Replaces `cross-fetch` (which bundles **node-fetch v2**) with the runtime's **native global
`fetch`** (undici in Node, the built-in in browsers) across the HTTP client.

## Why

On modern Node (CI runs Node 20 & 22), node-fetch v2 intermittently throws
`Premature close` while reading a response body. It surfaced as flaky integration failures on
the **http** matrix leg — a *different* endpoint each run (`getFileInfo`, `getComputeResult`, …) —
while the **p2p** leg passed, proving the data was fine and the fault was in the HTTP stack, not
the node. Node's native `fetch` (≥18) doesn't have this bug, and the library already targets
Node ≥20 and browsers where `fetch` is universal (`src/utils/FetchHelper.ts` already used the
global `fetch`), so this just makes the rest of the client consistent.

## Changes

- **Drop the `cross-fetch` import** in `src/services/providers/HttpProvider.ts` and
  `src/services/Aquarius.ts` — all calls now use the global `fetch`. The ~34 non-streaming calls
  (`.json()` / `.text()` / `.arrayBuffer()`) are unchanged.
- **Streaming responses:** native `fetch` exposes `response.body` as a WHATWG `ReadableStream`
  (node-fetch returned a Node `Readable`). Added `responseBodyToAsyncIterable()` in
  `src/utils/bytes.ts` — a small, portable wrapper (works in Node 20+ and browsers, which don't
  yet support `for await` on `ReadableStream` directly), with `reader.cancel()` on early exit so
  an unconsumed stream doesn't leak the socket — and used it in `getComputeResult` and
  `computeStreamableLogs`. This also makes `Premature close` impossible (undici ends streams
  cleanly). `downloadNodeLogs` now parses the response with `.json()` to return a real
  `NodeLogEntry[]` (matching the P2P path and the declared return type) — the old code returned
  the raw body stream, a latent contract bug.
- **Streaming upload** (`uploadPersistentStorageFile`): unified the Node/browser branches into a
  single WHATWG `ReadableStream` request body with `duplex: 'half'` (required by undici for
  streamed bodies); removed the `Readable.from` / `import('stream')` Node path.
- **`package.json`:** removed `cross-fetch`; added `engines.node: ">=18"`. Kept `node-fetch` as an
  explicit **devDependency** — one integration test (`PublishEditConsume`) uses it together with
  the `form-data` package for a multipart upload, which doesn't port cleanly to undici; that test
  is left as-is.

## Scope

Standalone change off `main` — independent of the Service-on-Demand work (separate PR). No public
API changes; the streaming methods keep returning `AsyncIterable<Uint8Array>` and callers already
iterate with `for await`.

## Verification

- `npm run build` (cjs/esm/umd), `npm run type-check`, and `eslint` all pass.
- Runtime behaviour is exercised by the CI integration matrix — the **http** leg in particular,
  on the endpoints that previously flaked with `Premature close` (`fileInfo`, `getComputeResult`,
  streamable logs, persistent-storage upload).
- `grep -rn "cross-fetch" src/ test/` returns nothing; a fresh `npm install` drops `cross-fetch`
  and installs `node-fetch` for the one test (lockfile not touched in this PR).

## Summary by CodeRabbit

* **New Features**
  * Added a utility to normalize streamed response bodies into a consistent async-iterable format for easier processing.

* **Bug Fixes**
  * Improved handling of streamed compute results and log downloads, returning completed log responses as structured entries.
  * Strengthened persistent storage uploads with safer chunk encoding, better stream construction, and more reliable cleanup on failure.

* **Chores**
  * Reduced reliance on bundled fetch polyfills by using native `fetch` where available and requiring Node.js 18+.

* **Tests**
  * Updated IPFS integration tests to use native `fetch` and `FormData` behavior.

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,6 @@
     "decimal.js": "^10.4.1",
     "eciesjs": "^0.4.5",
     "ethers": "^6.15.0",
-    "form-data": "^2.3.3",
     "jsonwebtoken": "^9.0.2"
   },
   "devDependencies": {
@@ -100,8 +99,6 @@
     "@types/jsonwebtoken": "^9.0.10",
     "@types/mocha": "^10.0.10",
     "@types/node": "^24.3.0",
-    "@types/node-fetch": "^3.0.3",
-    "node-fetch": "^2.7.0",
     "@typescript-eslint/eslint-plugin": "^7.0.0",
     "@typescript-eslint/parser": "^6.4.1",
     "auto-changelog": "^2.4.0",
--- a/package.json
+++ b/package.json
@@ -64,6 +64,9 @@
     "url": "https://github.com/oceanprotocol/ocean.js/issues"
   },
   "homepage": "https://github.com/oceanprotocol/ocean.js#readme",
+  "engines": {
+    "node": ">=18"
+  },
   "peerDependencies": {
     "web3": "^1.8.0"
   },
@@ -71,7 +74,6 @@
     "@oasisprotocol/sapphire-paratime": "^1.3.2",
     "@oceanprotocol/ddo-js": "^0.3.0",
     "bignumber.js": "^9.3.1",
-    "cross-fetch": "^4.0.0",
     "crypto-js": "^4.1.1",
     "decimal.js": "^10.4.1",
     "eciesjs": "^0.4.5",
@@ -99,6 +101,7 @@
     "@types/mocha": "^10.0.10",
     "@types/node": "^24.3.0",
     "@types/node-fetch": "^3.0.3",
+    "node-fetch": "^2.7.0",
     "@typescript-eslint/eslint-plugin": "^7.0.0",
     "@typescript-eslint/parser": "^6.4.1",
     "auto-changelog": "^2.4.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,6 @@
         "decimal.js": "^10.4.1",
         "eciesjs": "^0.4.5",
         "ethers": "^6.15.0",
-        "form-data": "^2.3.3",
         "jsonwebtoken": "^9.0.2"
       },
       "devDependencies": {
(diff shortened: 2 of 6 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR successfully migrates the codebase from `cross-fetch` to native `fetch`, minimizing dependencies and modernizing the code to standard Web APIs. It cleanly unifies Node and browser streaming logic using WHATWG `ReadableStream` and introduces a robust async iteration polyfill. There is one typing concern regarding a forced cast to an array that should be reviewed, but overall the architectural modernization is excellent. LGTM!

Comments:
• [WARNING][bug] Casting an `AsyncIterable<Uint8Array>` to `NodeLogEntry[]` is highly suspicious. A stream of raw bytes does not possess array methods (like `.map` or `.length`) and does not contain parsed `NodeLogEntry` objects. If the caller expects an array of objects, this will cause a runtime crash.

If the caller expects a byte stream, the method's return type should be updated to reflect that instead of using `as unknown as NodeLogEntry[]`. If they actually expect an array, you need to buffer and parse the stream into an array before returning.
• [INFO][style] Excellent implementation of a cross-environment async iterator for `ReadableStream`. This ensures robust compatibility with environments (like older browsers) that lack `Symbol.asyncIterator` support natively on streaming responses.
• [INFO][performance] Unifying the request body streaming using standard `ReadableStream` and adding `duplex: 'half'` is a great modern approach. It ensures proper memory management during uploads and fulfills standard requirements for both Node 18+ (Undici) and Chrome.

package.json
```diff
@@ -99,6 +101,7 @@
     "@types/mocha": "^10.0.10",
     "@types/node": "^24.3.0",
     "@types/node-fetch": "^3.0.3",
+    "node-fetch": "^2.7.0",
```
giurgiur99: Do we still need this?

dnsi0 (approved):

package.json
alexcos20: it's needed in one of the tests, let me try to remove it

alexcos20: see https://github.com/oceanprotocol/ocean.js/pull/2103/commits/f5317e27932a38ccf871ba4625471face608ac77

giurgiur99 (approved):

PR #2111, opened 2026-07-07
sign issuer peer id

giurgiur99:
Fixes # .

Changes proposed in this PR:

- Add issuerPeerId in signed message for cross-node auth tokens
- 
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved auth token creation and invalidation so requests are signed with the current node identity, which should make token operations more reliable.
  * Updated node status handling to use live status data for public key and peer ID lookup, reducing failures when connecting to nodes.

* **Chores**
  * Adjusted CI to run against a specific node version for test jobs.


alexcos20 (commented):

Overall risk: low

Summary:
The PR successfully implements cross-node signature logic by incorporating `issuerPeerId` into the `CREATE_AUTH_TOKEN` signature flow. It also standardizes signature generation for `INVALIDATE_AUTH_TOKEN` using `getSignature` and refactors P2P endpoints to use the more robust `getNodeStatus` method. Please ensure that the hardcoded `NODE_VERSION` PR branch in the CI workflow is changed to a stable branch before merging.

Comments:
• [WARNING][other] Hardcoding `NODE_VERSION=pr-1404` is useful for end-to-end testing of this PR, but it should be reverted to a stable tag (e.g., `main` or latest release) before merging. Merging this as-is will cause the CI to break on `main` once the upstream PR is merged and its branch is deleted. (Note: Same applies to line 153).
• [INFO][style] The outer `String(...)` wrapping the concatenated variables is slightly redundant since the `+` operator already ensures the result is a string. You can simplify this to:
```typescript
  const message = 
    String(await signerOrAuthToken.getAddress()) +
    String(nonce) +
    String(command) +
    String(issuerPeerId)
```
• [INFO][other] Great DRY refactoring here! Consolidating the signing logic into the unified `getSignature` method standardizes how messages are built and signed across different commands.
• [WARNING][bug] Just to verify: The previous code returned `endpoints?.nodePublicKey`, and now it returns `status?.publicKey`. Please double-check that the `NodeStatus` interface provides a `publicKey` field and not `nodePublicKey`, otherwise this function will quietly return `undefined`.

alexcos20 (commented):

Overall risk: medium

Summary:
This PR implements cross-node signatures by incorporating the `issuerPeerId` into the signature payload for authentication tokens. It standardizes signature generation using protocol commands, which effectively mitigates replay attacks across different operations. Please ensure the temporary CI branch hardcoding is removed prior to merging to prevent future pipeline breakages.

Comments:
• [ERROR][bug] Hardcoding `NODE_VERSION=pr-1404` ties the CI to an ephemeral PR branch from another repository. Please revert this to target the `main` or latest stable version before merging, as it will break the CI once the referenced PR is merged or deleted. This also applies to line 153.

```diff
-          export NODE_VERSION=pr-1404 && bash -x start_ocean.sh 2>&1 > start_ocean.log &
+          bash -x start_ocean.sh 2>&1 > start_ocean.log &
```
• [INFO][security] Concatenating fields directly for a signature (`address + nonce + command + issuerPeerId`) can theoretically lead to collision vulnerabilities if the boundaries between variable-length fields are ambiguous. While predefined protocol commands mitigate this risk here, consider transitioning to EIP-712 structured data for signatures in the future to ensure robust, collision-free signing.
• [WARNING][performance] Fetching the node status via `this.getNodeStatus(nodeUri, signal)` introduces an additional network roundtrip during `createAuthToken`. If `issuerPeerId` is static or changes infrequently for a given `nodeUri`, consider caching this property at the provider instance level to optimize performance.
• [INFO][security] Excellent refactor here. By consistently using `getSignature` and explicitly including `PROTOCOL_COMMANDS.INVALIDATE_AUTH_TOKEN` in the payload, you actively mitigate replay attacks where a signature could otherwise be reused across different commands.

giurgiur99:
@coderabbitai review

PR #2115, opened 2026-07-17
add service list

### LINKED ISSUE #2116: Add node's  service_list handler

alexcos20:
# Add `SERVICE_LIST` support: `ProviderInstance.getServices()`

## What

Client-side support for the new ocean-node `SERVICE_LIST` command
(`GET /api/services/serviceList`, ocean-node commit
[`d6153a0`](https://github.com/oceanprotocol/ocean-node/commit/d6153a0a735cd83991e8dc53cb958d9e6b20f71a)):
the **node-wide service listing** for Service on Demand, exposed as
`ProviderInstance.getServices(nodeUri, signerOrAuthToken, filters?, signal?)` on both the
HTTP and P2P transports.

- **Authenticated but NOT owner-scoped** — any consumer identity (signature or auth
  token) sees every owner's services. The per-owner full view remains
  `getServiceStatus` (`SERVICE_GET_STATUS`).
- **Default (no filters)**: only the services **currently holding a resource
  reservation** — exactly what the engines count against the shared pools:
  Running/Restarting/Stopping, the mid-start pipeline states, paid `Error` (container
  died, restartable), and explicitly `Stopped` within the paid window. `Expired` and
  never-paid jobs hold nothing and are not listed.
- **Filters** (`ServiceListFilters`):
  - `status` — ONE specific `ServiceStatusNumber` (any, incl. `Expired`); takes
    precedence over `includeAllStatuses`
  - `includeAllStatuses` — every status, regardless of reservations
  - `fromTimestamp` — only services created at/after that moment (ISO date string, or a
    Unix timestamp in seconds or milliseconds)
- **Listing-sanitized output** (`ServiceJobListed`): on top of the always-stripped
  `userData`, the node removes `dockerCmd`, `dockerEntrypoint`, `dockerfile` and
  `additionalDockerFiles`; identity, status, resources, endpoints and payment metadata
  are kept.

## Changes

- `src/@types/Provider.ts` — `PROTOCOL_COMMANDS.SERVICE_LIST = 'serviceList'` (the
  auth-signed `command` string and P2P dispatch key; matches ocean-node verbatim).
- `src/@types/Services.ts` — new exported types `ServiceJobListed`
  (`Omit<ServiceJob, 'dockerCmd' | 'dockerEntrypoint' | 'dockerfile' | 'additionalDockerFiles'>`)
  and `ServiceListFilters`.
- `src/services/providers/HttpProvider.ts` — `getServices()`: signed
  `GET /api/services/serviceList` following the `getServiceStatus` pattern (advertised
  endpoint discovery with well-known-path fallback, signature query params or
  auth-token `Authorization` header). Filters are serialized the way the node's route
  parses them (`status` stringified, `includeAllStatuses` sent as literal `'true'` only
  when set, `fromTimestamp` passed through).
- `src/services/providers/P2pProvider.ts` — `getServices()`: the `serviceList` P2P
  command with the filters inline (native number/boolean).
- `src/services/providers/BaseProvider.ts` — façade dispatch, so the method works
  transparently over HTTP or P2P like the rest of the service surface.

## Tests

- **Unit** (`test/unit/Services.test.ts`): asserts the `serviceList` command string
  matches the node value exactly, and that `ProviderInstance.getServices` is exposed.
- **Integration** (`test/integration/Services.test.ts`): new lifecycle step after
  `getServiceStatus` exercising, against a live node: the default listing returned to a
  **different identity than the owner** (proving it is not owner-scoped), sanitization
  of all five stripped fields, the `status` filter (`Running` matches, `Expired` does
  not), and a future `fromTimestamp` excluding the service. Skips cleanly when the
  Service-on-Demand stack isn't available, like the neighboring tests.
- **Drive-by fix**: the restart test was passing `opSignal()` into the `dockerCmd`
  parameter slot (latent since dockerCmd/dockerEntrypoint were added to
  `serviceRestart` in #2114); it now lands in `signal`.

## Compatibility

`getServices()` requires an ocean-node build that includes the `SERVICE_LIST` handler —
older nodes will 404 the HTTP route / reject the P2P command. No changes to existing
methods or types; purely additive.

## Verification

- `npm run type-check` passes; ESLint clean on all touched files (only pre-existing
  `prefer-destructuring` warnings remain in the integration test).
- `test/unit/Services.test.ts` passes (4/4).
- The integration step runs on the CI barge matrix (http + p2p legs) alongside the rest
  of the Service-on-Demand lifecycle.

## Summary by CodeRabbit

* **New Features**
  * Added the ability to list services across a node.
  * Supports filtering by status and start time, with an option to include all statuses.
  * Service listings protect sensitive configuration details while retaining relevant identity and payment information.
  * Available through both HTTP and peer-to-peer connections.

* **Tests**
  * Added coverage for service listing, filtering, response sanitization, and protocol integration.

### DIFF
--- a/src/@types/Provider.ts
+++ b/src/@types/Provider.ts
@@ -151,6 +151,7 @@ export const PROTOCOL_COMMANDS = {
   SERVICE_STOP: 'serviceStop',
   SERVICE_RESTART: 'serviceRestart',
   SERVICE_GET_STATUS: 'serviceGetStatus',
+  SERVICE_LIST: 'serviceList',
   SERVICE_EXTEND: 'serviceExtend',
   SERVICE_GET_STREAMABLE_LOGS: 'serviceGetStreamableLogs'
 }
--- a/src/@types/Services.ts
+++ b/src/@types/Services.ts
@@ -110,6 +110,32 @@ export interface ServiceJob {
   extendPayments?: ServiceJobPayment[] // one entry per successful SERVICE_EXTEND
 }
 
+// Returned by SERVICE_LIST, which is authenticated but NOT owner-scoped (any consumer
+// identity sees every owner's services). On top of the always-stripped userData, the
+// node removes everything that reveals HOW a service is configured — CMD/ENTRYPOINT
+// overrides and any inline Dockerfile. Identity, status, resources, endpoints and
+// payment metadata are kept; use the owner-scoped SERVICE_GET_STATUS for the full view.
+export type ServiceJobListed = Omit<
+  ServiceJob,
+  'dockerCmd' | 'dockerEntrypoint' | 'dockerfile' | 'additionalDockerFiles'
+>
+
+// Filters for SERVICE_LIST (getServices). With no filters the node returns only the
+// services currently holding a resource reservation — exactly what the engines count
+// against the shared pools: Running/Restarting/Stopping, the mid-start pipeline states,
+// paid Error (container died, restartable), and explicitly Stopped within the paid
+// window. Expired and never-paid jobs hold nothing and are not listed by default.
+export interface ServiceListFilters {
+  // filter to ONE specific status (any ServiceStatusNumber, incl. Expired); takes
+  // precedence over includeAllStatuses
+  status?: ServiceStatusNumber
+  // return services in EVERY status instead of only the resource-holding set
+  includeAllStatuses?: boolean
+  // only services created at/after this moment: ISO date string, or a Unix timestamp
+  // (seconds or milliseconds) as a string
+  fromTimestamp?: string
+}
+
 // ── Request shapes ─────────────────────────────────────────────────────
 
 export interface ServicePayment {
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -31,6 +31,8 @@ import {
   PersistentStorageObject,
   PersistentStorageUpdateBucketResponse,
   ServiceJob,
+  ServiceJobListed,
+  ServiceListFilters,
   ServiceTemplatePublic,
   ServiceStartParams,
   ServiceUserData,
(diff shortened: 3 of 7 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR successfully introduces the `SERVICE_LIST` command to `ocean.js`, allowing consumers to fetch an authenticated, non-owner-scoped list of services on demand. The architecture changes consistently follow the existing provider patterns (Base, Http, P2P). Query parameters and P2P payloads are securely and correctly mapped. LGTM!

Comments:
• [INFO][style] The JSDoc comment explicitly mentions that `userData` is 'always-stripped', and your integration tests verify that `userData` is `undefined` on the returned objects. To ensure TypeScript strongly enforces this runtime behavior, consider adding `userData` to the `Omit` list (assuming it is present on the base `ServiceJob` type).

```diff
 export type ServiceJobListed = Omit<
   ServiceJob,
-  'dockerCmd' | 'dockerEntrypoint' | 'dockerfile' | 'additionalDockerFiles'
+  'userData' | 'dockerCmd' | 'dockerEntrypoint' | 'dockerfile' | 'additionalDockerFiles'
 >
```
• [INFO][style] Good job properly handling the optional inclusion of filter properties to the query string while correctly casting values to strings where necessary.

alexcos20:
> The JSDoc comment explicitly mentions that `userData` is 'always-stripped', and your integration tests verify that `userData` is `undefined` on the returned objects. To ensure TypeScript strongly enforces this runtime behavior, consider adding `userData` to the `Omit` list (assuming it is present on the base `ServiceJob` type).
> 
> ```diff
>  export type ServiceJobListed = Omit<
>    ServiceJob,
> -  'dockerCmd' | 'dockerEntrypoint' | 'dockerfile' | 'additionalDockerFiles'
> +  'userData' | 'dockerCmd' | 'dockerEntrypoint' | 'dockerfile' | 'additionalDockerFiles'
>  >
> ```

userData is not present on the base ServiceJob type in ocean.js. The client-side mirror at Services.ts:86 already models the node's response shape, where userData is stripped from every service response (status, start, stop, extend, restart — not just the listing). That's why the interface comment says "As returned by the node (userData is always stripped from responses)" and why the integration tests have to use (job as any).userData to check it — the property doesn't exist on the type at all.

PR #2118, opened 2026-07-23
Feature/162 inference services es

giurgiur99:
Fixes # .

Changes proposed in this PR:

- Notify backend about service started
- 
- 

## Summary by CodeRabbit

* **New Features**
  * Added incremental service listing with an `updatedSince` filter to retrieve services created or changed after a specified time.
  * Service records now expose the timestamp of their latest status update.
  * Service starts now trigger notifications to the incentive backend.
  * Incremental filtering is supported across HTTP and peer-to-peer service providers.

### DIFF
--- a/src/@types/Services.ts
+++ b/src/@types/Services.ts
@@ -101,6 +101,7 @@ export interface ServiceJob {
   status: ServiceStatusNumber
   statusText: string
   dateCreated: string // ISO timestamp
+  updatedAt?: number // Unix ms; bumped on every status change (SERVICE_LIST updatedSince cursor)
   expiresAt: number // Unix ms timestamp
   duration: number // requested seconds
   exposedPorts: number[]
@@ -134,6 +135,9 @@ export interface ServiceListFilters {
   // only services created at/after this moment: ISO date string, or a Unix timestamp
   // (seconds or milliseconds) as a string
   fromTimestamp?: string
+  // only services updated (created OR any status change) at/after this moment, same
+  // formats as fromTimestamp; returns every status. The incremental-sync cursor.
+  updatedSince?: string
 }
 
 // ── Request shapes ─────────────────────────────────────────────────────
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -446,28 +446,31 @@ export class BaseProvider {
       const baseUrl = incentiveBackendUrl.replace(/\/+$/, '')
       const peerId = await this.resolveNodePeerId(nodeUri)
 
-      await fetch(`${baseUrl}/services/${encodeURIComponent(service.serviceId)}/started`, {
-        method: 'POST',
-        headers: { 'Content-Type': 'application/json' },
-        body: JSON.stringify({
-          peerId,
-          clusterHash: service.clusterHash,
-          owner: service.owner,
-          status: service.status,
-          statusText: service.statusText,
-          environment: service.environment,
-          image: service.image,
-          tag: service.tag,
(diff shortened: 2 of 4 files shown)


giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: medium

Summary:
The PR successfully introduces incremental synchronization capabilities by adding `updatedAt` and `updatedSince` filters to the Service interfaces and Http/P2p providers. The filtering implementation is clean and follows existing patterns. However, there is a logical flaw in the new incentive backend notification where only the first service of a potentially larger array is processed. Once the batch iteration is fixed, LGTM!

Comments:
• [WARNING][bug] If `serviceStart` starts and returns multiple services (as the `Promise<ServiceJob[]>` signature suggests), only the first service `services[0]` will be sent to the incentive backend. You should iterate over the array and notify the backend for all newly started services.

```diff
-    const service = Array.isArray(services) ? services[0] : services
-    this.notifyIncentiveBackendServiceStarted(nodeUri, service).catch(() => {})
+    const startedServices = Array.isArray(services) ? services : [services]
+    startedServices.forEach(service => {
+      this.notifyIncentiveBackendServiceStarted(nodeUri, service).catch(() => {})
+    })
```
• [INFO][other] Relying directly on `process.env.INCENTIVE_BACKEND_URL` inside a provider class can be brittle if `ocean.js` is used as a client-side library in browsers (depending heavily on bundler injection). Consider passing this URL via the provider's configuration object or constructor to maintain better architectural decoupling.
• [INFO][style] The `.catch(() => {})` on this line is redundant because `notifyIncentiveBackendServiceStarted` already wraps its entire asynchronous execution in a `try/catch` block and safely logs its own errors internally.
• [INFO][style] Good use of optional fields (`updatedAt` and `updatedSince`) to cleanly support incremental synchronization without breaking backwards compatibility. The implementations mapping these to the HTTP/P2P queries are excellent.

giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: low

Summary:
The PR successfully implements service update tracking and an automatic notification to an incentive backend upon service start. Interface and filter updates are clean and well-structured. However, there is an architectural concern regarding the use of `process.env` inside an SDK, which can cause compatibility issues in browser environments. Additionally, the fire-and-forget implementation for network requests could lead to dropped fetch calls if the main process exits prematurely.

Comments:
• [WARNING][architecture] Using `process.env.INCENTIVE_BACKEND_URL` directly within a library/SDK restricts this feature to environments that support `process.env` (e.g., Node.js, or bundlers with env shimming). For universal SDKs, it is an architectural best practice to accept backend URLs via a configuration object passed to the provider during initialization, rather than hardcoding environment variable reads.
• [INFO][style] Good use of the regex `.replace(/\/+$/, '')` to sanitize the base URL. This effectively prevents double-slash malformations in the final request path. LGTM!
• [WARNING][performance] The loop `startedServices.forEach(...)` initiates asynchronous `fetch` calls without awaiting them. While the `.catch(() => {})` prevents unhandled promise rejection crashes, if the consuming application terminates immediately after `serviceStart` resolves, the background fetch requests may be abruptly aborted mid-flight. Consider whether these promises should be bundled into a `Promise.allSettled()` or tracked centrally.
• [INFO][bug] Ensure that the `fetch` API is available in your target Node.js environments. Native `fetch` is available in Node >= 18. If this SDK supports older Node versions, you may need to ensure a polyfill like `cross-fetch` is imported.

dnsi0 (approved):

PR #2119, opened 2026-07-24
allow dockerspecs on service restart

### LINKED ISSUE #2120: ServiceRestart should allow to change docker image
See https://github.com/oceanprotocol/ocean-node/pull/1429

alexcos20:
Closes #2120

# Allow Docker specs on `serviceRestart` (RESPEC mode)

## What

Client-side support for the extended `serviceRestart` command from ocean-node
[#1429](https://github.com/oceanprotocol/ocean-node/pull/1429) ("Allow docker specs on
service restart"): a restart can now change the container's **image specification** while
keeping the same serviceId, payment, resources, duration/expiry and host port(s). This
covers the realistic recovery workflow — start a service, hit a bug, publish a new tag,
then restart onto the new image without losing the paid window or reserved resources.

`ProviderInstance.serviceRestart` now takes a single optional `params: ServiceRestartParams`
object (replacing the previous positional `userData`/`dockerCmd`/`dockerEntrypoint` args),
consistent with how `serviceStart` takes `ServiceStartParams`. Available on both the HTTP
and P2P transports.

The node treats the restart **atomically** ("all-or-nothing"), so the client just forwards
the fields the caller supplies and lets the node pick the mode:

- **REUSE mode** — no container-spec fields supplied → the container is bounced on its
  stored spec, unchanged (the previous behaviour).
- **RESPEC mode** — ANY of `image`/`tag`/`checksum`/`dockerfile`/`additionalDockerFiles`
  supplied → the container is rebuilt entirely from `params`. `image` becomes mandatory,
  exactly one of `tag`/`checksum`/`dockerfile` applies, and a `dockerfile` requires
  `allowImageBuild` on the environment (the node replies `403` otherwise). Sending a
  container param without `image`, or more than one image mode, is a `400` from the node.
- `userData`/`dockerCmd`/`dockerEntrypoint` keep their **replace-when-supplied** semantics:
  an omitted field reuses the node's stored value, an explicit value (including `[]`)
  REPLACES it. As before, `userData` is ECIES-encrypted to the node's key client-side.

Each field is only put on the wire when actually supplied, so the client never
accidentally forces the node out of REUSE mode.

## Changes

- `src/@types/Services.ts` — factored the container spec shared by start and restart into
  a new exported `ServiceContainerSpec` (`image?`, `tag?`, `checksum?`, `dockerfile?`,
  `additionalDockerFiles?`, `userData?`, `dockerCmd?`, `dockerEntrypoint?`).
  `ServiceRestartParams` is that spec (all fields optional; atomic REUSE/RESPEC contract
  documented inline); `ServiceStartParams` now `extends ServiceContainerSpec`, re-declaring
  `image` as required and adding `environment`/`exposedPorts`/`resources`/`duration`/
  `payment`. No structural change to `ServiceStartParams`.
- `src/services/providers/HttpProvider.ts` — `serviceRestart()` now accepts
  `(nodeUri, signerOrAuthToken, serviceId, params?, signal?)`; unpacks `params` and adds
  each image-spec field to the request body only when defined.
- `src/services/providers/P2pProvider.ts` — same signature/body change over the
  `serviceRestart` P2P command.
- `src/services/providers/BaseProvider.ts` — façade signature updated to the new `params`
  shape (dropped the now-unused `ServiceUserData` import).

## Tests

- **Unit** (`test/unit/Services.test.ts`): unchanged — the `serviceRestart` command string
  and method exposure assertions still hold.
- **Integration** (`test/integration/Services.test.ts`): the existing REUSE restart step
  is updated to the new `params` signature (passes `undefined` for `params`).

## Compatibility

- **Breaking (client API):** `serviceRestart`'s positional `userData`/`dockerCmd`/
  `dockerEntrypoint` arguments are replaced by a single `params: ServiceRestartParams`
  object. Callers on `9.0.0-next.*` that passed those positionally must move them into the
  object. REUSE-mode restarts (no container spec) behave exactly as before.
- **Node requirement:** RESPEC mode requires an ocean-node build that includes #1429;
  older nodes ignore/na the extra fields and restart in REUSE mode. Purely additive on the
  wire — new fields are only sent when supplied.

## Summary by CodeRabbit

* **New Features**
  * Added flexible service restart options, including container image, build, Docker command, entrypoint, and user data overrides.
  * Restart requests now support reusing existing container settings or rebuilding with explicitly provided specifications.
  * Standardized service start and restart configuration handling across providers.

* **Bug Fixes**
  * Ensured omitted restart options preserve existing service configuration.

* **Tests**
  * Updated service restart integration coverage for the streamlined request format.

### DIFF
--- a/src/@types/Services.ts
+++ b/src/@types/Services.ts
@@ -152,18 +152,35 @@ export interface ServicePayment {
 // so a plaintext secret can never be forwarded unencrypted.
 export type ServiceUserData = Record<string, unknown>
 
-export interface ServiceStartParams {
-  environment: string // required: the envId to run the service on
-  image: string // base image name (or build label when dockerfile is set)
+// The container specification shared by serviceStart and serviceRestart. Every field is
+// optional here; the consuming type decides which are required (serviceStart re-declares
+// `image` as mandatory). `userData` is always ECIES-encrypted to the node before sending.
+export interface ServiceContainerSpec {
+  image?: string // base image name (or build label when dockerfile is set)
   tag?: string // pull by name:tag
   checksum?: string // pull by digest: "sha256:<64 hex>"
   dockerfile?: string // build from inline Dockerfile; requires allowImageBuild on the env
   additionalDockerFiles?: Record<string, string>
+  userData?: ServiceUserData
   dockerCmd?: string[] // exec-form CMD override (no shell)
   dockerEntrypoint?: string[]
+}
+
+// Optional container-spec overrides for serviceRestart. The node treats the restart
+// atomically ("all-or-nothing"): providing ANY of image/tag/checksum/dockerfile/
+// additionalDockerFiles switches it into RESPEC mode — the container is rebuilt entirely
+// from these values (`image` becomes mandatory, and exactly one of tag/checksum/dockerfile
+// applies) instead of being bounced on the stored spec (REUSE mode). Passing none of them
+// reuses the stored container spec unchanged. `userData`/`dockerCmd`/`dockerEntrypoint`
+// keep their replace-when-supplied semantics: an omitted value reuses the stored one, an
+// explicit value (including `[]`) REPLACES it.
+export type ServiceRestartParams = ServiceContainerSpec
+
+export interface ServiceStartParams extends ServiceContainerSpec {
+  environment: string // required: the envId to run the service on
+  image: string // required for start (base image name, or build label when dockerfile is set)
   exposedPorts?: number[]
   resources?: ComputeResourceRequest[]
   duration: number // seconds; capped by serviceOnDemand.maxDurationSeconds
-  userData?: ServiceUserData
   payment: ServicePayment
 }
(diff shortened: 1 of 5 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR refactors the `serviceRestart` method to support updating Docker specifications during a service restart. It introduces `ServiceContainerSpec` to share fields between start and restart operations. The providers (`HttpProvider` and `P2pProvider`) are updated to handle the new `params` object and properly conditionally-spread properties into the payload. The implementation is clean, though it introduces a breaking API change for library consumers calling `serviceRestart` directly.

Comments:
• [INFO][style] Excellent refactoring here. Extracting a shared `ServiceContainerSpec` interface reduces duplication and clearly establishes the common container attributes used by both `start` and `restart` actions.
• [WARNING][other] Changing the method signature from positional arguments (`userData, dockerCmd, dockerEntrypoint`) to a `params` object constitutes a breaking change in the public API. Ensure this is highlighted in the changelog/release notes so that consumers of `ocean.js` can update their implementations accordingly.
• [INFO][style] Good use of the conditional spread syntax (`...(image !== undefined ? { image } : {})`). This guarantees that keys are entirely omitted from the payload when undefined, cleanly satisfying the node's strict 'omitted means reuse' semantics. LGTM!

dnsi0 (approved):

PR #2125, opened 2026-07-29
Feat/162 service restart notify

giurgiur99:
Fixes # .

Changes proposed in this PR:

- Update model on restart
- 
- 

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced service restart handling to notify the incentive backend when restart metadata (image/tag/docker command) is available.
  * Added safeguards to only send restart notifications when required configuration and restart details are present.
  * Preserved existing restart behavior and outcomes when no restart command/metadata is provided.


giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: high

Summary:
This PR introduces a notification mechanism for service restarts. However, it contains a critical `ReferenceError` where `dockerCmd` is accessed without being declared in the local scope, which will fatally crash the `serviceRestart` method. Furthermore, there is an unused parameter in the new method and risky usage of `process.env` that may crash browser-based environments consuming this library.

Comments:
• [ERROR][bug] Critical: `dockerCmd` is not defined in the scope of `serviceRestart`. This will throw a `ReferenceError` and break the function. Assuming `dockerCmd` is a property of the optional `params` object, you should access it via `params?.dockerCmd`.

```diff
-    if (dockerCmd !== undefined) {
-      this.notifyIncentiveBackendServiceRestarted(nodeUri, serviceId, dockerCmd).catch(() => {})
+    if (params?.dockerCmd !== undefined) {
+      this.notifyIncentiveBackendServiceRestarted(nodeUri, serviceId, params.dockerCmd)
```
*Note: The trailing `.catch(() => {})` can also be safely removed because `notifyIncentiveBackendServiceRestarted` wraps its logic in a try/catch and never rejects.*
• [WARNING][style] The parameter `nodeUri` is declared but never used inside the `notifyIncentiveBackendServiceRestarted` method. You can safely remove it to keep the method signature clean.

```diff
-  private async notifyIncentiveBackendServiceRestarted(
-    nodeUri: OceanNode,
-    serviceId: string,
-    dockerCmd: string[]
-  ): Promise<void> {
+  private async notifyIncentiveBackendServiceRestarted(
+    serviceId: string,
+    dockerCmd: string[]
+  ): Promise<void> {
```
*(Be sure to update the call site at line 899 to match the removed argument).*.
• [WARNING][bug] `ocean.js` is a library often consumed in both Node.js and browser environments. Accessing `process.env` directly can throw a fatal `ReferenceError` in browsers where the global `process` object is undefined (unless specifically polyfilled by the consumer's bundler). It is much safer to guard this check.

```diff
-      const incentiveBackendUrl = process.env.INCENTIVE_BACKEND_URL
+      const incentiveBackendUrl = typeof process !== 'undefined' ? process.env.INCENTIVE_BACKEND_URL : undefined
```

giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: low

Summary:
The PR introduces a helpful feature to notify the incentive backend when a service is restarted with new parameters. The changes are largely well-implemented, correctly preventing URL injection and gracefully handling missing configuration. However, there are a few architectural and performance improvements to consider. Reading from `process.env` directly in an SDK can cause issues in browser environments, the `fetch` call is unawaited and missing the `AbortSignal`, and there is a redundant `.catch()` statement.

Comments:
• [WARNING][architecture] Directly accessing `process.env` inside an SDK (which may be used in browsers or non-Node environments) can cause reference errors or require specific build-time polyfills. It's generally a better practice to accept such configuration values through the class constructor (e.g., via a `Config` object injected into `BaseProvider`) rather than hardcoding environment variable reads.
• [WARNING][performance] The `fetch` request does not use the `signal` provided to the parent `serviceRestart` method. This means if the user aborts the restart operation, this fetch will keep running. Additionally, without a timeout, a hanging endpoint could cause resource leaks. Consider passing the `AbortSignal` down to this method and adding it to the `fetch` options.

```diff
-  private async notifyIncentiveBackendServiceRestarted(
-    serviceId: string,
-    params: ServiceRestartParams
+  private async notifyIncentiveBackendServiceRestarted(
+    serviceId: string,
+    params: ServiceRestartParams,
+    signal?: AbortSignal
   ): Promise<void> {
```
And in the `fetch`:
```diff
       await fetch(`${baseUrl}/services/${encodeURIComponent(serviceId)}/restarted`, {
         method: 'POST',
+        signal,
         headers: { 'Content-Type': 'application/json' },
```
• [INFO][style] The `.catch(() => {})` here is redundant because the `notifyIncentiveBackendServiceRestarted` method already wraps its entire body in a `try/catch` and gracefully handles any errors without re-throwing. If this was added to satisfy a linter rule like `no-floating-promises`, consider using `void this.notifyIncentiveBackendServiceRestarted(...)` instead.

Additionally, because this call is 'fire-and-forget' (not `await`ed), if this code runs in a short-lived serverless environment (like AWS Lambda), the Node process might exit before the background `fetch` completes, causing the notification to be lost. If reliable delivery is required, you should `await` it.

giurgiur99:
@coderabbitai review

alexcos20 (commented):

Overall risk: medium

Summary:
The PR adds a notification mechanism to the incentive backend when a service is restarted. The logic is generally sound and correctly executes as a fire-and-forget background task. However, the direct use of `process.env` in a library that supports browser environments poses a significant risk of runtime crashes. A fallback check for the `process` object is needed, or ideally, the configuration should be injected.

Comments:
• [ERROR][bug] Accessing `process.env` directly can cause `ReferenceError: process is not defined` in browser environments. If `ocean.js` is intended to be run in browsers without a bundler that polyfills `process`, you must verify that `process` exists before accessing its properties. Alternatively, consider passing the `INCENTIVE_BACKEND_URL` through the provider's configuration or constructor rather than relying on environment variables inside a core library class.

```diff
-      const incentiveBackendUrl = process.env.INCENTIVE_BACKEND_URL
+      const incentiveBackendUrl = typeof process !== 'undefined' && process.env ? process.env.INCENTIVE_BACKEND_URL : undefined;
```
• [INFO][security] Good use of `encodeURIComponent` on `serviceId` here. This prevents potential path injection vulnerabilities in the outgoing fetch URL.
• [INFO][style] The `notifyIncentiveBackendServiceRestarted` method has an internal `try-catch` block that catches and logs any errors without re-throwing them, which means its returned Promise will never reject. Because of this, the appended `.catch(() => {})` here is redundant and can be safely removed.

```diff
-      this.notifyIncentiveBackendServiceRestarted(serviceId, params).catch(() => {})
+      this.notifyIncentiveBackendServiceRestarted(serviceId, params)
```

dnsi0 (approved):

PR #2132, opened 2026-08-14
add node compute job/service runTimeMetrics

alexcos20:
# Add `runtimeMetrics` typings + `includeMetrics` plumbing for compute/service status

## Summary

Mirrors [ocean-node PR #1436](https://github.com/oceanprotocol/ocean-node/pull/1436) ("Feature/docker stats") on the client side: adds the typed shape of the new Docker/NVML runtime-metrics snapshot the node can attach to compute-job and service-job status responses, and teaches `computeStatus()` / `getServiceStatus()` how to request it via the new `includeMetrics` flag. No runtime behavior changes for any existing caller that doesn't pass the new parameter.

## Background

Ocean-node PR #1436 adds best-effort, owner-only runtime metrics (CPU, memory, disk, network, block I/O, PIDs, exit info, optional NVIDIA GPU stats) for C2D compute jobs and Service-on-Demand containers, sampled on a fixed interval and exposed through the existing `COMPUTE_GET_STATUS` / `SERVICE_GET_STATUS` endpoints behind a new `includeMetrics` flag. This PR brings the corresponding client-side types and opt-in wiring into `@oceanprotocol/lib` so consumers (e.g. `ocean-cli`) can request and type-check the new data once a node running that feature is available.

### New types (`src/@types/Compute.ts`)

- `GpuVendor` — `'nvidia' | 'amd' | 'intel'`
- `GpuMetricsSnapshot` — per-GPU utilization/memory/temperature/power; `null` (not `0`) marks a metric the backend couldn't read
- `ContainerMetricsSnapshot` — the full snapshot: `containerState`, `cpu`, `memory`, `disk`, optional `network`, `blockIO`, `pids`, optional `gpu[]`. Deliberately omits the node's internal `prev` CPU-delta accumulator, which the node always strips before a snapshot reaches any client.
- `NodeComputeJob` gains an optional `runtimeMetrics?: ContainerMetricsSnapshot`.

### `src/@types/Services.ts`

- `ServiceJob` gains the same optional `runtimeMetrics?: ContainerMetricsSnapshot`.
- `ServiceJobListed`'s omitted-keys union now also excludes `'runtimeMetrics'`, matching the node's `toListedServiceJob` — `SERVICE_LIST` never returns metrics regardless of any flag.

### Provider methods — `getServiceStatus()` (`BaseProvider`, `HttpProvider`, `P2pProvider`)

- New trailing optional parameter: `includeMetrics?: boolean`.
- This command is already authenticated (owner-scoped), so metrics are included **by default**; pass `includeMetrics: false` to opt out. `HttpProvider` forwards it as a query param, `P2pProvider` forwards it in the signed command body.

### Provider methods — `computeStatus()` (`BaseProvider`, `HttpProvider`, `P2pProvider`)

- New trailing optional parameter: `includeMetrics?: boolean`.
- Return type widened from `ComputeJob | ComputeJob[]` to `NodeComputeJob | NodeComputeJob[]` — the type that actually matches this endpoint's wire payload (it already carries `environment`, `resources`, `payment`, etc.), so `runtimeMetrics` is visible on the value callers get back.
- `COMPUTE_GET_STATUS` supports unauthenticated calls, so unlike `getServiceStatus()`, metrics are opt-in here:
  - An **auth token** (`signerOrAuthToken` passed as a string) is presented to the node unconditionally, on every call, regardless of `includeMetrics` — this was already true before this change.
  - A **`Signer`**'s nonce + signature are only computed (an extra nonce round-trip) and sent when `includeMetrics === true`, to avoid paying that cost on every plain, unauthenticated status poll.
  - Net effect: `includeMetrics` omitted → node attaches metrics silently if the request already carries valid owner credentials (automatic for token callers, requires the flag for `Signer` callers) and behaves exactly as before otherwise; `includeMetrics: true` → required, node answers 400/401 if credentials don't verify; `includeMetrics: false` → never attached.

### CI (`.github/workflows/ci.yml`)

- Bumped the barge `NODE_VERSION` env from `pr-1408` to `pr-1436` so the integration suite runs against an ocean-node build that actually has this feature.

## Behavior / compatibility

- **Purely additive.** All new parameters are optional and appended after existing trailing parameters (`signal`), so no existing positional call site shifts argument meaning.
- **No change in network requests** for any caller that doesn't pass `includeMetrics` to `computeStatus()`/`getServiceStatus()` with a `Signer` — the extra nonce/signature round-trip only happens when `includeMetrics: true` is explicitly requested on `computeStatus()`.
- `computeStatus()`'s declared return type changes from `ComputeJob | ComputeJob[]` to `NodeComputeJob | NodeComputeJob[]`; since `NodeComputeJob extends ComputeJob`, this is a widening, not a narrowing, and is source-compatible with existing callers typed against `ComputeJob`.

## Summary by CodeRabbit

* **New Features**
  * Added optional runtime metrics for compute jobs and services, including resource, memory, disk, network, process, container, and GPU information.
  * Status requests can now explicitly include or exclude runtime metrics.
  * Detailed metrics are available only when requested, while standard status responses remain streamlined.
* **Bug Fixes**
  * Updated integration testing to use the latest Barge version.
* **Privacy**
  * Service listings continue to exclude runtime metrics and other configuration-sensitive details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,7 @@ jobs:
           bash -x start_ocean.sh 2>&1 > start-node.log &
         env:
           CONTRACTS_VERSION: '2.9.0'
-          NODE_VERSION: "pr-1408"
+          NODE_VERSION: "pr-1436"
       - name: Install deps & build
         run: npm ci && npm run build:metadata
 
--- a/src/@types/Compute.ts
+++ b/src/@types/Compute.ts
@@ -98,6 +98,8 @@ export type ComputeJobMetadata = {
   [key: string]: string | number | boolean
 }
 
+// GPU backend a device's metrics were sampled with. Only 'nvidia' (NVML) is emitted
+// today; 'amd' / 'intel' are reserved for when those backends exist.
 export type GpuVendor = 'nvidia' | 'amd' | 'intel'
 
 // Per-GPU runtime metrics. One entry per GPU resource the job/service holds. Only NVIDIA
--- a/src/@types/Compute.ts
+++ b/src/@types/Compute.ts
@@ -98,6 +98,63 @@ export type ComputeJobMetadata = {
   [key: string]: string | number | boolean
 }
 
+export type GpuVendor = 'nvidia' | 'amd' | 'intel'
+
+// Per-GPU runtime metrics. One entry per GPU resource the job/service holds. Only NVIDIA
+// is emitted today; `null` (not `0`) means the backend could not read that metric.
+export interface GpuMetricsSnapshot {
+  resourceId: string
+  vendor: GpuVendor
+  utilizationPercent: number | null
+  memoryUsedBytes: number | null
+  memoryTotalBytes: number | null
+  temperatureC?: number
+  powerWatts?: number
+  shared?: boolean // true → device-level number, may include other jobs' load
+}
+
+// Best-effort Docker/NVML runtime metrics snapshot for a compute job or service container,
+// as returned by COMPUTE_GET_STATUS / SERVICE_GET_STATUS when `includeMetrics` resolves to
+// true. Sampled on a fixed cadence (node-side `C2D_METRICS_INTERVAL_SECONDS`), so values can
+// be slightly stale — see `collectedAt`. Owner-only; never present on unauthenticated or
+// listing responses.
+export interface ContainerMetricsSnapshot {
+  collectedAt: string // ISO timestamp of the sample
+  containerState: {
+    status: string // 'running' | 'exited' | ...
+    startedAt?: string
+    finishedAt?: string
+    exitCode?: number
+    oomKilled: boolean
+    error?: string
+    restartCount: number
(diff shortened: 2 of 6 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR adds TypeScript definitions for Docker and NVML runtime metrics and introduces the `includeMetrics` parameter to `computeStatus` and `getServiceStatus` in the HTTP and P2P providers. It carefully handles signature requirements to avoid unnecessary rounds of authentication unless metrics are explicitly requested.

Comments:
• [INFO][performance] Excellent design choice here! By conditionally fetching `getSignedCommandParams` only when `includeMetrics === true`, you prevent unnecessary signature requests (and potential wallet popups for users) when fetching standard status. This is great for both performance and UX.
• [INFO][style] Changing the return type from `ComputeJob` to `NodeComputeJob` correctly reflects the richer object actually returned by the Node (including the newly added `runtimeMetrics`), while remaining backward-compatible since `NodeComputeJob` extends `ComputeJob`. LGTM!

PR #2133, opened 2026-08-14
hardcode http paths

### LINKED ISSUE #2127: Remove getServiceEndpoints call
- [ ] use direct urls for each method instead of a call to get endpoints and then create the final url
- [ ] remove the endpoint from ocean-node

alexcos20:
Closes #2127 
# Hardcode Ocean Node HTTP endpoints, drop root-document discovery

## Summary

Removes the runtime endpoint-discovery dance from `HttpProvider.ts` — every HTTP call used to `GET <nodeUri>/`, parse the announced `serviceEndpoints` map, then look up a path by name before doing the actual request. That "discovery" was an illusion: the announced paths come from a hand-maintained static map shipped in the same ocean-node release as this SDK, not anything that varies per deployment. This PR hardcodes each path directly inside the method that uses it — the same style `Aquarius.ts` already used for 100% of its routes — and fixes two path mismatches that were silently 404ing.

### `src/services/providers/HttpProvider.ts`

- All 38 call sites that fetched the root document and resolved a path via `getEndpoints()` → `getServiceEndpoints()` → `getEndpointURL()` now build the literal path inline (e.g. `nodeUri.replace(/\/+$/, '') + '/api/services/nonce'`), matching exactly what the node currently announces for each route.
- Deleted `getServiceEndpoints()` and `getEndpointURL()` — no longer reachable from anywhere.
- Deleted `resolvePersistentStorageRoute()` and `resolveServiceRoute()`, the two "try the announced name, else fall back to a hardcoded path" helpers — every persistent-storage and service-on-demand route already only ever hit the fallback (the node never advertises those names), so this is now just the fallback path, un-conditionally.
- `getEndpoints()` renamed to private `getNodeInfo()`, kept alive **only** for `getNodePublicKey()`, which still needs the root document's `nodePublicKey` field (not a route). `isValidProvider()` is untouched — it already fetched the root URL directly.
- `getSignedCommandParams()` no longer takes `providerEndpoints`/`serviceEndpoints` params — they only existed to avoid double-fetching the root document, which no longer happens at all.
- **Bug fix:** `serviceGetStreamableLogs()` was resolving via the fallback helper to `/api/services/serviceGetStreamableLogs`, but the node registers `/api/services/serviceStreamableLogs` (no "Get") — this call always 404'd. Now hardcoded to the correct path.
- No shared route-table module was introduced. `P2pProvider.ts` dispatches protocol commands, not HTTP paths, so it would've been the table's only other possible consumer — and it doesn't need one.

### `src/services/providers/BaseProvider.ts`

- `getNonce()`'s trailing `providerEndpoints`/`serviceEndpoints` params deleted outright (clean removal, not deprecated-and-ignored) — this ships as part of the v9.0.0 major, so there's no reason to keep dead parameters around. No caller in this repo passed them.

### `src/@types/Provider.ts`

- Deleted the `ServiceEndpoint` interface — it only ever described the now-deleted discovery response shape. Also a clean removal, not a deprecation.

### `src/services/Aquarius.ts`

- **Bug fix:** `querySearch()` POSTed to `/api/aquarius/assets/query`, which doesn't exist on the node — only `/api/aquarius/assets/metadata/query` does. Fixed to the real path. (`Aquarius.ts` had no discovery logic to remove; every route there was already hardcoded.)

### `src/@types/Services.ts`

- Trimmed a comment on `ServiceJobEndpoint` that referenced the now-deleted `ServiceEndpoint` type.

### `test/integration/Services.test.ts`

- Added a test for `serviceGetStreamableLogs()` against a running service, asserting a non-null, async-iterable response — there was no existing coverage for this method, so the bug above had no regression guard.

## Behavior / compatibility

- **Fewer HTTP requests.** Every migrated method drops one root-document fetch (two for persistent-storage/service-on-demand calls, which previously fetched it twice via `getSignedCommandParams`). Chatty flows (compute-status polling, persistent-storage file loops) see the biggest reduction.
- **Behavior change on the error path:** a node that doesn't implement a given route now returns a 404 instead of the SDK silently resolving to `null` (the `if (!path) return null` guards are gone because a hardcoded path is never absent). This is a more honest failure mode — an unreachable node previously also produced `null`, indistinguishable from "route not supported."
- **Two breaking type/signature removals**, both clean (no `@deprecated` shim), justified by this landing in the v9.0.0 major:
  - `BaseProvider.getNonce()` no longer accepts `providerEndpoints`/`serviceEndpoints`.
  - `ServiceEndpoint` type removed from the public export surface.
- **Two behavior fixes**, both previously-404ing routes now resolve correctly:
  - `serviceGetStreamableLogs()` → `/api/services/serviceStreamableLogs`.
  - `Aquarius.querySearch()` → `/api/aquarius/assets/metadata/query`.
- P2P transport (`P2pProvider.ts`) is untouched — verified it makes zero `fetch()` calls; everything there dispatches through `PROTOCOL_COMMANDS`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated metadata searches to use the correct Aquarius API route.
  * Improved provider request routing for more reliable node, service, job, file, and authentication operations.
  * Corrected handling of job URLs, query parameters, trailing slashes, and unsuccessful responses.

* **Tests**
  * Added coverage for job routes, metadata searches, and streamable service logs.
  * Added validation for successful and invalid job requests, including fallback behavior.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
           bash -x start_ocean.sh 2>&1 > start_ocean.log &
         env:
           CONTRACTS_VERSION: '2.9.0'
-          NODE_VERSION: "pr-1408"
+          NODE_VERSION: "pr-1448"
       - name: Install deps & build
         run: npm ci && npm run build:metadata
       - name: Wait for node to be ready
@@ -157,7 +157,7 @@ jobs:
           bash -x start_ocean.sh 2>&1 > start-node.log &
         env:
           CONTRACTS_VERSION: '2.9.0'
-          NODE_VERSION: "pr-1408"
+          NODE_VERSION: "pr-1448"
       - name: Install deps & build
         run: npm ci && npm run build:metadata
 
--- a/src/@types/Provider.ts
+++ b/src/@types/Provider.ts
@@ -41,12 +41,6 @@ export interface ProviderComputeInitializeResults {
   payment?: ProviderComputeInitializePayment
 }
 
-export interface ServiceEndpoint {
-  serviceName: string
-  method: string
-  urlPath: string
-}
-
 export interface NodeP2P {
   nodeId: string
   multiaddress?: Multiaddr[]
--- a/src/@types/Services.ts
+++ b/src/@types/Services.ts
@@ -56,8 +56,7 @@ export interface ServiceTemplatePublic {
 
 // ── Runtime service job ────────────────────────────────────────────────
 
-// Port mapping for a running service (named ServiceJobEndpoint to avoid colliding
-// with the provider-route `ServiceEndpoint` type used for node endpoint discovery).
+// Port mapping for a running service.
 export interface ServiceJobEndpoint {
   containerPort: number
   hostPort: number
--- a/src/@types/Services.ts
+++ b/src/@types/Services.ts
@@ -56,7 +56,6 @@ export interface ServiceTemplatePublic {
 
 // ── Runtime service job ────────────────────────────────────────────────
 
-// Port mapping for a running service.
 export interface ServiceJobEndpoint {
   containerPort: number
   hostPort: number
--- a/src/services/Aquarius.ts
+++ b/src/services/Aquarius.ts
@@ -239,7 +239,7 @@ export class Aquarius {
     signal?: AbortSignal,
     authorization?: string
   ): Promise<any> {
-    const path = this.aquariusURL + '/api/aquarius/assets/query'
+    const path = this.aquariusURL + '/api/aquarius/assets/metadata/query'
 
     try {
       const response = await fetch(path, {
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -12,7 +12,6 @@ import {
   ComputeResultStream,
   ProviderInitialize,
   ProviderComputeInitializeResults,
-  ServiceEndpoint,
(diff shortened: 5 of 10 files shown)


alexcos20 (commented):

Overall risk: low

Summary:
This PR simplifies the HTTP Provider by removing dynamic endpoint resolution and replacing it with standardized, hardcoded REST paths. This is a significant performance enhancement as it eliminates unnecessary HTTP roundtrips before each request. The PR also successfully resolves a 404 error in the Aquarius API route and provides excellent test coverage to explain edge case behaviors of the provider node. LGTM!

Comments:
• [INFO][performance] Removing the dynamic discovery of `providerEndpoints` and `serviceEndpoints` is a fantastic optimization. This eliminates redundant HTTP roundtrips on almost every node interaction, which will noticeably speed up SDK operations.
• [INFO][style] Good addition of the `baseUrl(nodeUri)` helper. It robustly strips trailing slashes using Regex (`/\/+$/`) to prevent malformed double-slash URLs when appending hardcoded paths.
• [INFO][bug] Great catch correcting the Aquarius query path to `/api/aquarius/assets/metadata/query`. This correctly patches the 404 issue without impacting other operations.
• [INFO][other] Passing the literal string `/:job` in the path initially looks like a placeholder bug, but the added test cases (`HttpProviderRoutes.test.ts`) clearly document and prove that this is a required quirk of the underlying ocean-node. Excellent job covering this oddity with explicit integration tests.

PR #2134, opened 2026-08-17
Feature/auth for policy server

alexcos20:
# Authenticate the PolicyServer calls, and stop logging credentials

## What

Client-side counterpart to ocean-node
[#1449](https://github.com/oceanprotocol/ocean-node/pull/1449) ("Authenticate
`PolicyServerPassthrough` / `initializePSVerification`, and stop logging credentials").

Both PolicyServer endpoints are now authenticated by the node: a caller must supply an
`Authorization` token or a `nonce` + `signature` pair together with `consumerAddress`, or
the request is rejected with `401` and never reaches the PolicyServer. `initializePSVerification`
also became its own protocol command (`PolicyServerInitialize`) so a signature is scoped to
one endpoint and cannot be replayed against the other.

Until now `ProviderInstance.PolicyServerPassthrough()` and
`ProviderInstance.initializePSVerification()` POSTed the caller's request object verbatim,
with no `Authorization` header and no identity fields — against a node carrying #1449 **both
calls would start failing with 401**. This PR makes them carry credentials the same way every
other authenticated Provider method already does, on both the HTTP and P2P transports, and
mirrors the node's third fix (credential redaction) on the client's own log lines.

- Credentials are derived from a `SignerOrAuthTokenOrSignature` via the existing
  `getSignedCommandParams()`: a `Signer` fetches `getNonce()+1` and signs
  `consumerAddress + nonce + <command>`; a JWT string is sent as the `Authorization` header;
  a precomputed `CompleteSignature` is forwarded as-is.
- The signed message differs per endpoint, matching the node:
  | method | command | signed message |
  | --- | --- | --- |
  | `PolicyServerPassthrough` | `PolicyServerPassthrough` | `consumerAddress + nonce + "PolicyServerPassthrough"` |
  | `initializePSVerification` | `PolicyServerInitialize` | `consumerAddress + nonce + "PolicyServerInitialize"` |
- The derived identity always wins over anything the caller put in the request object —
  the node overwrites those fields after verification anyway, so sending a different value
  would only produce a confusing mismatch.
- The caller's auth token is relayed by the node to the PolicyServer, so `POLICY_SERVER_URL`
  should be an operator-controlled HTTPS endpoint (see the node's `docs/PolicyServer.md`).

## Changes

- `src/@types/Provider.ts` — added `POLICY_SERVER_INITIALIZE: 'PolicyServerInitialize'` to
  `PROTOCOL_COMMANDS`.
- `src/@types/PolicyServer.ts` — `PolicyServerPassthroughCommand` gains optional
  `consumerAddress`/`nonce`/`signature`; `PolicyServerInitializeCommand` gains `nonce`/
  `signature`. Documented inline as node-verified identity that the lib fills in.
- `src/services/providers/HttpProvider.ts` — both methods take `signerOrAuthToken` as the
  new 2nd argument, resolve credentials through `getSignedCommandParams()`, send them in the
  body and set the `Authorization` header when a token is used.
- `src/services/providers/P2pProvider.ts` — same signature change; credentials are added to
  the P2P payload and `signerOrAuthToken` is now passed to `sendP2pCommand` (was `null`), so
  the `authorization` field actually reaches the node.
  **Bug fix:** `initializePSVerification` was sending the `PolicyServerPassthrough` command —
  the same bug #1449 fixed on the node's HTTP route — and so hit the wrong handler. It now
  sends `PolicyServerInitialize`.
- `src/services/providers/BaseProvider.ts` — façade signatures updated to thread the new
  argument through to the HTTP/P2P implementation.
- `src/utils/General.ts` — new exported `redactSensitiveFields()`, ported from the node's
  `validateCommands.ts`: a non-mutating recursive walk over plain objects/arrays that masks
  `authorization`, `signature`, `aes_encrypted_key` and `encryptedDockerRegistryAuth` with
  `[REDACTED]`, with a `WeakSet` cycle guard (`[CIRCULAR]`) and pass-through for non-plain
  objects (Buffers, Dates, streams).
- `src/services/providers/HttpProvider.ts` — every `LoggerInstance.error('Payload was:', …)`
  site now logs the redacted copy: `initializeCompute`, `computeStart`, `freeComputeStart`
  (all of which dumped the consumer's signature and encrypted docker/registry material on
  any failure) plus the two PolicyServer methods.

### Client-side guards

Rather than round-tripping to get the node's new `400`s, the lib now throws early on:

- a missing credential — `"a signer, auth token or signature is required"`;
- a `policyServerPassthrough` that is missing, not an object, or an **array** (arrays would
  be forwarded as `{"0":…,"1":…}` with no `action`);
- a resolved `consumerAddress` that fails ethers `isAddress`.

## Compatibility

- **Breaking (client API):** `PolicyServerPassthrough(nodeUri, request, signal?)` →
  `PolicyServerPassthrough(nodeUri, signerOrAuthToken, request, signal?)`, and identically
  for `initializePSVerification`. Callers on `9.0.0-next.*` must insert the credential as the
  second argument. There are no in-repo callers outside `BaseProvider`.
  ```ts
  await ProviderInstance.PolicyServerPassthrough(
    nodeUri,
    signerOrAuthToken,                 // Signer | JWT string | CompleteSignature
    { policyServerPassthrough: { action: 'newDDO', rawDDO: {} } }
  )
  ```
- **Node requirement:** the credentials are additive on the wire, so the calls keep working
  against a pre- # 1449 node (which simply ignores the extra fields). The P2P
  `PolicyServerInitialize` command, however, only exists on nodes carrying # 1449 — against an
  older node that P2P call now hits an unsupported command instead of silently reaching the
  passthrough handler.

## Summary by CodeRabbit

* **New Features**
  * Added authenticated Policy Server initialization and passthrough requests.
  * Added support for consumer addresses, nonces, and signatures in Policy Server commands.
  * Added a dedicated Policy Server initialization command.

* **Security**
  * Sensitive credentials are now redacted from error logs.
  * Requests validate consumer addresses and include authentication details securely.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,7 +64,7 @@ jobs:
           bash -x start_ocean.sh 2>&1 > start_ocean.log &
         env:
           CONTRACTS_VERSION: '2.9.0'
-          NODE_VERSION: "pr-1408"
+          NODE_VERSION: "pr-1449"
       - name: Install deps & build
         run: npm ci && npm run build:metadata
       - name: Wait for node to be ready
@@ -157,7 +157,7 @@ jobs:
           bash -x start_ocean.sh 2>&1 > start-node.log &
         env:
           CONTRACTS_VERSION: '2.9.0'
-          NODE_VERSION: "pr-1408"
+          NODE_VERSION: "pr-1449"
       - name: Install deps & build
         run: npm ci && npm run build:metadata
 
--- a/src/@types/PolicyServer.ts
+++ b/src/@types/PolicyServer.ts
@@ -1,10 +1,19 @@
 export interface PolicyServerPassthroughCommand {
   policyServerPassthrough?: any
+  // caller identity, verified by the node before anything is forwarded to the policy
+  // server. filled in from the signer / auth token, so any value set here is overwritten
+  consumerAddress?: string
+  nonce?: string
+  signature?: string
 }
 
 export interface PolicyServerInitializeCommand {
   documentId?: string
   serviceId?: string
-  consumerAddress?: string
   policyServer?: any
+  // caller identity, verified by the node before anything is forwarded to the policy
+  // server. filled in from the signer / auth token, so any value set here is overwritten
+  consumerAddress?: string
+  nonce?: string
+  signature?: string
 }
--- a/src/@types/PolicyServer.ts
+++ b/src/@types/PolicyServer.ts
@@ -1,10 +1,5 @@
 export interface PolicyServerPassthroughCommand {
   policyServerPassthrough?: any
-  // caller identity, verified by the node before anything is forwarded to the policy
-  // server. filled in from the signer / auth token, so any value set here is overwritten
-  consumerAddress?: string
-  nonce?: string
-  signature?: string
 }
 
 export interface PolicyServerInitializeCommand {
--- a/src/@types/Provider.ts
+++ b/src/@types/Provider.ts
@@ -129,6 +129,7 @@ export const PROTOCOL_COMMANDS = {
   HANDLE_INDEXING_THREAD: 'handleIndexingThread',
   COLLECT_FEES: 'collectFees',
   POLICY_SERVER_PASSTHROUGH: 'PolicyServerPassthrough',
+  POLICY_SERVER_INITIALIZE: 'PolicyServerInitialize',
   GET_P2P_PEER: 'getP2PPeer',
   GET_P2P_PEERS: 'getP2PPeers',
   GET_P2P_NETWORK_STATS: 'getP2PNetworkStats',
--- a/src/services/providers/BaseProvider.ts
+++ b/src/services/providers/BaseProvider.ts
@@ -643,18 +643,30 @@ export class BaseProvider {
 
   public async PolicyServerPassthrough(
     nodeUri: OceanNode,
+    signerOrAuthToken: SignerOrAuthTokenOrSignature,
     request: PolicyServerPassthroughCommand,
     signal?: AbortSignal
   ): Promise<any> {
-    return this.getImpl(nodeUri).PolicyServerPassthrough(nodeUri, request, signal)
+    return this.getImpl(nodeUri).PolicyServerPassthrough(
+      nodeUri,
+      signerOrAuthToken,
+      request,
+      signal
+    )
   }
 
   public async initializePSVerification(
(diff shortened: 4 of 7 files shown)


alexcos20 (commented):

Overall risk: medium

Summary:
This PR improves the security of the Policy Server interaction by authenticating passthrough and initialization requests. It introduces payload redaction to prevent sensitive credentials from leaking into logs and adds strict address validation. The implementation is generally solid, but there are potential edge cases regarding case-sensitivity in the redaction logic and potential integration issues with Auth Token support.

Comments:
• [WARNING][security] The redaction check `SENSITIVE_PAYLOAD_FIELDS.includes(key)` is case-sensitive. If a payload uses a different casing (e.g., `'Authorization'` or `'Signature'`), it will not be redacted and could leak into logs. Consider using a case-insensitive check and ensuring the constants array is completely lowercase.

```diff
-    if (SENSITIVE_PAYLOAD_FIELDS.includes(key)) {
+    if (SENSITIVE_PAYLOAD_FIELDS.includes(key.toLowerCase())) {
```

Make sure to also update the constant definition above:
```diff
-  'encryptedDockerRegistryAuth' // compute: encrypted docker registry credentials
+  'encrypteddockerregistryauth' // compute: encrypted docker registry credentials
```
• [INFO][style] Excellent implementation of `redactSensitiveFields`. The use of `WeakSet` for cycle detection and bypassing deep recursion on standard prototypes (like Buffers or Dates) is highly robust and follows best practices for deep traversal functions.
• [WARNING][bug] You are strictly enforcing `isAddress(consumerAddress)` here. The comment above mentions that an Auth Token can be provided instead of a Signer. If `signerOrAuthToken` is an Auth Token string, ensure that `getSignedCommandParams` actually decodes it and returns a valid `consumerAddress`. If it returns an empty string or `undefined` (which is common for non-signer credentials in previous versions), this check will incorrectly throw and break Auth Token support for this endpoint.
• [WARNING][bug] Similar to the `HttpProvider`, ensure that the strict `isAddress(consumerAddress)` check does not break the ability to use Auth Tokens if `getSignedCommandParams` does not output a valid address for Auth Token credentials.
• [INFO][bug] Good catch fixing the command string here. Previously, `initializePSVerification` was incorrectly sending `POLICY_SERVER_PASSTHROUGH` instead of `POLICY_SERVER_INITIALIZE`.

AdriGeorge (approved):

PR #2135, opened 2026-08-17
use node 22 & 24 to test builds

alexcos20:
Changes proposed in this PR:

- test build with node 22 & 24
- remove node v20
-

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,7 +228,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        node: ["22", "20"]
+        node: ["22", "24"]
 
     steps:
       - uses: actions/checkout@v3


alexcos20 (commented):

Overall risk: low

Summary:
Updated CI workflow to test against Node.js versions 22 and 24, replacing version 20. The change is small but entails dropping Node 20 from the automated testing matrix, which should be verified against the project's supported runtime policy.

Comments:
• [WARNING][other] The CI matrix has been updated to use Node versions 22 and 24, effectively dropping Node 20. If `ocean.js` still officially supports Node 20 (which is an active LTS version), it is highly recommended to retain it in the test matrix to prevent undetected regressions for users on that environment. Consider expanding the matrix rather than replacing:
```diff
-        node: ["22", "24"]
+        node: ["20", "22", "24"]
```
Additionally, depending on the current action runner environment, ensure that `actions/setup-node` successfully resolves version `24` to avoid immediate workflow failures. If dropping Node 20 is intentional, ensure the `engines` field in `package.json` reflects the newly supported minimum version.

PR #2137, opened 2026-08-20
update deps

### LINKED ISSUE #2138: Ocean.JS dependency updates

alexcos20:
# Update all dependencies, replace microbundle with tsup, go ESM-only, migrate to ESLint 10

## What

A dependency and toolchain refresh. Every dependency moves to the newest version that has been
on npm for more than 7 days, the unused ones are removed, the build tool is replaced, and the
package becomes ESM-only. Along the way this fixes a build that was only reproducible from the
committed lockfile, a type leak that broke consumer `tsc`, and three test defects.

Targets the `9.0.0` line, which is the right place for the breaking changes below.

---

## ⚠️ Breaking changes

| change | impact |
| --- | --- |
| **Package is ESM-only** — `main` and `exports["."].require` removed, no more `dist/lib.cjs` | `require('@oceanprotocol/lib')` now needs Node's `require(esm)` (Node ≥ 22.12) or `await import()`. See "ESM-only" below — the CJS entry never actually worked. |
| **`engines.node` `>=18` → `>=22`** | Node 18/20 are no longer supported. CI already only tested 22 and 24, and `.nvmrc` was already 22. |
| **UMD build dropped** (`umd:main` and `dist/lib.umd.js` removed) | The UMD file externalised 27 modules, so it was unusable from a `<script>` tag without supplying every one as a global. |
| **libp2p and friends moved to `dependencies`** | They are now installed for every consumer instead of being inlined into the bundle. This *fixes* consumer type-checking (see below) at the cost of a heavier install. |

### ESM-only — why this is a fix, not a regression

`dist/lib.cjs` could never work. 16 of the 25 runtime dependencies (the libp2p stack,
`multiformats`, `uint8arrays`) publish no `require` condition, so the CJS bundle threw on load:

```
node v18.20.4  require(CJS) -> ERR_PACKAGE_PATH_NOT_EXPORTED
node v20.16.0  require(CJS) -> ERR_PACKAGE_PATH_NOT_EXPORTED
node v20.19.0  require(CJS) -> ERR_PACKAGE_PATH_NOT_EXPORTED
node v22.23.1  require(CJS) -> ERR_PACKAGE_PATH_NOT_EXPORTED
```

It also failed to bundle — esbuild could not resolve `@libp2p/peer-id` from it. Verified the
same failure against the previously published `8.6.2` tarball, so this predates the PR.

The ESM entry, by contrast, is requireable from CJS on modern Node:

```
node v20.19.0  require(ESM) -> OK (74 exports)
node v22.23.1  require(ESM) -> OK (74 exports)
```

So consumers gain a working CJS path on Node ≥ 20.19 where they previously had none. README
documents this.

### Also fixed: `exports["."].default` pointed at a file that was never published

microbundle wrote its `modern` and `esm` outputs to the same `lib.module.mjs`, so
`dist/lib.modern.mjs` never existed — in this branch *or* in any released version (checked the
`8.6.2` tarball). Anything whose resolver fell through to `default` got a missing file. tsup now
emits it for real.

---

### Updated

All to the newest release older than 7 days.

**Runtime:** `@oceanprotocol/ddo-js` 0.3.0 → 0.4.1, `ethers` 6.15 → 6.17, `bignumber.js` 9 → 11,
`eciesjs` 0.4.5 → 0.5.0, `decimal.js`, `crypto-js`, `jsonwebtoken`,
`@oasisprotocol/sapphire-paratime` 1.3.2 → 2.3.0, the whole libp2p family, `multiformats` → 14.

**Dev:** `@types/node` 24 → 26, `typescript-eslint` → 8.67, `eslint` 8 → 10, `prettier` 2 → 3,
`typedoc` 0.25 → 0.28, `release-it` 19 → 21, `mocha`, `tsx`, `chai`, `auto-changelog`.

Two deliberate exceptions:

- **`typescript` stays on 6.0.3.** TypeScript 7.0 ships no programmatic API (that lands in 7.1),
  so `typescript-eslint` refuses it outright ("typescript-eslint does not support TS 7.0",
  [typescript-eslint#10940](https://github.com/typescript-eslint/typescript-eslint/issues/10940))
  and `typedoc` crashes. No release of either supports TS 7, including their canary/dev channels.
  TS 7 *was* trialled successfully via Microsoft's side-by-side layout
  (`@typescript/typescript6` for the API, `typescript@7` for `tsc`) and the emitted declarations
  were equivalent — 30 differing lines across 62 files, all semantically identical. Reverted by
  choice; the escape hatch is documented if we want it later.
- **`@oceanprotocol/ddo-js` is on 0.4.1**, which is younger than 7 days. Adopted deliberately
  since we control that release.

### Removed — 10 unused

`@truffle/hdwallet-provider` (never imported; was the source of most audit findings), `c8`
(scripts use `nyc`), `cross-env`, `ora`, `ts-node` and `ts-node-register` (`ts-node` was named by
two scripts but no loader was ever registered, so it did nothing), `chai-spies` +
`@types/chai-spies` (zero usages), `fs` (npm's empty security-holder stub — `import fs from 'fs'`
resolves to the Node builtin), and the `web3` **peerDependency** (not imported anywhere).

Dropping `ts-node` meant `npm run mocha` had to go — it was broken anyway
(`Cannot find module '.../src/index.js'`), and `test:sapphire` went through it, so **the Sapphire
suite was unrunnable**. It now uses the same `tsx` runner as the other suites.

### Moved devDependencies → dependencies

The libp2p family, `@multiformats/multiaddr`, `multiformats`, `uint8arrays`, `cross-fetch`.

`src/` imports these at runtime, but they sat in `devDependencies`, so microbundle inlined them
into the bundle instead of externalising them. Two consequences, both now fixed:

1. **Consumer type-checking was broken.** `dist/types` leaked 5 unresolvable imports. Verified by
   installing the packed tarball into a clean project with `skipLibCheck` off:

   ```
   before: 5 × error TS2307: Cannot find module 'libp2p' / '@libp2p/interface' /
                             '@multiformats/multiaddr' / 'multiformats/cid'
   after:  0 errors
   ```

   This was invisible internally because our own `tsconfig.json` sets `skipLibCheck: true`.
2. Consumers could not dedupe or pin libp2p, and an app using libp2p directly ended up with two
   copies.

Six of these were **not declared at all** — `cross-fetch`, `uint8arrays`, `@libp2p/interface`,
`@libp2p/peer-id`, `@libp2p/ping`, `@libp2p/utils` were imported by `src/` and only resolved by
hoisting. Removing `@truffle/hdwallet-provider` broke the build by taking `cross-fetch` with it,
which is how they surfaced.

### Result

- runtime deps 8 → 25, devDeps 45 → 27, installed packages **1732 → 721**
- `npm audit` **68 → 3** findings, criticals **3 → 0**
- published ESM bundle **~346 kB → ~49 kB gzipped**

The 3 remaining audit findings (`diff`, `mocha`, `serialize-javascript`) are dev-only, have no
upstream fix (mocha 11.8.0 is latest and still pins the vulnerable ranges), and each has a
precondition this repo does not meet — mocha's parallel mode, untrusted patch input.

---

## Build: microbundle → tsup

microbundle 0.15.1 has been unmaintained since 2022 and **the build was only reproducible from the
committed lockfile**. Deleting `package-lock.json` and reinstalling broke it with a babel error
pointing at our own source:

```
(babel plugin) SyntaxError: src/@types/Assets.ts: Support for the experimental syntax 'flow'
isn't currently enabled
> 1 | export type { DDO } from '@oceanprotocol/ddo-js'
```

Root cause: microbundle pins `rollup-plugin-typescript2 ^0.32`, whose default `include` is the
extglob `["*.ts+(|x)", "**/*.ts+(|x)"]`. **picomatch 2.3.2 stopped matching the empty extglob
alternative `+(|x)`**, so rpt2's filter rejected every `.ts` file and silently skipped
transpiling — it never even called `getEmitOutput`. babel then received raw TypeScript and, having
no `preset-typescript` (rpt2 was meant to strip types), parsed it as Flow. Demonstrated directly:

```
picomatch 2.3.1  "**/*.ts+(|x)" -> src/index.ts: true
picomatch 2.3.2  "**/*.ts+(|x)" -> src/index.ts: false
```

Rather than pin picomatch backwards, the build now uses **tsup** (`tsup.config.ts`), with
declarations from a separate `tsc --emitDeclarationOnly` step so the per-file
`dist/types/**/*.d.ts` tree that `exports["."].types` points at is preserved — tsup's own `dts`
would flatten it into one file.

Outputs are pinned to the filenames `package.json` already advertised:

| output | field |
| --- | --- |
| `dist/lib.module.mjs` | `module`, `exports["."].import` |
| `dist/lib.modern.mjs` | `exports["."].default` (now actually emitted) |

`rm -rf node_modules package-lock.json && npm install && npm run build` is now green, and the
whole rollup-2/babel/rpt2 chain is out of the tree.

**One override remains:** `"overrides": { "tsup": { "esbuild": "^0.28.2" } }`. tsup pins
`esbuild ^0.27.0`, but the fix for the esbuild dev-server advisory landed in 0.28.1. Scoped to
tsup so tsx's already-patched copy is untouched. Verified byte-for-byte safe — output identical
before and after, `dist/types` diff 0 lines, externals identical.

---

## Lint: ESLint 8 → 10, flat config

`.eslintrc` is replaced by `eslint.config.js`. `eslint-config-oceanprotocol` could not come along:
it is eslintrc-only and built on `eslint-config-standard@17`, which peers on eslint ^8. The stack
is now `@eslint/js` recommended + `typescript-eslint` recommended +
`eslint-plugin-prettier/recommended`, with `eslint-plugin-security` registered for the single
`security/detect-non-literal-fs-filename` rule that `src/` and `test/` actually rely on. Scripts
are plain `eslint .` — ESLint 10 removed `--ignore-path` and `--ext`.

The migration is deliberately **behaviour-neutral**: the gate still fails on exactly what it
failed on before. Raw migration produced 222 errors; the rules the old standardjs config never
applied are switched off and documented inline:

- `@typescript-eslint/no-explicit-any` — off, 173 hits
- `preserve-caught-error` — off, 25 hits, new in ESLint 10

Both are worth enabling as separate cleanups.

It also found real problems, now fixed: 5 wrapper-object types in public signatures
(`Promise<Boolean>`, `Promise<String>`, `Promise<Object>`, `BigInt` used as a type) corrected to
primitives, one `prefer-const`, two dead initializers, and 16 obsolete `eslint-disable`
directives removed.

Result: **0 errors, 47 warnings** (was 0/36; the rise is `@typescript-eslint/no-unused-vars`
catching type-aware cases the core rule missed).

---

## TypeScript config

`moduleResolution` `node` (node10) → `bundler`, `module` `ES2020` → `esnext`, in both
`tsconfig.json` and `test/tsconfig.json`. `strict: false` is now pinned explicitly because TS 6
flipped that default to `true`, and `rootDir: ./src` is set so declarations keep landing at
`dist/types/index.d.ts` rather than `dist/types/src/…`. README updated: the old
`"moduleResolution": "node"` advice cannot read this package's `exports` map and is removed in
TS 7.

---

## Test fixes

Three defects found while validating, all in tests:

1. **`config` shadowing in the guide-generating tests.** `CodeExamples.test.ts` and
   `ComputeExamples.test.ts` declared `let config: Config` at describe level, then `before()`
   declared a *second* `const config = new ConfigHelper()...` that shadowed it — so every `it()`
   block read `undefined`. Fixed by dropping the inner `const`. **`CodeExamples.md` and
   `ComputeExamples.md` are regenerated**, since they carried the broken pattern into the public
   docs.
2. **Three silently unregistered tests.** In `PublishEditConsume.test.ts`,
   `it('Should update arweave dataset')` opened at line 465 and did not close until 524, so three
   `it()` blocks sat inside its async callback. Mocha only registers tests during collection, so
   they never ran — absent from every log as passing, failing *and* pending. Two of them read DDOs
   from the onchain/graphql flows that are commented out end to end in that file and are removed
   with their six now-unused declarations; `Should resolve updated datasets` is lifted to describe
   level and **now runs and passes**, taking the integration suite from 75 to 76 tests. A sweep of
   every declared `it()` title against the run log confirms this was the only instance.
3. With those references gone, `no-unassigned-vars` is set to **`error`**.

---

## Verification

Ocean Market (`OceanProtocolEnterprise/market`) was checked for ESM-only impact: it pins
`@oceanprotocol/lib ^8.6.2`, is already `"type": "module"` with `engines.node: "22"`, and uses
`moduleResolution: "bundler"` + `skipLibCheck: true` with Next 15 and `esmExternals: 'loose'` — so
build and type-check are unaffected. Its Jest suite cannot load this library today either (v8
fails the same way), so ESM-only introduces no new breakage there; it will need Jest in ESM mode
or a mock whenever it moves to v9.

---

## Summary by CodeRabbit

* **New Features**
  * Added ESM-only package builds with modern and ES2020 output support.
  * Added generated TypeScript declaration files.
  * Updated Sapphire integration for Ethers v6 compatibility.

* **Improvements**
  * Improved policy-server verification and validation.
  * Redacted sensitive information from error logs.
  * Updated public TypeScript return types to use primitive values.
  * Raised the minimum supported Node.js version to 22.

* **Documentation**
  * Added guidance for ESM imports and CommonJS interoperability across Node.js 22 versions.

* **Maintenance**
  * Refreshed linting, build, and test configurations.


alexcos20 (commented):

Overall risk: low

Summary:
Excellent refactoring PR to update build tools (microbundle to tsup), switch to flat ESLint config, bump dependencies (libp2p, ethers, eciesjs), and clean up various TypeScript linting issues including primitive type usage. The shift to an ESM-only package aligns well with the broader JS ecosystem trends and is properly documented in the README. LGTM!

Comments:
• [INFO][style] Great job updating `Boolean` to `boolean`. Primitive types should always be preferred over boxed object types in TypeScript. The same applies to the `String`, `BigInt`, and `Object` changes made in other contract files.
• [INFO][style] Good catch on handling the `Uint8Array` returned by `eciesjs` 0.5+. Using `TextDecoder().decode(...)` is the standard and correct approach for buffer conversions here.
• [INFO][other] Switching to `tsup` is a solid choice given that `microbundle` is no longer actively maintained. The configuration faithfully reproduces the necessary output formats while dropping the legacy builds.
• [INFO][style] Good fix to prevent unnecessary variable shadowing and reassignment by splitting out `nftAddress` into its own `const` declaration.

alexcos20:
released 9.0.0-next.11 with this branch

PR #2139, opened 2026-08-26
P2p improve

alexcos20:
# P2P transport hardening, typed compute-provider search, cancellation

## Why

DHT provider records are valid for **48 hours**, but libp2p's peer store expires addresses
after **1 hour**, and re-storing an address it already holds does **not** refresh it
(`toPeerPB` carries the previous `observed` timestamp forward). So a provider record routinely
points at a peer whose addresses have already been discarded, and resolution fails.

Fixing that surfaced a second, worse problem in the same path: the length-prefixed frame reader
could return a **truncated response as a success**, and a large body consumed slower than it
arrives was silently discarded and partly replaced with corrupt frames.

### Peer store lifetime
`peerStore: { maxAddressAge, maxPeerAge }` both set to **48 h**, matching provider-record
validity. Placed *before* the user's `libp2p` spread, so an application override still wins.
`maxPeerAge` must stay `>=` `maxAddressAge` or the peer record is evicted while its addresses
are still valid.

### Frame accounting — truncation can no longer read as success
`LpFrameReader` counts bytes delivered by the transport against bytes accounted for by complete
frames. End-of-file with nothing pending is a clean end; anything else is truncation and
propagates. Replaces a message-regex heuristic that a peer could defeat by appending two bytes.

### Read-buffer overflow
`lpStream` was called with no options, taking the library's 4 MiB `maxBufferSize`. On overflow
the library **discards the entire buffer** and the rejection is a no-op when no read is pending
— i.e. exactly while a consumer is working. Result: ~10% of a 40 MiB body delivered, plus
corrupt out-of-sequence frames, and at baseline **no error at all**.

Fixed with four coordinated limits plus real `pause()`/`resume()` backpressure around each
yield: explicit `maxDataLength`, a raised `stream.maxReadBufferLength` (yamux grows its window
to 16 MiB while the read buffer stopped at 4 MiB), a resume water mark, and an explicit
`maxBufferSize`. See the comments at those constants for the derivation.

### Typed compute-provider search — new public API
`findComputeProviders(node, request)` on the provider, plus a pure capability module
(`c2dCapabilityContent`, `c2dBucketFor`, `C2dCapability`, `C2D_BUCKET_OVERRIDES`). Consumers
never build a JSON string, compute a bucket, or see a CID:

```ts
const result = await ProviderInstance.findComputeProviders(node, {
  free: false,
  resources: [{ resource: 'cpu', value: 8 }, { resource: 'fpga', value: 4 }],
  models: { gpu: 'A100' }
})
```

`resource` is an **open string**, and bucketing is a rule (largest power of two `<= value`), not
a table — so a resource type added in future needs no release of this package. `fpga` above is
not known to the code and works.

Bucketing rounds **down** on both the announce and query sides, which produces false positives
by design; **verification against real compute environments is therefore required for
correctness, not an optimisation**. A request for `cpu: 9` matches a node whose true maximum is
8 until verification drops it.

### Diagnostics — known versus reachable
`getDiscoveredNodes()` answered "who is in the peer store, at what addresses" and stopped there,
which is the half that decides nothing. A peer can sit in the store with three addresses and be
unreachable, be connected only over a circuit relay that drops the connection after a byte
budget, or be connected only because *it* dialled *us*.

Each entry now carries `connections`, `direction` (`inbound`/`outbound`/`mixed`), `transports`
and `limited`. `peerId` and `multiaddrs` are untouched, so every existing consumer is unaffected.
`limited` is true only when **every** connection to the peer is relay-limited — one unrestricted
connection is enough to carry a transfer, and getting that backwards would mark most healthy
peers as degraded.

A new `getP2pDiagnostics()` carries the node-wide facts: routing-table size, DHT mode, the peer
store lifetimes **the running node applies**, and the resolution counters. Separate rather than
folded in, because widening the array method's element type would break consumers that
destructure it. The routing table is the one to read first — a DHT lookup against an empty table
finds nothing however many connections are open, since a connection to a peer that does not speak
the DHT protocol is not a place a walk can start. An unreachable DHT service reports `undefined`
rather than `0`: not knowing is not the same as knowing the table is empty.

### Framing lives in its own module
The four flow-control constants, the `lpStream` wrapper, the pause/resume helpers and
`LpFrameReader` moved out of `P2pProvider.ts` into `services/providers/lpFraming.ts` — internal,
not added to the package root, so no public surface changes. The constants only make sense as a
set, each derived from the others.

A test now pins their **absolute** values. They are a cross-repository contract with ocean-node:
a peer declaring a frame larger than the other side's `maxDataLength` is rejected outright, and a
read buffer smaller than the window the muxer grows to is what silently discards a large body.
The relations between them would still hold if both repos moved a value in different directions,
so the numbers themselves are what get pinned.

### `libp2p` option bags merge instead of replacing
`peerStore`, `connectionManager` and `connectionMonitor` are now merged field by field with a
user override; everything else in the `libp2p` spread still replaces wholesale, which is right
for whole subsystems like `transports` and `services`.

For `peerStore` the old behaviour was a trap. It holds two values with an invariant between them,
so `libp2p: { peerStore: { maxAddressAge: … } }` — the obvious way to tune one — silently dropped
`maxPeerAge` back to the library's **6 h**, *shorter* than the address lifetime just set. A peer
entry that expires before its own addresses takes them with it: configuring the 48 h fix produced
the very failure it exists to prevent.

An inverted pair is still honoured as given — it is the caller's node — but it now logs a warning
rather than being silently corrected.

### A split reply is joined, not thrown away
A buffered reply does not arrive as one frame: ocean-node writes the status envelope and then
**one frame per chunk of its response stream**, so any reply larger than a Node stream chunk
(16—64 KiB) is split — a large DDO, a `GET_LOGS` dump, a busy `computeStatus`.

The client accumulated every frame and then parsed them in a loop that **overwrote** its result
each pass, so the caller received the last frame alone. For a split JSON reply that fragment is
not valid JSON, so it came back as raw bytes. Measured against a two-frame `{"nonce":42,…}`:
`getNonce()` returned **`null`**. Replies that fitted one frame — most of them, and every reply
in a test that did not deliberately split one — were unaffected, which is how this survived.

The body frames are now joined before parsing, with the status envelope excluded from the join
(it was already parsed, and concatenating it would make the result unparseable). Single-frame
replies and envelope-only replies behave exactly as before. Frames are copied out of the reader
rather than retained as views, matching what the download path already does.

### Buffered responses are capped
The framing limits bound one *frame* and the transport's backlog. They said nothing about how
many frames a caller accumulates, and two paths accumulate all of them: a buffered command reply,
and a P2P download, which returns the whole file as one `ArrayBuffer`. Both loops read to
end-of-stream, so a peer that simply keeps sending grows the array until the process dies — and
the per-frame idle timeout is no defence, because a peer sending steadily is never idle.

Two separate ceilings, because the sizes are not comparable: `maxBufferedCommandBytes`
(default 64 MiB, matching what ocean-node applies to its own accumulating readers, so the two
ends of the same exchange agree) and `maxBufferedDownloadBytes` (default 512 MiB — larger,
because with a download a large body is the point). Both config-overridable. Over-limit raises
`P2pError` with `type: 'protocol_failed'`, which is deliberately **not** retryable: another
attempt fetches the same oversized body.

Note what the download ceiling really measures — this API hands back the entire file in memory,
so a caller who needs to raise it should treat that as the moment to notice it buffers rather
than streams.

### Cancellation
Optional trailing `signal?: AbortSignal` on `getDownloadUrl` and `getComputeResult`, threaded
through the nonce round-trip, the dial, the first frame **and every tail frame**. Previously
these two had no cancellation at any point. The stream is reset on cancellation so the peer
stops producing.

**The HTTP transport honours it too.** The façade dispatches through a member typed `any`, so
the extra argument was accepted by the call and then dropped by a method with no parameter for
it — an HTTP consumer passing a signal got silence. Two requests were affected, and only one is
obvious: the result body, and the **nonce round-trip** that runs first whenever the credential is
a `Signer` rather than a token. `getDownloadUrl` performs no transfer at all — it returns a URL
for the caller to fetch — so that nonce request is the *whole* of its I/O, and it was
uncancellable. `getComputeResultUrl` gained the same optional trailing parameter for the same
reason.

### The dial budget is composed with the caller's signal, not replaced by it
`dialAndStream` used `signal ?? AbortSignal.timeout(dialTimeout)`, so the moment a caller passed
a signal — which the cancellation work above makes the recommended thing to do — the dial lost
its only time bound. A peer that accepts a connection and never completes the upgrade then parks
the call *and its slot in the process-wide concurrency gate* for as long as that signal lives,
which for the normal session-scoped `AbortController` is hours.

The bound is on the dial and the stream open only. The writes that follow stay on the caller's
own budget, because one of them can be a large request body that legitimately outlasts any dial.
With no caller signal it is the same single timer as before, so nothing changes for callers that
never passed one. The composite uses the local-controller helper rather than `AbortSignal.any`,
so it leaves no listener behind on a long-lived caller signal — asserted directly.

### Lifecycle
- A **rejected** `setupP2P()` no longer leaves its configuration installed; it rolls back to the
  last configuration that produced a live node.
- A P2P call racing `stopP2P`/`dispose`/`setupP2P` now rejects rather than proceeding against a
  stale node, and `stopP2P` no longer waits unbounded on an in-flight node creation.

### Smaller
Concurrency gate moved into `dialAndStream` so downloads and compute results take a slot;
`AbortSignal.any` composites replaced with a local controller (the old shape retained one
uncollectable entry per call on a long-lived caller signal); `isLocalPageOrigin` fails closed on
host-less origins (`data:`/`blob:`/`about:srcdoc` are no longer treated as local);
`maxConcurrentRequests: 0` no longer deadlocks; `CLIENT_MAX_CONNECTIONS` raised to 32 (kad-dht
tags up to 20 peers plus 4 bootstraps = 24, so a cap of 20 caused prune/redial churn).

Also: 15 relative import specifiers gained explicit `.js` / `index.js`, fixing **13 of 22**
errors a strict `NodeNext` consumer saw from our own emitted declarations.

## Breaking / behaviour changes — release notes needed

1. **`getComputeResultUrl()` now rejects on the P2P transport.** It was declared
   `Promise<string>` but returned the streaming generator cast to `string`, so
   `fetch(await getComputeResultUrl(...))` requested the literal `"[object AsyncGenerator]"` and
   `url.startsWith()` threw. The P2P protocol answers with the result **bytes** and carries no
   URL, so one never existed. Use `getComputeResult(...)`. HTTP is unaffected; the published
   type is unchanged.
2. **`getProvidersForString` return type narrowed** from `multiaddrs: any[]` to `Multiaddr[]`.
   Runtime is unchanged. Compile-only break for consumers that accessed arbitrary properties.
   The `Multiaddr` type is now re-exported from the package root.
3. **Truncated transfers throw** where they previously returned short payloads as success.
4. **Streaming commands throw** on a `4xx/5xx` status frame instead of yielding it as the first
   data chunk.
5. **A call racing a stop/setup rejects** instead of proceeding against a stale node.
6. **Opaque origins are no longer local**, so plain-`ws` dials to loopback/RFC1918 from
   `data:`/`blob:`/`about:srcdoc` pages are denied.
7. **A signal passed to the HTTP transport now actually cancels.** Only reaches callers who
   already pass one on `getDownloadUrl`/`getComputeResult`, and it does what they asked for —
   but a caller that passed a signal it never intended to be honoured will now see the call
   reject where it previously ran to completion.
8. **A dial now has a time bound even when the caller supplies a signal.** A dial that used to
   hang until a session-scoped signal fired fails after `dialTimeout` instead.
9. **A buffered reply or download past its ceiling now throws.** Previously it grew until the
   process ran out of memory. Defaults are 64 MiB for a command reply and 512 MiB for a
   download; both are configurable.
10. **A reply split across frames now returns the whole reply.** Previously it returned the raw
    bytes of its last fragment — a fix, but a caller that had worked around the old behaviour
    will see something different.

Everything else is additive; no other exported signature changed.

## Where to look hardest

- The four framing constants in `P2pProvider.ts` — each is load-bearing except the resume water
  mark, which two independent test passes showed is currently redundant (`maxReadBufferLength`
  caps any single resume flush below `maxBufferSize`). It is kept as margin against future
  constant changes.
- The rollback in `setupP2P`'s `catch`. A simpler "restore what I saw on entry" guard is
  **wrong** under concurrency — two rejecting callers leave the loser's bad config installed.
- `findComputeProviders`'s verification step. Skipping it returns providers that cannot satisfy
  the request.
- **Which signal each stage of `dialAndStream` runs on.** The dial and the stream open take the
  composed one; everything after must not, or a large request body inherits the dial budget.
- **`maxBufferedDownloadBytes`'s default.** 512 MiB is a judgement about what a caller of a
  buffering API can survive, not a protocol constraint. If real datasets are bigger, this is the
  number to argue with.

**Deploy order:** the announce side in ocean-node must ship before this is useful for compute
search — until then `findComputeProviders` returns empty against the live fleet, because the
fleet announces exact values while this client looks up buckets.

## Summary by CodeRabbit

- **New Features**
  - Added compute-provider discovery with resource, capacity, environment, and model matching.
  - Added capability serialization and resource-based capacity bucketing.
  - Added cancellation for downloads and compute-result retrieval.
  - Added P2P diagnostics, lifecycle controls, concurrency, and connection limits.

- **Bug Fixes**
  - Improved partial results, retries, truncated responses, stream backpressure, and connection recovery.
  - Clarified that P2P compute results are streamed directly rather than provided as URLs.

- **Tests**
  - Expanded coverage for discovery, cancellation, streaming, concurrency, lifecycle, diagnostics, and capabilities.


alexcos20 (commented):

Overall risk: low

Summary:
Outstanding PR that substantially improves the P2P architecture, concurrency management, streaming backpressure, and compute provider discovery. The transition to a process-wide shared libp2p node combined with detailed control over frame buffering handles significant performance bottlenecks and memory leak issues perfectly. The code quality, including the inline documentation and tests, is exceptional. LGTM!

Comments:
• [INFO][performance] Excellent architectural decision to move `libp2p` node initialization to a shared, process-wide singleton. Previously, generating a heavyweight node per provider instance could severely bottleneck the application and exhaust connections.
• [INFO][performance] Very clever implementation of `timeoutSignal` to avoid memory leaks. A common pitfall when combining `AbortSignal`s (such as with `AbortSignal.any()`) is that listeners remain attached to the parent signal until it aborts, leading to unbounded memory growth on long-lived connections. The manual `{ once: true }` listener cleanup handles this perfectly.
• [INFO][style] Good use of manual string building for the canonical capability content (`c2dCapabilityContent`). Relying on native `JSON.stringify` on objects can lead to non-deterministic hashing due to JS engine key ordering (especially with numeric-like keys), so enforcing the exact byte structure manually is highly robust for DHT lookups.
• [INFO][bug] The concurrency implementation inside `ConcurrencyGate` coupled with `dialAndStream` is extremely solid. You appropriately handle the lifecycle of the slot, assuring the release function is idempotent and that failed dials or dropped streams do not permanently reduce the allowed concurrency pool.

alexcos20:
@coderabbitai full review

alexcos20:
@coderabbitai review

PR #2148, opened 2026-09-02
add metadata for services

### LINKED ISSUE #2147: add service metadata

alexcos20:
Closes #2147 
# Add service metadata to Service-on-Demand client flows

## Summary

Adds an optional, owner-scoped `metadata` bag to the Service-on-Demand client surface in
`@oceanprotocol/lib`, mirroring [ocean-node #1464](https://github.com/oceanprotocol/ocean-node/pull/1464).

`metadata` is an arbitrary key→value map of scalar values (`string | number | boolean`,
≤ 1 KB) that a consumer can attach to a service when starting or restarting it. The node
persists it and returns it for both owner-scoped status query — and from
the node-wide service listing.

This is the service-on-demand analogue of the `metadata` already supported on compute jobs
(`computeStart` / `freeComputeStart`), and it reuses the same `ComputeJobMetadata` type
rather than introducing a new one.

## Motivation

ocean-node #1464 taught the node to accept, store, and return a per-service metadata bag on
`SERVICE_START` / `SERVICE_RESTART` / `SERVICE_GET_STATUS` /  `SERVICE_LIST`. Without the corresponding client
support, ocean.js callers had no typed way to set or read it. This PR closes that gap so
apps can label their services (e.g. an environment tag, a run counter, a UI flag) and read
those labels back from status.

### Types — `src/@types/Services.ts`
- Import `ComputeJobMetadata` from `./Compute.js` and reuse it (its shape,
  `{ [key: string]: string | number | boolean }`, matches the node's `DBComputeJobMetadata`).
- `ServiceJob` gains `metadata?: ComputeJobMetadata` — returned on owner-scoped
  `getServiceStatus` (`SERVICE_GET_STATUS`).
- `ServiceContainerSpec` gains `metadata?: ComputeJobMetadata`. Because `ServiceStartParams`
  extends this spec and `ServiceRestartParams` is an alias of it, the field is available on
  **both** start and restart params from a single declaration.

### HTTP transport — `src/services/providers/HttpProvider.ts`
- `serviceRestart`: destructure `metadata` from `params` and forward it with the existing
  replace-when-supplied pattern (`...(metadata !== undefined ? { metadata } : {})`).
- `serviceStart`: no code change — `metadata` is carried through the existing `...rest`
  spread. It is sent as-is (not routed through `encryptServiceUserData`), since it is a
  public label bag rather than secret `userData`.

### P2P transport — `src/services/providers/P2pProvider.ts`
- `serviceRestart`: same destructure + conditional-forward change as HTTP.
- `serviceStart`: no code change (forwarded via `...rest`).

### Façade — `src/services/providers/BaseProvider.ts`
- No change. `serviceStart` / `serviceRestart` pass `params` straight through, and
  `metadata` rides inside `params`. The incentive-notify path only inspects
  `image` / `tag` / `dockerCmd`, so it is unaffected.

### Tests — `test/integration/Services.test.ts`
- Start passes a `metadata` object; the owner-scoped `getServiceStatus` test asserts it
  round-trips exactly.
- The node-wide `getServices` (`SERVICE_LIST`) test asserts `metadata` is stripped, added
  to the existing list of sanitized fields.
- The restart test now supplies **only** `metadata` (a metadata-only restart stays in REUSE
  mode — it does not force RESPEC) and asserts the new value replaced the start metadata on
  status.

## Behavior & semantics

- `metadata` is returned by `getServiceStatus` (owner-scoped)
  and by `getServices` (node-wide listing).
- **Not encrypted.** Unlike `userData`, `metadata` is a public label bag and is sent
  verbatim.
- **Restart semantics.** On `serviceRestart`, supplying `metadata` replaces the stored bag;
  omitting it leaves the stored bag unchanged. Supplying `metadata` alone does **not** switch
  the restart into RESPEC mode (that is still triggered only by
  `image` / `tag` / `checksum` / `dockerfile` / `additionalDockerFiles`).
- **Size cap.** The 1 KB limit is enforced by the node, which fail-fasts with a 400 on
  oversized input. The client does not duplicate the check — consistent with how compute
  metadata already relies on the node — and documents the cap in the type comments.

## Design decisions

- **Reuse `ComputeJobMetadata`** instead of adding a `ServiceJobMetadata` alias: identical
  shape, and ocean.js already uses one type for both compute and service metadata.
- **No client-side size validation:** mirrors existing compute-metadata behavior and keeps
  the node as the single source of truth for `MAX_JOB_METADATA_SIZE`.

## Compatibility

Fully additive and backwards-compatible — `metadata` is optional everywhere. Callers that
don't set it are unaffected, and the wire format is unchanged when it's absent.

## Summary by CodeRabbit

* **New Features**
  * Added optional metadata support for service jobs and container specifications.
  * Metadata can be provided when starting or restarting a service.
  * Service status and service-list responses now include associated owner-scoped metadata.
  * Metadata is supported across HTTP and peer-to-peer service operations.
* **Bug Fixes**
  * Preserved and updated metadata correctly throughout service lifecycle operations.
* **Documentation**
  * Clarified that restart metadata replaces stored values and must not contain sensitive information.


alexcos20 (commented):

Overall risk: low

Summary:
Adds support for `metadata` in compute services, including types, provider payload mappings for `serviceRestart`, and comprehensive integration tests. The changes correctly reflect the new API contract and safely handle optional parameters. LGTM!

Comments:
• [WARNING][other] Setting `NODE_VERSION: pr-1464` (here and on line 156) is a great approach for cross-repository PR testing. Just a reminder to ensure this is reverted or updated to a stable release version before merging into the `main` branch, to prevent CI from relying on a temporary PR build.
• [INFO][style] Excellent inline documentation. Explicitly calling out the 1 KB cap, owner-scoped visibility, and the relationship with `SERVICE_LIST` makes the API contract very clear for future developers and consumers.
• [INFO][style] Good use of conditional object spreading. This ensures that `metadata` is not sent as an explicit `undefined` in the JSON payload when it is omitted by the caller, keeping the request clean.

alexcos20:
@coderabbitai full review

giurgiur99 (approved):

PR #2152, opened 2026-09-02
persistent-storage file download 

### LINKED ISSUE #2151: PersistentStorage:  download file from bucket

alexcos20:
Closes #2151 
# Feat: client binding for persistent-storage file download — `downloadPersistentStorageFile`

## Problem

Ocean Node now serves **raw file bytes** out of a persistent-storage bucket — a new
`persistentStorageDownloadFile` command reachable over both HTTP and P2P (see ocean-node
[#1466](https://github.com/oceanprotocol/ocean-node/pull/1466), "download a file from a
persistent-storage bucket"). `ocean.js` already exposes the rest of the persistent-storage surface
(create/update/get buckets, list/upload/get-object/delete files) but had **no way to pull a file's
bytes back down**: consumers would have to hand-roll a `fetch`/`dialAndStream` and the byte-stream
plumbing themselves. This PR adds the typed client binding so the download is first-class on
`ProviderInstance`, transport-agnostic, exactly like the sibling persistent-storage methods.

## Approach

Follow the existing Provider layering — `BaseProvider` is the transport-dispatching façade that
routes to `HttpProvider` or `P2pProvider` via `getImpl(nodeUri)`. Because the node returns raw bytes
(not JSON), the return shape mirrors `getComputeResult`: an `AsyncIterable<Uint8Array>` (the existing
`ComputeResultStream` type). This is memory-safe for large files, matches how P2P already streams
bytes back, and lets callers consume incrementally.

- **HTTP** GETs `/api/services/persistentStorage/buckets/:bucketId/files/:fileName` and wraps the
  response body with the existing `responseBodyToAsyncIterable` helper.
- **P2P** dispatches the `persistentStorageDownloadFile` command and reuses `getComputeResult`'s
  bulk-transfer streaming path (`dialAndStream` + status-frame check + flow-controlled generator).

Both accept an optional `offset` to resume a partial download (HTTP `Range` header / P2P payload
field), matching `getComputeResult`.

### 1. Types — `src/@types/Provider.ts`

- `PERSISTENT_STORAGE_DOWNLOAD_FILE: 'persistentStorageDownloadFile'` added to `PROTOCOL_COMMANDS`,
  next to the other persistent-storage commands. No new response type — the return is
  `ComputeResultStream` (bytes), already exported via the `@types` barrel.

### 2. HTTP transport — `src/services/providers/HttpProvider.ts`

- `downloadPersistentStorageFile(nodeUri, signerOrAuthToken, bucketId, fileName, offset?, signal?)`:
  signs the request with the standard `address + nonce + command` scheme (via
  `getSignedCommandParams`), GETs the file route (no `/object` suffix — that is the metadata call),
  sets an `Authorization` header for auth-token callers, adds `Range: bytes=<offset>-` when resuming,
  and returns `responseBodyToAsyncIterable(response.body)`.

### 3. P2P transport — `src/services/providers/P2pProvider.ts`

- `downloadPersistentStorageFile(...)` with the same signature. Cloned from `getComputeResult`'s
  streaming path rather than the JSON `sendP2pCommand` helper: `dialAndStream`, first-frame
  status-JSON check, then a flow-controlled `async function*` with the same idle-timeout, backpressure
  (`resumeReads`/`pauseReads`/`readFrame`), clean-end handling and stream-abort/`release` cleanup.

### 4. Façade — `src/services/providers/BaseProvider.ts`

- One public method taking `nodeUri: OceanNode`, delegating through `getImpl(nodeUri)` so HTTP/P2P
  selection is automatic. Placed alongside the other persistent-storage delegators.

### 5. Tests — `test/integration/Provider.test.ts`

- Made the uploaded file content deterministic (`fileContent`) so the round-trip can be asserted.
- New "downloads the uploaded file bytes" test: calls `downloadPersistentStorageFile`, collects the
  async-iterable chunks, and asserts the decoded bytes equal the uploaded content. Runs under both
  the HTTP and P2P integration matrices, and is skipped when the node lacks persistent storage.

## Why it's safe

- **Additive & backwards-compatible.** One new read-only client method; no existing method,
  signature, or type changes.
- **Consistent with the surface it joins.** Same signing scheme, façade dispatch, and naming
  (`downloadPersistentStorageFile`) as the existing `upload/delete/getPersistentStorageFile*` methods;
  same byte-streaming machinery as `getComputeResult`.
- **Depends on node support.** A node without the `persistentStorageDownloadFile` command will reject
  the request; the new test only runs where persistent storage is enabled.

## Summary by CodeRabbit

* **New Features**
  * Added support for downloading files from persistent storage over HTTP and P2P connections.
  * Downloads support streaming, authorization, cancellation, and resuming from a specified byte offset.
  * Service restart requests can now include optional metadata.

* **Bug Fixes**
  * Improved reliability when P2P download or compute-result streams encounter errors.
  * Improved persistent-storage integration coverage by verifying downloaded files match their uploaded content.
  * Added coverage confirming correct behavior for job-related routes.


alexcos20 (commented):

Overall risk: low

Summary:
This PR implements persistent storage file downloading across BaseProvider, HttpProvider, and P2pProvider. The implementation is solid, including proper flow control for large file streams in P2pProvider. However, a CI workflow configuration temporarily points to a PR version of the ocean-node which should be updated prior to merging.

Comments:
• [WARNING][other] You have pinned `NODE_VERSION` to `pr-1466`. Make sure to revert this or update it to the proper release version/tag before merging, to avoid testing against an ephemeral PR build on `main`.
• [WARNING][other] Same as above, ensure this `pr-1466` environment variable override is removed or updated to a stable tag before merging this pull request.
• [INFO][performance] Excellent handling of backpressure and flow control here! This is crucial for correctly downloading large files without exhausting memory or desynchronizing the frame parser.
• [INFO][style] Good use of the HTTP `Range` header for implementing the `offset` parameter.

giurgiur99 (approved):

dnsi0 (approved):

